PR #16, opened 2020-05-15
Add Dracula color theme

mwahlig:


Miodec:
Thanks. Checking out this theme i feel like i should removed the "active-word" variable. The change in color is quite confusing i think. What do you think?

PR #19, opened 2020-05-15
Add nord theme

mendelmaleh:


mendelmaleh:
First, thank you for making monkey-type, it is really nice!

There where some other things that I wanted to try doing, but I couldn't figure out how to properly run this. I tried making an app in firebase, but its my first time trying firebase, and didn't really succeed. Some instructions for running/deploying this would be very nice!

mwahlig:
@mendelmaleh Agreed, some instructions in the README would be helpful. I ended up creating a firebase app titled `monkey-type` to match the `.firebaserc` and commented out all the `firebase.analytics()` calls in order to get it to run locally for me.

Miodec:
When using firebase to host a server i noticed it just runs a node server, so you could try that with the server root being /public.

Im still very new to this so im not exactly sure whats the best course of action when collaborating on a firebase project

PR #34, opened 2020-05-18
Add Thai language

ammarptn:
Add Thai language


Miodec:
Hey, thanks for adding Thai. Before i merge it, does Thai use input processing, as in do some characters require multiple keystrokes like korean or japanese?

Sadly support for these kinds of special characters is not working at the moment. Im in the process of figuring it out.

ammarptn:
> Hey, thanks for adding Thai. Before i merge it, does Thai use input processing, as in do some characters require multiple keystrokes like korean or japanese?
> 
> Sadly support for these kinds of special characters is not working at the moment. Im in the process of figuring it out.

For the Thai's input. It is a single keystroke per character, same as the English input method. I already tested on my branch and it worked fine. I really enjoy do typing on the Monkey-type. let me know if you have any question.

Miodec:
Alright, Thai should be live :)

PR #80, opened 2020-05-28
Scroll Longform text

bhomie:
The window will scroll after the cursors passes text that is beyond the middle of the page, keeping the active text in the middle of the window.

fixes https://github.com/Miodec/monkey-type/issues/67


bhomie:
@Miodec good to go.

bhomie:
not entirely sure what broke 🤔 but this works as intended now.

PR #83, opened 2020-05-28
Add themes

bhomie:
Adds 3 Themes:
* Botanical
* Bingsu
* 9009

Trying some different things with some, feel free to change as you see fit but I think they all free pretty good.

### DIFF
--- a/public/css/style.scss
+++ b/public/css/style.scss
@@ -23,7 +23,7 @@ input{
 
 body {
   margin: 0;
-  padding: 2rem 2rem 0 2rem;
+  padding: 0;
   min-height: 100vh;
   background: var(--bg-color);
   font-family: "Roboto Mono";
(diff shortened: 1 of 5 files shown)


Miodec:
9009 - Currently the text typed has lower contrast to the background that the 'future' words. I want all the themes to be the same in the sense that the  text typed has a high contrast. That way, when i add an option for higher text contrast (soon) i can just flip the sub and main colors on the text. I think i would filp sub and main colors around, so that the settings and account is also a bit more readable.

Bingsu - The background gradient cuts off on long pages (settings, account). Buttons account and settings page have the same text color as the background. Command line has the same issue.

Botanical - Pretty sick, but same as 9009, i would swap sub and main around to keep the text typed higher contrast. I will be adding the contrast swap thing later on.

bhomie:
I hear ya, either way you decide to go in the future, having all of them the same just makes it easier to change things around. 

All themes made the same by swapping sub/main colors and changed the body css so that it can scale if content get's bigger. (this was causing the bingsu gradient bg issue)

Miodec:
These places are still a bit hard to read

image

image

bhomie:
These all work now, finally.

PR #91, opened 2020-05-28
changed styling of suggestions container to scroll and made the input…

Rangoons:
… sticky

### DIFF
--- a/public/css/style.scss
+++ b/public/css/style.scss
@@ -144,6 +144,9 @@ a:hover {
       text-align: center;
     }
   }
+  #fuzzyInput {
+    position: sticky;
+  }
   #commandLine {
     width: 50vw;
     background: var(--bg-color);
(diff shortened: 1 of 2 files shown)


Miodec:
The suggestions box still gets cut off in the screen is very, very small. You can scroll with the mouse, but when using arrow keys to make a selection, the suggestions box does not scroll

Rangoons:
@Miodec what is the smallest resolution you want to support?

Miodec:
I mean, ideally as much as we can. I think its getting cut off because the height is measured in vh but its distance from the top is static, measured in rem.

Rangoons:
changing the vh to 60 works a bit better on the smaller screen size. I'll have to investigate the arrow key scrolling. I'm going to guess its because the event is being repurposed to switch the highlighted item. I'll mess around with it this weekend

Miodec:
Any update on this?

Miodec:
Already implemented

PR #92, opened 2020-05-28
New Themes

Le-SirH:
### Themes
* Shadow
* Mint
* Miami

## Shadow Theme
A pure black background with a fading text color as you type.

### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)


Miodec:
Nice work! May i suggest changing neon's name to miami? (named after a keycap set) 

image

Le-SirH:
I like that name idea. I will do so. 👍

bhomie:
Shadow is dope, nice job!

public/themes/shadow.css
```diff
@@ -0,0 +1,83 @@
...
+    70% {
+        color: #444;
+    }
+    80% {
+        color: #aaa;
+    }
+    100% {
+        color: #eee;
+    }
+}
```
bhomie, line 51: You could reduce what you're typing by combining the keyframes:
`0%,100%{...}
20%, 80%{...}`
and so on.

public/themes/shadow.css
```diff
@@ -0,0 +1,83 @@
...
+    60% {
+        color: #444;
+    }
+    80% {
+        color: #222;
+    }
+    100% {
+        color: #000;
+    }
+}
```
bhomie, line 27: It's not necessary to have all these keyframes instead of just 'from' and 'to' since it will interpolate the values anyways.
`from {...} to {...}`

Le-SirH: Indeed. I didn't catch that for some reason. Thanks!

PR #136, opened 2020-06-13
Add Top 100 Bahasa words

rhesamu:
Hi, I just wanted to say I really like your project. I'd like to add top 100 commonly used words in Bahasa language as an option.

Cheers! :)


Miodec:
Just did a little google search and found that bahasa can mean many languages? It looks like maybe this should be called Malay? I dont know, would be awesome if you could explain it.

rhesamu:
@Miodec good findings, I could change it to Indonesian instead to create less confusion 👍

Issue #114, opened 2020-06-04
Custom theme


Miodec:
Be able to define all colors of a theme with a color picker

bhomie:
Going to work on this.

PR #147, opened 2020-06-17
Custom Themes

bhomie:
**Currently you can:**
- Change tabs for custom theme and preset themes
- You can change the values to actually change the theme
- Change back to preset themes after setting custom colors
- Able to save custom theme to user profile
- Default tab is open based on if theme was set or not

**Still to be done:**
- Make the initial values of the selectors match the current theme

**Known Bugs**
- If a theme has custom values set outside of the main variables, they will not be overridden. (see dots menu for example)

**Feature Ticket**
- [ ] #114


bhomie:
@Miodec in order to support themes in this way I may have to re-factor how the are applied since we're adjusting it based on the DOM unless you can think of another way. Like generating a css file with the variables? not sure... thoughts?

Miodec:
Since we only change the stylesheet href - maybe create a custom.css theme that has the default serika dark colors, and then when applying the custom colors use something like this


https://css-tricks.com/updating-a-css-variable-with-javascript/

That way you override the values inside the file, and then can still go back to the presets

public/js/settings.js
```diff
@@ -401,4 +401,29 @@ $(document).on("click",".pageSettings .section.tags .tagsList .tag .removeButton
...
+                $('[tabContent="custom"]').addClass("reveal")
+            }, 250);
+        }
+})
+
+$('.colorPicker').change(e => {
+    var $colorVar = $(e.currentTarget).attr('colorVar')
+    var $pickedColor = $(e.currentTarget).val();
+
+    document.documentElement.style.setProperty($colorVar, $pickedColor)
```
bhomie, line 428: Right so that's what I'm doing here but I believe what's happening is it's setting the variable as an override versus actually changing the property in the respective CSS file.

bhomie:
@Miodec take a look? Updated the description to the current state. Pull the branch and test it out :3

Miodec:
Here is some more changes / fixed i suggest: 
image

- the `preset` / `custom` still using the wrong font for me, not sure if its the css compilation issue on my end
- change the bg color picker to use `sub` color instead of `main` - just a suggestion, at first it wasnt quite clear to me which color its going to use
- maybe make the background picker wide (or keep the same size but have empty space to the right of it) - this is so the error and extra error pickers are above each other
- fix the typo inside 'colorful mode' and remove 'options:'
- also set that paragraphs color to `sub` and remove margin

PR #156, opened 2020-06-21
Changed name placeholder text to username for registering an account

CampAsAChamp:
Not sure how to not include the .firebaserc as I had to change the project id and I didn't want to update the .gitignore

Let me know what to do in the future or how to fix this.


Miodec:
Im not sure how to merge only one file from a PR so ill close this, but make the change on my end and push it - thanks for it.

For the future i think if you just dont commit the firebaserc file it should be fine

PR #185, opened 2020-07-14
Word History Bug Fix For Blind Mode

OctahedronV2:
Words with the 'hidden' class have this class removed when looking at the words history in results


Miodec:
Not the fix for the issue as discussed on Discord

PR #272, opened 2020-08-10
Remove uneeded tag and added matrix display for keymap

dmartin-tech:
Simple addition of a matrix view for the keymap display, as well as removing the [`</sup>`](https://github.com/Miodec/monkey-type/blob/51c8ad6932273227d8b341706595e6c9cebb8ba2/public/index.html#L566) that was uneeded
image
image


Miodec:
Ill suggest some changes: for the ortho layout, it would be awesome if you could figure out a way to hide some of the key on the right side, and make it a nice rectangle. And maybe make the spacebar smaller (kinda try to match the plank layout).

I would also change the config element from keymapMatrix to keymapStyle, and make it hold string instead of a boolean. That way we can have more in the future like staggered, split, ortho, ergo and so on.

PR #288, opened 2020-08-13
fix colemak-dh and dhm bottom left row

zigotica:


zigotica:
OK, apparently it's not really a bug, but the defaukt positions of ANSI vs ISO

Miodec:
Yeah, ill change the dhm to dh_matrix and swap x and z. In the colemak_dh no changes are needed

PR #335, opened 2020-08-31
use rounded accuracy in post-game summary

g-harel:


g-harel:
Now that I think of it, it would probably be better to only show 100 if there are no errors. I can make that a special case, but I'd like to test that more thoroughly tonight.

Miodec:
Thats why the accuracy is floored and not rounded

g-harel:
So what do you think, is this worth changing? Most other metrics are rounded so it was a little surprising, but I can see the logic.

Miodec:
Yeah i would just leave it as it is.

PR #340, opened 2020-09-03
Add Numbers as Option 

kunal-rp:
Just like punctuation, allow user to add in numbers into test.

For simplicity sake, kept chance of number generation for each word @ 10% and number range between 0 - 1000


Miodec:
Did some fixes, you missed a few things. Will merge soon.

PR #344, opened 2020-09-07
Add read ahead mode.

typerqeo:
Add a new behavior mode that hides the next two words.


Miodec:
Converted it to a funbox. Thanks!

PR #350, opened 2020-09-10
added Hebrew support

typerqeo:
Added support for Hebrew, our first right-to-left language.

Credits to Dan Malachi for the word list.


Miodec:
Need to read right to left. Just gotta change the flex direction, all tho I’m not sure if it will work correctly with line scrolling

typerqeo:
I can try this

typerqeo:
Alright, it should work now.

Issue #390, opened 2020-10-01
Typography Symbols are not mapped to universal keyboard standards.


RebonackAl:
Common typography type sets, such as:

> “ (\&ldquo; \&#8220;)
> ” (\&rdquo;	\&#8221;)
> ’ (\&lsquo;	\&#8216;)
> ‘ (\&rsquo;	\&#8217;)
>  , (\&sbquo;	\&#8218;)
>  — (\&mdash;  \&#8212;)
>  … (\&hellip; \&#8230;)

remain when entered in custom text. Example: For the opening quote mark “ a user would have to enter Shift+Option+[ instead of the universal quote ".

It is very common for ebooks to have this correct typography.
The use case it breaks is a user wanting to enter in text from an ebook or source that has correct typography, perhaps to practice touch typing or improve speed, whilst also reading something they enjoy.

Miodec:
Merged

PR #392, opened 2020-10-01
[#390] Typography Symbols are not mapped to universal keyboard standards.

RebonackAl:
Replaces specific typography characters with universal standards.
This is available to the user by a default checked checkbox.

The option is given to the user rather than always automatically applied because:
a) I'm not 100% sure other languages or keyboard layouts do not use these typography symbols as standard inputs.
b) There may be some who enjoy the challenge of entering Shift+Option+[ just to get a fancy quote mark.

enter text


Miodec:
Nice, thanks

PR #399, opened 2020-10-05
Polling timer approach

willerbee:
Would be interested to know your thoughts about this approach with the timer. AFAIU recursive is alright when the function is simple. Thinking of die hard typers typing at least 3600 seconds means 3600 more depth in the call stack (I'm aware the current code has already been proven even with 24hour sessions! :D Still good to have a bit more stability/efficiency in memory). All good if this gets closed, but still keen to learn your thoughts on why recursion was used :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@
 
 ## Standards & Conventions
 
-1. Use a SCSS compiler, and make sure that a `style.mis.css` file is generated in the same directory as the `style.scss`. For VSCode I recommend `Easy Sass` extension by Wojciech Sura. After installing it, to configure it:
+1. Use a SCSS compiler, and make sure that a `style.min.css` file is generated in the same directory as the `style.scss`. For VSCode I recommend `Easy Sass` extension by Wojciech Sura. After installing it, to configure it:
 
     - Navigate to the VSCode settings, `Extensions` section and find `Easy Sass configuration`
     - Under the `Formats` group click `edit in settings.json`
--- a/public/js/script.js
+++ b/public/js/script.js
@@ -72,6 +72,7 @@ let keypressStats = {
 
 let errorSound = new Audio("../sound/error.wav");
 let clickSounds = null;
+let timerPollRateHz = 10;
 
 function initClickSounds() {
   clickSounds = {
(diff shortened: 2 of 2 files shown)


Miodec:
So, the self adjusting timer was not written by me, but I kinda understand the whole idea. However I don't really understand the advantages of this approach? Could you explain that a bit?

willerbee:
I used to have an old reference that I can't find now. This link isn't the exact same but the point is somewhat similar from what I can remember (lol i'm old, been writing code since 1997). https://www.codespot.org/javascript-solving-problems-recursive-iterative/

Basically, the advantage of this approach is it keeps on polling the time and do action when it's "time to fire". The previous one uses a self adjusting timer where it keeps on calling its function every estimated "time to fire". This is a good approach and a bit more efficient by estimating when to trigger. However it also needs to implement recursion which is expensive in memory (mostly produces stutter specially in long tests as chrome or the OS paging to make more memory available).

typerqeo:
For clarification, Ennui did his 24h test before the self-adjusting timer was introduced in https://github.com/Miodec/monkey-type/pull/352. That PR explains the motivation somewhat. Admittedly, I haven't tried letting it run for 24 hours, and I don't know what will happen if one does attempt a >~1 day test.

I'm all for optimizing memory usage, but I don't think that's needed here. The call stack doesn't appear to grow with the current approach because it's not your typical recursion but rather recursive callback scheduling. Here is a SO answer that supports this: https://stackoverflow.com/questions/24631041/why-the-function-called-by-settimeout-has-no-callstack-limit

It's possible that you're seeing unbounded stack growth in Chrome's debugger since async stack traces are enabled by default. In that case, all you have to do is turn it off.

disable async stack traces

Then, the call stack looks shallow.

call stack

I would assume, though I haven't checked, that Chrome only keeps track of the async call stack if the F12 window is open. Have you tried profiling? It would be nice to see some data that indicates excessive memory use.

Comparing the implementation in this PR with the current implementation, the boundaries of each measured second are less likely to be precise with polling. You'd have to increase the polling frequency, or you may easily be ~100 ms off at 10 Hz. Bumping it up to 50 Hz would get you closer (but probably not quite) to the precision of the current approach.

It definitely is worth exploring other approaches to timing because the current implementation is not yet perfect, so thanks for that.

willerbee:
Thanks @typerqeo, yes I agree that the current implementation is more efficient in timing vs bumping the frequency in this PR. I appreciate your time shedding some light. I'm happy to have understand that there are no worries regarding the exceeding the call stack size (it's not even remotely possible after realising!). Thanks for that. Closing this PR now

PR #410, opened 2020-10-09
Fixed missing 'on' button text in settings under blind mode

theamanjs:
I've described this issue in detail [here](https://github.com/Miodec/monkey-type/issues/409) in detail. Nevertheless, this pull request is about fixing the missing `on` button text in settings under blind mode.
Screenshot_2020-10-09_08-29-36

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1556,7 +1556,7 @@
                   off
                 </div>
                 <div class="button on" tabindex="0" onclick="this.blur();">
-                  ⠀
+                  on
                 </div>
               </div>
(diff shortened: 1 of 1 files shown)


willerbee:
I believe that is the point of that option @theamanjs, there's no "on" because you can't see it when you're blind

willerbee:
My bad @theamanjs, there's actually a braille icon on that option.
image Perhaps it just didn't render on your end.

Miodec:
Intentional. Not a bug.

PR #418, opened 2020-10-11
Feature: option to use single list command line

willerbee:
Current command line is good, but there are also use cases where a single list is very useful. (e.g. when searching for a command that doesn't get exposed because it is burried in a submenu.) This proposed feature will not hinder existing workflow(i hope) as it is triggered manually by default in the command line by typing **>** and then your search term.

Sample commands:

- Switching to 8008 theme, type **>8008**
- Changing to Nunito font, type **>Nunito**
- Changing to words mode, type **>words**
- Changing custom word count, type **>custom word**

You get the idea. It's basically a consolidation of commands that can be enabled or triggered.

Early version demo here: https://discordapp.com/channels/713194177403420752/713196019206324306/764527921670389821


Miodec:
Amazing work!

PR #474, opened 2020-10-26
Keymap null check and layoutfluid saves layout

dmartin-tech:


public/js/script.js
```diff
@@ -45,7 +45,7 @@ let verifyUserWhenLoggedIn = null;
 let modeBeforePractise = null;
 let memoryFunboxTimer = null;
 let memoryFunboxInterval = null;
-
+let memoryFunboxKeyboardLayout = config.layout;
```
dmartin-tech, line 48: Not sure if I should instead assign this to "default" since it would be changed later when layoutfluid is selected

dmartin-tech:
Gonna rework it i just remembered the funbox is a cookie thing

public/js/userconfig.js
```diff
@@ -37,6 +37,7 @@ let defaultConfig = {
   flipTestColors: false,
   capsLockBackspace: false,
   layout: "default",
+  layoutBeforeLayoutfluid: "default",
```
dmartin-tech, line 40: rename to`savedLayout` maybe?

dmartin-tech:
This time I actually fixed #281 hooray

PR #490, opened 2020-10-28
Improve support for international (ISO) keyboards

stevep99:
The purpose of this patch is to allow support for the "ISO key" - that's they key between left shift and Z on an ISO keyboard. ISO keyboards are the default in European countries but in the current version of the app, this key cannot be mapped to anything. 

A side effect of this fix is that the key lookup system is simplified. 

Detailed explanation follows:

The existing code essentially does a two-step process to lookup a key:
1. keyboard event code  -->  qwerty mapping
2. qwerty mapping --> selected layout mapping.

A consequence of this approach is that keys cannot be mapped unless they exist in the default ANSI qwerty set. For example, the key event for the extra ISO key mentioned above is "IntlBackslash", but it's not possible to add this as there is already a mapping to Backslash.

To avoid this problem, this patch simplifies the key lookup to a single step process:
keyboard event code -->  selected layout mapping

The existing Qwerty defaults ("qwertyMasterLayout") are replaced by a simple array, keyEventCodes, which defines the list of keys that should be in the layout mapping. Notice I have added IntlBackslash just before Z. Consequently, I have also added a mapping for this key (usually the backslash character) to each of the layout mappings. On ANSI boards, this key doesn't exist, so the event won't be called. As an aside, on the Azerty keyboard, this extra key is actually "<>", so a side effect of this change is that Azerty support is improved too.

Now, coming on to my motivation for this patch in the first place, I wanted to fix the [Colemak-DH ](http://colemakmods.github.io/mod-dh/gfx/colemak_dh_keyb_iso.png)and [Colemak-DHk](http://colemakmods.github.io/mod-dh/gfx/colemak_dhk_keyb_iso.png) layouts, as they should use the [angle mod](https://colemakmods.github.io/ergonomic-mods/angle.html), and on ISO keyboards, the extra ISO key should be mapped to Z, So I have fixed those too. For ANSI users, there should be no effect as the IntlBackslash event won't be triggered, and in the Colemak-DH layout files, Z is still mapped to the Qwerty B key as before, so ANSI users should not see any change.


Miodec:
Is it possible that there are some ISO users who use the colemak dh layout override now, and merging this will cause their bottom row to be shifted?

Also on another note - keymaps will need to be updated to somehow support ISO too.

stevep99:
The current implementation of the Colemak-DH layout override on the site is incorrect. Fixing it was my main motivation for this PR.

The keymaps are already updated to support ISO in the PR. I have done it such a way that ANSI users should see no difference, but ISO users will be able to use their ISO key. 

Feel free to test the patch with an ANSI keyboard to be sure though, as I only use ISO.

nickcoutsos:
I'm very new to colemak, all of the possible mods and recommendations for non-staggered layouts is very confusing but on an ANSI keyboard this change has really confused things for me.

How should ANSI/ISO be handled by the app with layout mapping or display?

stevep99:
@nickcoutsos  What do you mean by layout mapping or display? The app doesn't show the layout on screen AFAIK.

Thanks for the merge, just tried it at https://monkeytype.com/ and seems to be working great!

Miodec:
See, i knew something would go wrong with keymaps, I just couldn't figure out what. Today I found it:

Screenshot 2020-10-30 at 12 44 01

Staggered keymap is now misaligned.

Miodec:
Probably just need to add ansi/iso keymap setting to show/hide the iso key

stevep99:
Ah, OK, I see. I didn't realize there was a visual representation of the keyboard. 

I'm happy to take a look at that and put in a second PR to correct it.

stevep99:
I put in a [PR to fix this](https://github.com/Miodec/monkeytype/pull/505/commits/da3c20b9bc541f389409d863fcc3add11f65ea3e).

PR #502, opened 2020-10-30
Remove private methods from simplePopup (fixes in Firefox, Safari, Internet Explorer)

### LINKED ISSUE #503: Site doesn't load at all in Firefox (and probably also not in Safari or Internet Explorer)
Firefox gives the error "private fields are not currently supported", which points at `misc.js` - it looks like simplePopup which was recently added uses private fields. This is a syntax error, so it doesn't load `misc.js` at all, which breaks everything else.
Screenshot_2020-10-29_23-57-05

cincodenada:
This fixes #503.

The added simplePopup had private methods, which totally breaks Firefox and Safari since it's a syntax error and this file then doesn't load at all. We can simply make these methods public, there's no reason they need to be private to work, it's just a code organization thing.

### DIFF
--- a/public/js/misc.js
+++ b/public/js/misc.js
@@ -510,7 +510,7 @@ class SimplePopup {
     this.element = $("#simplePopup");
     this.buttonText = buttonText;
   }
-  #reset() {
+  reset() {
     this.element.html(`
     <div class="title"></div>
     <input>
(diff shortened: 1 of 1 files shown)


dmartin-tech:
Works perfectly

PR #505, opened 2020-10-30
Fix incorrect bottom row offset of onscreen layout

stevep99:
So, as a result of my [previous PR](https://github.com/Miodec/monkeytype/pull/490), the ISO key now works, but this had a knock-on effect of offsetting the visual representation of the keyboard.

This PR is a quick fix for that issue.

What I have done is move the ISO key to the last entry in the list of keys, so that it doesn't insert itself into an unexpected place. This keeps support for the key, but doesn't mess up the graphic, as the keys in the layout file are in the same order as before.

This means the ISO key will not be appear in the layout graphic - only the ANSI keys will appear - but the ISO key will still register when typed.

Ideally, a better solution would be to add the new key to the graphic (i.e. add the extra key at the start of row4 in the html, and adjust the offset accordingly). However, this approach requires recompiling the scss, which I'm not familiar with. Hence this simpler solution instead.


Miodec:
Ive got a better solution where the bottom row is actually shifted depending on the layout, which should be ready soon - so I'm going to close this one

PR #515, opened 2020-10-31
Added new language

JosueLeovani:
Spanish most common words 1k, it really has 1033 words, is that ok?


Miodec:
Make sure to name the language `spanish_1k` inside the file and also to edit the `list.json` file.

PR #519, opened 2020-11-01
Fixed switching back to english when exiting quote mode

Cosmeeeen:
As referenced in [this](https://github.com/Miodec/monkeytype/issues/475) issue, I found the line that caused the switch to English when entering quote mode, and commented it as for it not to force the user in English mode. From my understanding, this won't affect the quotes, as they are all pre-written in English. 
As I tested it, the fix worked for the problem, but if it somehow conflicts with the scoreboard, I'm ok to it not being accepted.
I didn't notice the language in which a test was taken anywhere in the recent tests scoreboard. 

_Thank you for the opportunity, I love your project btw 😄_

### DIFF
--- a/public/js/script.js
+++ b/public/js/script.js
@@ -3075,7 +3075,7 @@ function changeMode(mode, nosave) {
     $("#top .config .numbersMode").addClass("hidden");
     $("#result .stats .source").removeClass("hidden");
     $("#top .config .quoteLength").removeClass("hidden");
-    changeLanguage("english", nosave);
+    // changeLanguage("english", nosave);
   }
   if (!nosave) saveConfigToCookie();
 }


Cosmeeeen:
I can't tests how the scores are saved, mostly due to the fact that I can't set up a proper user page for the local build. (I'm quite of a noob to this stuff, so sorry for that) 
I managed to create a login with firebase, but the use page still won't load. 
It would be nice if we could find some documentation as to how to set it up on the local version, in case more newbs like me would like to start contributing on more complex issues. 
Thank you!

Miodec:
The problem is that we have to switch the language to English because the quotes are in English. Plus this will make it easier to filter quotes by language when more quotes in different languages are added.

kevinjihwanlee:
It's unclear to me why the language has to be switched to English upon mode switch to quote when the loading of the English quotes are hardcoded. Leaving the language as is would in fact make it easier to filter quotes by language later. From my local testing, the only bug appears to be the language display under Test Type, which can be hardcoded to English until multilingual quote support is added. Is there anything else I'm missing (maybe in score saving? still relatively new to the codebase)?

Miodec:
Beacuse language is still saved to each result, no matter if you're currently doing a quote. So, not forcing to English might lead to results being in quote mode and having the wrong language.

Miodec:
This means that if we ignore the language thing for now (and people end up doing english quotes while their language is set to Spanish), when multi language quotes are introduced, some past results will be in the wrong language.

kevinjihwanlee:
Is it possible to hardcode the language when saving results for quotes specifically as a temporary solution? That seems like a good temporary fix until multi language quote support arrives.

Cosmeeeen:
From my understanding of the code, there is a problem with the nosave parameter of the changeLanguage function. That (from my understanding) is supposed to make the language change temporary. I'm looking into it slowly, but my college classes kinda get in my way for giving it full focus. So don't assign me this issue yet, but I'll update if I find a possible fix in the little time I have.

Miodec:
@kevinjihwanlee Yeah, this solution will work

PR #522, opened 2020-11-02
Adding a build process to MT development

typerqeo:
Lots of changes to MT's project structure and development workflow! If you don't contribute to MT, you can ignore this PR.

- To run MT you must now build the code. This can now be done with `npm run build`. However, it's even better to use `npm run start:dev`, which will launch a dev server that automatically builds whenever the code is changed.
- The build generates the entire `dist/` directory.
- Monkeytype JS and sass files have been moved to `src/`.
- Firebase now hosts `dist/` instead of `public/`.
- The build performs four simple steps:
    1. clean - removes the `dist/` directory
    1. static - copies all files from `public` to `dist/`
    1. sass - builds sass in `src/sass/` and outputs to `dist/css/style.css`
    1. cat - concatenates all code in `src/js/` and outputs to `dist/js/monkeytype.js`
- Each step can be run independently with `npx gulp <stepname>`, or you can run them all with `npx gulp rebuild`. You can also use `npx gulp build` to run only the last three steps.
- This means, besides dependencies, there is now a single `monkeytype.js` file for clients to download. I haven't tested the performance gain, but there *probably* is some performance gain. It's currently unminified to allow for in-browser debugging.
- The build is made using [Gulp](https://gulpjs.com/) as a stepping stone towards [Webpack](https://webpack.js.org/) (not in this PR). I would like to switch to Webpack in the near future, but that proved to be too much for one PR. In any case, these two build tools can coexist during the transition.
- Updated CONTRIBUTING.md.
- Many dev dependencies added for building.

@Miodec , you might have to make some deployment changes to run the build, like adding an `npm run deploy` script for instance.


gulpfile.js
```diff
@@ -0,0 +1,46 @@
...
+  return src("./public/**/*").pipe(dest("./dist/"));
+});
+
+task("clean", function () {
+  return src("./dist/", { allowEmpty: true }).pipe(vinylPaths(del));
+});
+
+task("build", series("static", "sass", "cat"));
+
+task("watch", function () {
```
Miodec, line 42: Im guessing this is the line that runs the 'build' command on save of any file

Miodec:
Im also thinking maybe rename the gulp `rebuild` command to something else since its called by a 'build' npm command, which is quite confusing at first - unless its some sort of standard.

Miodec:
The `monkeytype.js` file can remain unminified - Cloudflare minifies js files anyway.

gulpfile.js
typerqeo: Specifically, L43 does the watching. L42 adds the command `gulp watch`.

typerqeo:
Gulp tasks have been renamed.

PR #540, opened 2020-11-06
Add Iceberg Dark and Iceberg Light themes

gkeep:
Port of [iceberg.vim](https://github.com/cocopon/iceberg.vim).

Before you merge though, is there any way I can override the color for time/word counter? I find the automatic one not distinct enough on the dark variant.

Previews:
light
image

dark
image

(screenshots updated to reflect latest commit)

### DIFF
--- a/static/themes/iceberg_dark.css
+++ b/static/themes/iceberg_dark.css
@@ -1,8 +1,8 @@
 :root {
   --bg-color: #161821;
   --caret-color: #d2d4de;
-  --main-color: #d2d4de;
-  --sub-color: #84a0c6;
+  --main-color: #84a0c6;
+  --sub-color: #c6c8d1;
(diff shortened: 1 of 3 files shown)


Miodec:
Timer and progress colors are controlled via the settings

gkeep:
Oh yeah, colorful mode. Totally forgot about it while testing on localhost. Ready for merge!

Miodec:
The only thing is that the dark theme here is very similar to `nord`, but I can merge the light version and call it `iceberg`.

gkeep:
> The only thing is that the dark theme here is very similar to `nord` or `drowning`

I don't think so. The thing that pushed me here is that `nord` (which I was using before) is not very contrast-y. For me, Iceberg is a lot easier on the eyes, because of that contrast.

PR #555, opened 2020-11-09
Added result options to commandline

decafff:
Added result page options to the commandline, so you don't have to take your hands off of the keyboard after a test.

### DIFF
--- a/src/js/commandline.js
+++ b/src/js/commandline.js
@@ -614,6 +614,44 @@ let commands = {
         window.open("https://discord.gg/monkeytype");
       },
     },
+    {
+      id: "repeatTest",
+      display: "Repeat test",
+      exec: () => {
+        if (resultVisible) {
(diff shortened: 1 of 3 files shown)


decafff:
Added onedark theme from atom as well

src/js/commandline.js
```diff
@@ -614,6 +614,44 @@ let commands = {
...
+        if (resultVisible) {
+          restartTest(true);
+        }
+      },
+    },
+    {
+      id: "practiceMissedWords",
+      display: "Practice missed words",
+      exec: () => {
+        if (Object.keys(missedWords).length > 0) {
```
Miodec, line 630: We should have a `resultVisible` check here otherwise you could start it while the test is active.

decafff: Sure thing.

PR #576, opened 2020-11-11
Added available() method to commands to hide them from the list

decafff:


Miodec:
Just noticed that this change will not hide suggestions correctly if `single list command line` is enabled.

Issue #495, opened 2020-10-29
Refactoring the frontend code


typerqeo:
## Problems
1. Some files are large. `script.js` is especially massive at >5000 LOC.
1. The code is written in a mostly procedural style and has many globals. This makes it hard to follow and error-prone. There is also a lack of tests.
1. External dependencies are currently managed manually.

## Proposed Solutions
1. Let's keep files reasonably small and self-contained. We can use [Webpack](https://webpack.js.org/) to bundle the resources. This would be much nicer for both code organization and deployment.
1. Let's write new JS files in ES6 object-oriented style.
    - This makes it easier to write unit tests and more modular components.
    - The transition doesn't have to take place all at once. The code can be refactored one feature at a time. We should make sure to add unit tests at each step.
    - This might break some functionality in the short term but is undoubtedly good for code health in the long term. We may need to work on another branch for a while.
1. Now that there is a `package.json`, we can use npm to manage dependencies.

---

I'll try to put together a PR to kick off this effort, but I might not have the time to personally see it through to completion. Judging from the hours I've put into this already, this plan is pretty ambitious.

Help or feedback is welcome and appreciated!

Miodec:
Simmilar to #493, not sure which one to close

typerqeo:
Oh... maybe we should repurpose #493 to address the backend? The cloud functions can be dealt with independently from the frontend.

Miodec:
Yeah, sounds good

typerqeo:
Feel free to assign this to me.

ivanjermakov:
+1
I would like to contribute, if the code was cleaner.

Marco-Pagani:
seconded, I would like to help with the refactoring too. @typerqeo once you create a bit of code to establish some conventions feel free to delegate some work to me and I'll try to keep it consistent.

typerqeo:
Work done so far (because I forgot to tag them here earlier):
https://github.com/Miodec/monkeytype/pull/483 https://github.com/Miodec/monkeytype/pull/522 https://github.com/Miodec/monkeytype/pull/526 https://github.com/Miodec/monkeytype/pull/553

typerqeo:
With the above two PRs merged, the stage will be set for

1. converting internal dependencies to ES6 modules
2. converting external dependencies to npm

If you'd like to take on any work converting dependencies, feel free to announce it here. After these PRs are merged, I'll do some more converting if no one's signed up.

[...]

PR #581, opened 2020-11-13
Adding Browserify to the build process

typerqeo:
- Split up the set of JS files into two groups: refactored files and global (legacy) files.

- Changed the build process to be the following:

    1. delete dist/
    1. lint the refactored JS files
    1. concatenate and lint the legacy JS files, and output to dist/gen/index.js
    1. copy the refactored files to dist/gen/
    1. bundle the JS files together using Browserify, with the generated index.js as the entry point
    1. copy html and static files to dist/
    1. compile and copy sass to dist/css

- Made db.js into an ES6 module. I chose to start with this file since it's already pretty small and self-contained. The global `dbSnapshot` was replaced with getter and setter functions. Eventually all of the JS files should be made into ES6 modules.

- Added global-dependencies.js, which imports all the dependencies that the legacy files use. These dependencies are both external (from npm) and internal (from refactored JS files).

- Added exports.js to expose any debug info you'd want to have in the browser console. In this PR, we're exporting a `snapshot()` function which exposes the `dbSnapshot`.

- Added Babelify to use ES6 syntax with Browserify. This adds more lines to the generated JS, which is not good, but it's possible to reduce the size eventually.

This PR is part of the refactoring effort in https://github.com/Miodec/monkeytype/issues/495


Marco-Pagani:
you mentioned in #495 using webpack, is building with browserify an intermediate step so that you can keep the gulp file for now? Or do you plan to stick with this instead of webpack?

typerqeo:
Yeah, I'd like us to use Browserify at least until we can get rid of Gulp (specifically, the JS concatenation part). Then we can switch to Webpack more easily.

src/js/account.js
```diff
@@ -284,7 +284,7 @@ firebase.auth().onAuthStateChanged(function (user) {
       verifyUser(verifyUserWhenLoggedIn).then((data) => {
         showNotification(data.data.message, 3000);
         if (data.data.status === 1) {
-          dbSnapshot.discordId = data.data.did;
+          db_getSnapshot().discordId = data.data.did;
```
Miodec, line 287: Will this make a reference to the object or a copy? Im thinking if setting `discordId` will work like this.

typerqeo: `db_getSnapshot()` should return the reference (by value), so setting a property of the snapshot should work. It's similar to

```
const snapshot = db_getSnapshot();
snapshot.discordId = data.data.did;
```

Note however that assigning the snapshot from the getter (`db_getSnapshot() = otherSnapshot;`) won't work. That's why we also have `db_setSnapshot(otherSnapshot)`.

Issue #579, opened 2020-11-12
Refactoring dependencies to modern practices


Marco-Pagani:
This may be better treated as part of #495 so just let me know if so.

As part of refactoring the frontend code I believe it would be helpful to switch from static js files to npm, that way dependencies can stay updated and be centrally managed and easily understood.

As part of this I would also like to switch from moment.js to [date-fns](https://date-fns.org/). Moment is no longer being actively developed and is considered legacy, and it also a very heavy library for the few things it's used for. Chart.js is removing moment as of V3 but currently allows you to replace it via plugins, so I can do that as well as replace moment in the frontend code.

I think this along with the other refactoring will go a long way to making the project more maintainable and easier to approach for new users.

Marco-Pagani:
I can handle this myself, so if you think it makes sense feel free to assign me

typerqeo:
Any updates here @Marco-Pagani?

Marco-Pagani:
Sorry, had some other projects to attend to this week. The bulk of the work is done here, but now that #582 is merged I need to rebase and make sure my commits behave before I make a pr. Will update soon

PR #582, opened 2020-11-13
Converting ChartJS to an npm dependency

typerqeo:
This PR is dependent on https://github.com/Miodec/monkeytype/pull/581, so please merge that one first. Despite looking rather long, this PR will be quite short once the preceding PR is merged.

- Made the ChartJS dependency managed by npm and bundled with our JS code.

- We should be able to convert other external dependencies similarly. Would you like to take that on, @Marco-Pagani?

This PR is part of the refactoring effort in #495 and #579


Marco-Pagani:
I can definitely help converting the other external dependencies. Let me look at your changes on this issue and I'll try to follow the same format

Marco-Pagani:
I converted the other libraries to npm packages in 1077ace. It needs to be tested more but it builds and runs for me. I have some trouble with importing jquery using es6 imports but it's still much better than before.

PR #601, opened 2020-11-19
Adding jquery gsap plugin

typerqeo:
This is a monkey patch that hopefully makes animations smoother at high refresh rates. It simply overrides `jQuery.animate()` with gsap, which claims to be up to 20x faster than jquery. See example speed test https://greensock.com/js/speed.html

The gsap jquery plugin is discontinued. For a more permanent solution, I think we should switch to css animations or (non-jquery) gsap.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -3430,6 +3430,8 @@
   <script src="js/jquery.color.min.js"></script>
   <script src="js/easing.js"></script>
   <script src="js/jquery.cookie-1.4.1.min.js"></script>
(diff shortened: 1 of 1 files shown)


typerqeo:
@Marco-Pagani fyi

Marco-Pagani:
sounds like a good drop in improvement, since it's not on npm it'll have to remain a static file for now. In the long run I could look into converting the existing animations to gsap format so we can eliminate it.

PR #606, opened 2020-11-20
Disable 'next' keymode when memory funbox mode is active

a1usha:
When activating *memory* funbox mode, keymap mode sets to *react* if it was set to *next*. When keymap *next* mode activated after *memory* mode activation, it also sets to *react* mode automatically.


a1usha:
Also added 1k rusian words

PR #621, opened 2020-11-23
Add the most common Filipino words as language source.

ranelpadon:
I've created the [Ngram Type](https://ranelpadon.github.io/ngram-type/) typing trainer which my friends and I use regularly (and some Colemak users as well). Recently, I added the **Top 100 Filipino Words** as one of the training **Source**. So, it would be great if we could add this also in MonkeyType since I use it regularly as well. `Amaranth`, one of the top typists in your site, is from Philippines also. Thanks :)


Miodec:
Is there any chance you could get the list to include 200 words instead of 100? This would make the words repeat less.

ranelpadon:
@Miodec I've tried searching before but the list I found are questionable, and it seems there's no official one. So, I used the one on this list, since from my experience, they're indeed very common:
https://learningfilipino.com/100-most-commonly-used-filipino-words/ 

But, I've tried again now and found a good/promising list which is based on **Zuraw 2006** corpus/papers:
> The following set of flashcards rank the most common words used in Tagalog/Filipino based on a UCLA corpus of Tag/Fil texts (Zuraw 2006). Excludes English terms adopted into Filipino. Conjugated verb forms have been grouped together (e.g. magsalita, nagsalita, salita) but paired only with the English infinitive (e.g. to speak) translation.

https://quizlet.com/23822693/tagalog-1-50-1000-most-common-words-flash-cards/

Skimming the words, it seems to be the common ones indeed, and kinda consistent with my `Top 100` list previously. Will get back to you later today with the `Top 200` list. Thanks!

ranelpadon:
@Miodec Updated, contains now the **Top 200** lists (https://github.com/Miodec/monkeytype/pull/621/files) as per this commit (https://github.com/Miodec/monkeytype/pull/621/commits/ef75478f55b51b5f5a9288a77bef62ae209c65d0). 

Pushed also the changes to my [Ngram Type](https://ranelpadon.github.io/ngram-type/) website, changed from **Top 100** to **Top 200**, hence verified to have no syntax/punctuation errors. 

Thanks

PR #627, opened 2020-11-25
Punctuation mode updated

vuong-buihv:
Added
  single quotes ''
  parentheses ()
  braces {}
  brackets []
to the punctuation mode


Miodec:
Im not sure if i fully agree with adding `{}` and `[]` to the punctuation mode as they are not really commonly used while typing day to day.

vuong-buihv:
{} and [] are now removed

PR #641, opened 2020-11-28
Revert "added jquery gsap library to make jquery animations more perf…

typerqeo:
…ormant"

This reverts commit a0af701f60f605c6793cbdfc32bba2713a88c880.

---

Seems this is affecting animation behavior in notifications. Let's revisit this later using a more principled approach.


typerqeo:
@Marco-Pagani fyi

PR #648, opened 2020-11-30
Update index.html

Andororo:
Hello,
I don't know if this is a bug or not. But I notice in the "blind mode" setting there wasn't an "on" print when toggled


Miodec:
Not a bug, button is blind :)

PR #650, opened 2020-11-30
Add 1k common word list

bataak:


Miodec:
I would just rename it to 10k for consistency

PR #653, opened 2020-12-01
added greek

Vitaman02:


Miodec:
Make sure to add it to the `list.json` file

Vitaman02:
I added it below german :)

PR #659, opened 2020-12-02
Java Language Word List

LeSirH:
Let me know if there are any words I might have missed or should remove.


Miodec:
Please rename to `code_java` (including the name inside the file) to make sure the file will be found. Currently it wont work.

PR #661, opened 2020-12-04
Convert external dependencies to npm packages

Marco-Pagani:
This is related to #579. All static libraries have been replaced with npm packages


Miodec:
Getting `$ is undefined` after merging. Im guessing because the `global-dependencies.js` is loaded after the refactored files? All though i tried moving that file into the `refactoredSrc` array and I'm still getting this error.

Miodec:
Ok i think this is the correct fix: edbc845def961172573972b21d2d41f0e78277c3. Let me know if this is not the way.

Miodec:
Nevermind, it compiles fully, but is broken on runtime. 

image

Miodec:
Reverted for now so i can continue working on other things

Marco-Pagani:
I will look into this further and let you know if I can find out what's going on. It built fine for me when I made the PR

PR #689, opened 2020-12-09
Add missing on text to blind option

mlrv:


Miodec:
Not a bug - the button is blind.

PR #706, opened 2020-12-12
Added 1k words in filipino language

KungPaoChick:


Miodec:
Im pretty sure japanese will not work unless you tested it

KungPaoChick:
> Im pretty sure japanese will not work unless you tested it

Yeah, you're right. It was dumb, so I just deleted it just now. Only the Filipino one remained.

Miodec:
You commited `deno.yml`, please remove. Also make sure to add your language to the `list.json` file.

PR #711, opened 2020-12-14
Added swedish, swedish_1k

minerscale:
Added Swedish and Swedish 1K. Data from here: https://www.101languages.net/swedish/most-common-swedish-words/


minerscale:
Ran the list through someone who knows Swedish and they found some non Swedish words, so I took them out and filled in the rest with more words.

PR #715, opened 2020-12-15
Added Yoruba 1k

joshuamango:


Miodec:
Please dont prettify `list.json`. Commas are first to avoid merge conficts.

joshuamango:
Apologies. Reverted the formatting in `list.json`.

Miodec:
Oh, also just realized that you've got spaces in words - this will not work with the current input implementation I'm afraid :< Right now every time you press space I'm moving to the next word, meaning if you have spaces inside a word, you won't be able to type the whole thing.

joshuamango:
That makes sense. I removed words with spaces and some duplicates. Although, the number of words is now ~730. Is that okay or should I try to get some more/or change the name to just "yoruba" (without the 1K)?

Miodec:
Thats fine. You can add more later if you find any, to bring it closer to 1k.

Miodec:
Also the name in the language json doesnt match the filename - wont work.

PR #728, opened 2020-12-17
Added Māori 1k

ariebh:


ariebh:
The list is slightly shorter than 1000 words as the [source](https://tereomaori.tki.org.nz/Teacher-tools/Te-Whakaipurangi-Rauemi/High-frequency-word-lists) contained some multiword expressions which were removed.

Miodec:
I think the `ā` character will break the regex check for sure and possibly other stuff. Could you just replace with with an `a` ?

ariebh:
Unfortunately no, the tohutō (the little bar over vowels) is essential in te reo Māori. It's often used to designate plurals and some words hold completely different meaning depending on it's presence; Keke translates to cake, whereas kēkē translates to armpit. 

The tohutō is often omitted by the ignorant/lazy, so it being available in Monkeytype, where one can practice reaching for the correct modifier key to produce the symbol, would serve well to mitigate bad practices from being formed. 

I didn't notice any errors in my brief play around on the front end, but I didn't look very deep into the codebase prior to this change. Could you let me know where/ how I could check for regex compatibility?

Miodec:
I meant just for the filename. Words will be fine

ariebh:
Ah, oops. Sorted now.

PR #735, opened 2020-12-18
Added list of commonly misspelled english words

ariebh:


Miodec:
Im not sure why add a commonly misspelled language if everybody's commonly misspelled words will be different?

ariebh:
True, a words list that is uniquely built for each user based would be better, but much like the 1k and 10k lists I think there is still merit using a list built from a corpus with numerous authors. This list has been sourced from [Wikipedia's list of common misspellings ](https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings) and then thinned down to only words that already appear in the 10k most frequently use english words list.

If you look through the list, regardless of how good you are at spelling, you will see words that trip you up in there. The number of false positives will vary for each user, but will still provide an increase in difficulty from the base lists.

Miodec:
Hm, maybe it should be at least renamed to something that indicates its an English list - something like `english_hard` or something.

UnrealApex (changes requested):
I would suggest changing the file name `commonly_misspelled.json` to `english_commonly_misspelled.json` because there is no context of which language the words are for in the file name. Also do the same in the `list.json` file and resolve the current merge conflicts.

Miodec:
Added in #1694

PR #751, opened 2020-12-22
Ignore keys with Alt modifier

SeerLite:
Hi!
Firefox uses Alt for a number of shortcuts, like switching tabs with Alt + N. This makes it so these shortcuts don't start the test.

### DIFF
--- a/src/js/script.js
+++ b/src/js/script.js
@@ -4936,6 +4936,7 @@ function handleAlpha(event) {
   if (/Numpad/.test(event.key)) return;
   if (/Volume/.test(event.key)) return;
   if (event.ctrlKey) return;
+  if (event.altKey) return;
   if (event.metaKey) return;
   event = emulateLayout(event);
 


Miodec:
This might not be a good idea since some languages may use alt for special characters

SeerLite:
Isn't that just AltGr though? Most programs have a lot of shortcuts bound to Alt, so it'd be weird if Alt alone was used for special characters.

Special characters with AltGr work fine on my layout with this change.

Miodec:
Well, looks like this change broke some support on mac devices. Its true that it should be using AltGr (and it is, on windows at least) but on Mac its actually using the Alt key, so special characters are unavailable. Gonna have to revert this change,

SeerLite:
I understand. On Mac there's no Alt key so they use the Option key, which gets interpreted as Alt.
What do you think about checking the OS and if it's Mac, allow Alt?
Or the other way around, checking if it's Linux (because it looks like it's Ctrl + N on Windows and Alt on Linux) and disabling it there?

PR #790, opened 2021-01-03
allow the screenshot feature in other browsers 

TayIorRobinson:
currently, if you visit the website on a browser that doesn't support the api's used by the screenshot api you just an error message or, you're not allowed to do this in firefox.

this pr allows you to open the image in a new tab if we fail to copy it to your clipboard.


Miodec:
Nice!

PR #803, opened 2021-01-06
Fix swapEscAndTab sometimes failing on Esc

SeerLite:
Introduced by my own #750, I missed the last `else`. This fixes it by wrapping it inside another `if`, I hope it's fine like that!


SeerLite:
I can't believe it I'm so stupid. By doing this I also completely undid #750. Sorry Miodec for doing causing such a mess while trying to get this small bit working :s
Also sorry for not testing it out correctly, it won't happen again (in case I do make another PR)

Now I'll try to fix this, again

Issue #370, opened 2020-09-24
Personal average pace caret


necdetaliozdur:
A pace caret that moves at personal average would be great for people who use monkey-type to gradually improve their typing skills. It would set a more reasonable goal than personal best.

Miodec:
Not available at the moment but you can set the pace caret to a custom speed

Jonny-exe:
Anyone working on this? I would like to do it. @Miodec @necdetaliozdur

Miodec:
I believe nobody is working on it. What would be your approach to it?

Jonny-exe:
I already implemented it.
I created a function called `db_getUserAverageWpm10` which I modeled after function `db_getUserHighestWpm`.
In `commandsPaceCaret` I added a new entry for average.
In `initPaceCaret()` I added an else portion to the if.

I tested everything locally, but for testing purpuses I set the average number to a hardcoded number.

I tried testing with firebase but when I try to register a user I get the following error back. `[ CONFIGURATION_NOT_FOUND ]`. How can I do testing with firebase?

Miodec:
@Jonny-exe make sure email authentication is enabled on your firebase web console

Jonny-exe:
Thanks that worked. I sent PR #809

PR #809, opened 2021-01-09
Added an personal average pace maker

Jonny-exe:
Now you can use your average speed as your pacemaker speed. #370 

I do it with the 10 last speeds because I think it is better than with all time scores. If you think that it would be better with all time scores let me know.


src/js/script.js
```diff
@@ -3931,6 +3935,14 @@ async function initPaceCaret() {
       config.language,
       config.difficulty
     );
+  } else if (config.paceCaret === "average") {
+    await db_getUserResults();
```
Miodec, line 3939: I think this is a problem since it will always get the user data again, when switching to the average pace caret. I think this should be moved istead of the `retval = 0` line inside `getUserAverageWpm10`, so that it only gets user data if no data was found.

Jonny-exe: Ok

PR #811, opened 2021-01-10
Welp... used up my sunday by making most frequently used korean words!

randomizermk1:
I would also like to add korean quotes like english. any pointers on that ?

I have a nice collection of korean classics ready to be implemented


Miodec:
Korean won't be fully supported because of the core method of input handling basically considers only single characters, and not groups like Korean needs. It needs a major rewrite, and when thats done ill be sure to merge this.

UnrealApex (changes requested):
I would suggest changing the file names to all lowercase because all other language json files follow this standard and are called that way.

ffactory-ofcl:
This PR title could be improved. I cannot edit it but would suggest `Add Korean, Korean_1k and Korean_10k`.  @UnrealApex

benwoodward:
> Korean won't be fully supported because of the core method of input handling basically considers only single characters, and not groups like Korean needs. It needs a major rewrite, and when thats done ill be sure to merge this.

Sorry to hijack thread, but just wanted to let people know that I'm building https://keeby.app to solve this problem. 

image

@randomizermk1 I can incorporate these words into the lessons in Keeby, obviously this isn't the right place to discuss so please tweet me if you have any suggestions: https://twitter.com/benvskorean

PR #832, opened 2021-01-13
Adding more polish words

jacekdobrowolski:
Added over 2k new words to polish from wiktionary.org most used words. Old worlds are still there just shuffled.


Miodec:
Please keep the original file ~200 words. If you want to create a larger words file, create a new one and name it something like `polish_1k` (1k meaning the amount of words in the list). Dont forget to add it to the `_list` file

static/languages/polish_2k.json
```diff
@@ -0,0 +1,2361 @@
...
+    "milion",
+    "reszta",
+    "artysta",
+    "byłoby",
+    "nerwowo",
+    "strzelać",
+    "treść",
+    "mąż",
+    "dać się",
+    "sierżant",
```
Miodec, line 14: Words with spaces inside of them will not be included in tests so make sure to split them up.

static/languages/polish_2k.json
```diff
@@ -0,0 +1,2361 @@
+{
+  "name": "polish",
```
Miodec: Name property must be the same as filename.

Miodec (changes requested):
You need to include the new language file in the `_list.json` file aswell.

jacekdobrowolski:
sorry for messy commits polish_2k  was added to _list.json in commit https://github.com/Miodec/monkeytype/pull/832/commits/37c171aacb9b010f26dccb9704b18d36f1fc9583

PR #833, opened 2021-01-13
add KAM 80's After Dark theme

FakeDazon:
a theme based off KAM 80's After Dark
image


Miodec:
Delete your public folder, its no longer used.

Miodec:
Also, the themes error color seems to be blending in a lot with the text. Errors should generally be really in your face, so you can spot them quickly.

FakeDazon:
i changed the error color to a yellow, and removed the public folder
image

PR #860, opened 2021-01-19
Duration not being passed due to which error notification remains till infinite

avi123nash:
Adding a language or a theme? No

Duration was not being passed to error notification which is why it'll stay for infinite time and needs to be closed manually

error_notif

### DIFF
--- a/src/js/leaderboards.js
+++ b/src/js/leaderboards.js
@@ -218,7 +218,7 @@ function updateLeaderboards() {
     })
     .catch((e) => {
       hideBackgroundLoader();
-      Notifications.add("Something went wrong: " + e.message, -1, 3);
+      Notifications.add("Something went wrong: " + e.message, -1);
     });
 }
 
--- a/src/js/leaderboards.js
+++ b/src/js/leaderboards.js
@@ -218,7 +218,7 @@ function updateLeaderboards() {
     })
     .catch((e) => {
       hideBackgroundLoader();
-      Notifications.add("Something went wrong: " + e.message, -1);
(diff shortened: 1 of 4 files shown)


avi123nash:
@Miodec Please review

Miodec:
Im not exactly sure whats this meant to do

src/sass/style.scss
```diff
@@ -710,7 +710,7 @@ a:hover {
   display: grid;
   gap: 1rem;
   position: fixed;
-  right: 1rem;
+  right: 12rem;
```
Miodec, line 713: Why this change?

src/js/leaderboards.js
```diff
@@ -218,7 +218,7 @@ function updateLeaderboards() {
     })
     .catch((e) => {
       hideBackgroundLoader();
-      Notifications.add("Something went wrong: " + e.message, -1);
+      Notifications.add("Something went wrong: " + e.message, -1, 3);
```
Miodec, line 221: Errors are supposed to have no duration, so that they stay on screen.

Miodec (changes requested):

src/sass/style.scss
avi123nash: Reverted it back. I tested in local the error message was going out of screen but when I checked the actual website the error message was correct. My Bad

src/js/leaderboards.js
avi123nash: Reverting this change too. Thanks

PR #878, opened 2021-01-23
added handdown alt layout

ramnathk:
https://sites.google.com/alanreiser.com/handsdown#h.7vp2l0jfbr5z

Adding a language or a theme? Make sure to edit the `_list.json` file as well, otherwise, it will not work!

### DIFF
--- a/src/js/layouts.js
+++ b/src/js/layouts.js
@@ -235,7 +235,7 @@ const layouts = {
             " "
         ]
     },
-    handsdownalt: {
+    handsdown_alt: {
         keymapShowTopRow: false,
         keys: [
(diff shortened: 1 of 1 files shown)


src/js/layouts.js
```diff
@@ -235,5 +235,15 @@ const layouts = {
             " "
         ]
     },
+    handsdownalt: {
```
Miodec, line 238: IMO update the name to `handsdown_alt`

Miodec (changes requested):

src/js/layouts.js
ramnathk: done

PR #899, opened 2021-01-29
Updates Spanish 1K using RAE data

yaazkal:
According to the "Real Academia Española" this an updated list of first 1000 frequencies in Spanish. Check http://corpus.rae.es/lfrecuencias.html for more.


yaazkal:
RAE list the names without the first letter in Capital, I think this facilitates the practice... but if preferred I can change the PR to keep names with the first letter in capital.

PR #905, opened 2021-01-30
Fixed Exploit Allowing JS triggered KeyboardEvents

EmNudge:
We can instantly reject non trusted events, such as those coming from client-side JS scripts. 
This doesn't fix all cases, but it stops a very simple and major exploit which I was previously able to take advantage of.

### DIFF
--- a/src/js/script.js
+++ b/src/js/script.js
@@ -4818,7 +4818,9 @@ $(document).on("click", "#commandLineMobileButton", () => {
 
 let dontInsertSpace = false;
 
-$(document).keyup(() => {
+$(document).keyup((event) => {
+  if (!event.isTrusted) return;
+  
   if (resultVisible) return;
   let now = performance.now();
   let diff = Math.abs(keypressStats.duration.current - now);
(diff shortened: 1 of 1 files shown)


SeerLite:
Can't type at all with this on Linux, both Firefox and Chrome

SeerLite:
jquery's `keydown()` event doesn't have `isTrusted` at all.
`event.originalEvent.isTrusted` works correctly on my end.

EmNudge:
> jquery's `keydown()` event doesn't have `isTrusted` at all.
> `event.originalEvent.isTrusted` works correctly on my end.

Wow, I legitimately did not know this. I assumed jQuery's event object was the same as the DOM's. Is there any reason we're using jQuery anyway? Wouldn't `document.addEventListener('keydown', e => {})` work all the same?

 I'm unable to test since firebase is absolutely broken on my machine, sorry about that. If someone can list steps to produce a build that doesn't require firebase access, I can test.

SeerLite:
> I assumed jQuery's event object was the same as the DOM's.

Don't worry, I did too! Which left me really confused when the code seemed completely correct and yet it didn't work. I found this [stackoverflow answer by luck](https://stackoverflow.com/a/52330689) :p

> Is there any reason we're using jQuery anyway? Wouldn't document.addEventListener('keydown', e => {}) work all the same?

I've also had this question for a while but haven't bothered to ask @Miodec. I've heard vanilla javascript runs a lot faster than jquery so trying to minimize the amount of jquery used might be beneficial, especially for something like Monkeytype. I'm no expert though, maybe it doesn't really matter in the end

PR #918, opened 2021-02-03
Add release and unstable deployment targets, to use with unstable/experimental channel

SeerLite:
So I did the research and this is how an unstable channel for non-release updates can be set up. I'd really like this to be implemented because I feel this way the actual releases would have more value and be more stable (more testing from community users, less bugs encountered by casual users). The changelog also becomes more consistent with previous versions

## Setup
1. Add a new site: Firebase console > Hosting > Advanced > Add another site
2. Set up a domain/subdomain for the site: `testing.monkeytype.com` and `unstable.monkeytype.com` are ones that I personally like. I don't think `beta` would be correct here, because it's use is more common for pre-release/ready-to-use versions. (Though you could add as many sites/targets as you like, and beta could be one of them!)
3. Set up `.firebaserc` according to `.firebaserc_example` or use `firebase use` (which I think does the same thing). I say this just in case you don't have it set up, since `package.json` has the `-P monkey-type` flag, which I removed in this PR to make it less "hardcoded"
4. Set up targets locally:
```shell
firebase target:apply hosting release <release-site-id>
firebase target:apply hosting unstable <unstable-site-id>
```
Notes:
* The deployment scripts for each channel is called `deploy:live:<channel>`. `deploy:live:functions` is still functions.
* The commands from step 4 edit `.firebaserc` and save the target to site-name linking there.

And that's about it! I'm pretty sure I didn't miss anything, All the steps are explained in more detail in the Firebase [docs](https://firebase.google.com/docs/hosting/multisites) if you want to check them out

Edit: Another good subdomain: `dev.`

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -1,20 +1,40 @@
 {
-  "hosting": {
-    "public": "dist",
-    "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
-    "rewrites": [
-      {
-        "source": "/privacy-policy",
-        "destination": "/privacy-policy.html"
-      },
-      {
-        "source": "**",
-        "destination": "/index.html"
-      }
-    ],
-    "cleanUrls": true,
-    "trailingSlash": false
-  }
+  "hosting": [
+    {
+      "target": "release",
+      "public": "dist",
+      "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
+      "rewrites": [
+        {
+          "source": "/privacy-policy",
+          "destination": "/privacy-policy.html"
+        },
+        {
+          "source": "**",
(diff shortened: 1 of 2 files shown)


SeerLite:
I played around with it a bit more and I see that I was wrong about how Firebase works. You're right, each site doesn't have its own set of functions but share the same one, and that may cause problems if you update the functions for the unstable target if the release one expects a different one.

I updated the script names in case you still wanna consider using this in the future. However I can understand if you can't or don't want to, or if it's too difficult to manage backwards-compatible cloud functions.

Miodec:
Ill keep it open, i might run into something that would help with the functions problem.

PR #947, opened 2021-02-10
Add 'french_2k' and 'french_10k'

iwa:
thought it was missing so... here's my contribution!

(new words aren't organized in alphabetic order)


Miodec (changes requested):
You need to add those files to the `_list` file.

iwa:
forgot to push the commit 😅
it should be fine now

PR #958, opened 2021-02-15
add international english 1k and 10k language variations

setek:
Added international english 1k and 10k languages, updated _list.json with the two new files.


Miodec:
Whats the difference between normal and international?

setek:
> Whats the difference between normal and international?

We spell things differently, for instance:

OUR: neighbour - neighbor, colour - color
SE: analyse - analyze, atomise - atomize
LL: travelling - traveling, jewellery - jewelry

Miodec:
I think a better way of going about this would be to compile a list of words that are spelled differently and add a switch to the settings that replace that word if it comes up in the generation.

setek:
> I think a better way of going about this would be to compile a list of words that are spelled differently and add a switch to the settings that replace that word if it comes up in the generation.

That sounds good! I can uh, provide the changelist:

[intl-us-eng-changelist.txt](https://github.com/Miodec/monkeytype/files/5992414/intl-us-eng-changelist.txt)

I'm not sure I can make the changes myself though …

Ahmad-A0:
I think the international version of the english_450k file should also be included.

[english_450k_international.txt](https://github.com/Miodec/monkeytype/files/6016217/english_450k_international.txt)
[changed.txt](https://github.com/Miodec/monkeytype/files/6016230/changed.txt)

**The files are in JSON format but JSON file upload is not supported by GitHub**

Miodec:
Added lazy mode which does the same thing without adding a new list

PR #973, opened 2021-02-16
Add button to export all results as JSON

g-harel:
The profile page only shows the most recent 1000 results for performance reasons. Exporting all results like this would allow people to run their own visualizations/stats without hitting the DB more than once.

I simulated a file with 50k results to test the downloading functionality and it worked after a few seconds of hanging.

### DIFF
--- a/src/js/db.js
+++ b/src/js/db.js
@@ -153,6 +153,26 @@ export async function db_getUserResults() {
   }
 }
 
+export async function db_exportRawUserResults() {
+  let user = firebase.auth().currentUser;
+  if (user == null) return false;
+
+  try {
+    return await db
+      .collection(`users/${user.uid}/results/`)
+      .get()
+      .then((data) => {
+        return data.docs.map((doc) => doc.data());
+      })
+      .catch((e) => {
+        throw e;
+      });
+  } catch (e) {
+    console.error(e);
+    return false;
+  }
+}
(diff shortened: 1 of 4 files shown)


Miodec:
Im afraid i can't merge this. Firebase bills per document reads and freely allowing all users to read all their results will increase the costs a lot.

g-harel:
Any way to restrict the usage then? I'd really like to be able to export all my stats and I'm sure there are others that feel the same.

- Could this be gated by some kind of patreon check or discord bot if this is a dollar concern?
- Could this be rate-limited to once a week?

Miodec:
Yes, i was planning to add patreon integration to allow people to view more results that way, and potentially request all data once every x days/weeks.

g-harel:
I'll gladly help out, do you have a more concrete plan yet?

Miodec:
The general plan was:
- Add a button to sync up with patreon using oauth2. Make sure that this account is not already paired with a different monkeytype account.
- Check if the user is a patreon and at which tier.
- Instead of requesting results with the client, create a cloud function that serves results to the user.
- If the user is a patron, give the ability request more results (1k per request) up to a limit, that is increasing with thieir pledge (im not sure about the numbers yet).
- Make sure that users cant just pay once and keep the perks forever.

PR #984, opened 2021-02-16
Add 1K most used Dutch words

barthakstege:
Added the 1000 most used Dutch words.


Miodec:
Please make a new `dutch_1k` file instead of expanding the default file. I would like all the default word lists to be around 200 words for consistency.

PR #990, opened 2021-02-17
Add Dutch 1K

barthakstege:
Added a seperate dutch_1k.json file containing the thousand most commonly used Dutch words. Added dutch_1k to _list.json


barthakstege:
Also added Dutch 10K now

PR #998, opened 2021-02-18
Add option to reset personal bests (#980)

bsamseth:
Ref. discussion https://github.com/Miodec/monkeytype/discussions/980

The new settings button:
image

After clicking it:
image

Stats page after reset:
image


static/quotes/english.json
```diff
@@ -1,22 +1,10 @@
...
-      600
-    ],
-    [
-      601,
-      9999
-    ]
+    [0, 100],
+    [101, 300],
+    [301, 600],
+    [601, 9999]
```
bsamseth, line 7: This, and the other language file, was changed automatically while merging updates, so I assume it is fine to leave them in this PR even though they are unrelated formatting changes.

functions/index.js
```diff
@@ -455,6 +455,22 @@ exports.removeSmallTestsAndQPB = functions.https.onCall(
   }
 );
 
+exports.resetPersonalBests = functions.https.onCall(
+  async (request, response) => {
+    console.log("heyheyhey");
```
0xfr34ky, line 460: hey

Miodec: hey hey hey!

static/quotes/english.json
Miodec: Yeah this is fine

functions/index.js
bsamseth: I see I left that one in there on accident. I would have removed it, but I see it is merged already. Oh well, very happy to have it merged so quickly! Hey hey hey!

bsamseth:
@Miodec I see this has already been pushed to the live website, but I don't think the deployment did everything it should. Trying to click the new reset button doesn't do anything and leaves this error in the console:

image

Looking through the sources I see other `new SimplePopups` definitions present, but not the one added here. Perhaps only parts of the code are deployed currently? I.e. the static index page is updated but the built js code has not?

I guess this sorts itself out as soon as you do another build update, but thought I'd let you know! 👍🏻

PR #999, opened 2021-02-18
Make chart font follow user settings

leduyquang753:
Currently the charts always use the font _Roboto Mono_. This is a simple change that applies the selected UI font instead. This makes the appearance consistent.

image


leduyquang753:
Update: This appears to not be behaving properly, looking for a fix..

leduyquang753:
The issue has been fixed by taking another approach.

PR #1003, opened 2021-02-19
Enable custom font entry when changing font through the command line

leduyquang753:
This allows the user to type a custom font name into the command input box then press Enter to set a custom font in the _Change font family_ command.

This also fixes an issue where the settings page does not reflect the current settings when directly accessing `/settings`.

### DIFF
--- a/src/js/commandline.js
+++ b/src/js/commandline.js
@@ -1658,6 +1658,18 @@ Misc.getFontsList().then((fonts) => {
       },
     });
   });
+  commandsFonts.list.push({
+    id: "setFontFamilyCustom",
+    display: "custom...",
+    input: true,
+    hover: () => {
+      previewFontFamily(config.fontFamily);
+    },
+    exec: (name) => {
+      setFontFamily(name.replace(/\s/g, "_"));
+      settingsGroups.fontFamily.updateButton();
+    },
+  });
 });
 
 let commandsFunbox = {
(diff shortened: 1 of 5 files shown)


Miodec:
This doesn't work in single line command list command line. Plus why go through all the issues of adding a `defaultBehavior()` and changing half the command line script when you could just use `input: true` like the pace caret fields?

```js
{
  id: "setPaceCaretCustom",
  display: "custom...",
  input: true,
  exec: (input) => {
    setPaceCaretCustomSpeed(input);
    setPaceCaret("custom");
  },
},
```

Am I missing something?

leduyquang753:
Well to be honest I didn't notice that existing custom implementation, but anyway I feel it's better to directly type in the custom name instead of having to pick _custom_ beforehand. If you still feel like this is unnecessary then I will switch.

Miodec:
I think I would prefer the command line to be consistent in its functionality, so ideally you should use the already present input functionality.

Plus as I mentioned your implementation does not support single list command line.

Miodec:
Could you explain to me the config load promise change?

leduyquang753:
There are two points where I made promises:

1. `configLoadPromise` is waited for by `fillSettingsPage`. Because immediately using JQuery to set active option buttons after creating those option buttons seems to not work (properly), I add the active status directly into the buttons when filling them into the settings page, which requires the current config to be loaded.
2. `settingsFillPromise` is waited for by `updateSettingsPage` so that by the time the function is executed, the settings page is sure to have the necessary buttons to be worked on and set active according to the current config.

PR #1094, opened 2021-03-13
Quote Search

Smithster:
Added a quote search option in quotes.
allows users to search by quote id, sources and text.


Miodec:
Noice

PR #1128, opened 2021-03-20
Added a few Capitalizations

MysteriousLog6:
Literally says it all :)

I just wanted to add this as a small touch.
I mainly was irritated by the test in "Restart test" not being capitalized (I don't know if it isn't supposed to be capitalized but it does look more neat).
And added one smily (Got carried away)


MysteriousLog6:
Going to capitalize more stuff (I just want it to look super even as in "Restart Test" instead of "Restart test" the latter might be more grammatically correct.
But this is just personal preference

Miodec:
I kinda feel like the smiley is out of place

MysteriousLog6:
Ok removing it!

PR #1145, opened 2021-03-28
custom test duration input parser

0x8b:
I created a simple parser for determining test duration. A number without a specified suffix is treated as a number of seconds.

test-duration


Miodec:
Nice

Issue #937, opened 2021-02-08
Clear funbox classes when taking a screenshot


uShldGetCeleste:
**Screenshot doesn't work with nausea enabled**
When nausea is enabled the screenshot doesn't work properly

**To Reproduce**
1. Enable nausea in funbox
2. Do a typing test
3. Take a screenshot
4. view broken screenshot

**Expected behavior**
Should show a screenshot that shows your stats for that typing test

**Screenshots**
![image](https://user-images.githubusercontent.com/68669265/107178889-61b00780-6992-11eb-9e12-f0ac7a884987.png)
![image](https://user-images.githubusercontent.com/68669265/107178917-712f5080-6992-11eb-8900-a6ecaccb366c.png)

**Desktop:**
 - OS: Windows 10
 - Browser Chrome Beta
 - Version: Browser - 89.0.4389.40, OS - 1909

MysteriousLog6:
This might be because the website just takes a snap of whatever is on the browser tab.

Miodec:
Funbox styles are now disabled on the result screen.

PR #1184, opened 2021-04-05
Made it so funbox gets temporarily disabled when not taking a test

Saint-dev:
Changed how funbox was called to have whatever was selected be saved to a variable. For example, this makes it so when the "nausea" or "space balls" themes (or any future visual theme) are selected in settings they are "saved" but not "active". The settings page does not change and is still usable. This also makes it so that the funbox theme is temporarily disabled going into leaderboards, account, or about pages. Will temporarily disable when results screen is shown as well to address #937

### DIFF
--- a/src/js/commandline-lists.js
+++ b/src/js/commandline-lists.js
@@ -126,9 +126,8 @@ let commandsFunbox = {
       id: "changeFunboxNone",
       display: "none",
       exec: () => {
-        if (Funbox.activate("none", null)) {
-          TestLogic.restart();
-        }
+        Funbox.setFunbox("none", null);
+        TestLogic.restart();
       },
     },
   ],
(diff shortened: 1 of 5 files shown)


Miodec:
The first issue i see with this is that there is no way to tell on the result screen if a user used for example nausea. This creates a small problem for submitting challenges on the Discord server where the moderators have to verify you completing a test with a certain funbox active.

Miodec:
image

`Funbox.activate`, 
```js
if (mode === null || mode === undefined) {
    let list = await Misc.getFunboxList();
    mode = list.filter((f) => f.name === funbox)[0].type;
  }
```
is where it breaks

Saint-dev:
Sorry I didn't see that error, I am still quite new to javascript and webdev. I think that was happening because the funbox JSON list does not have a "none" entry and the funbox activate() function was being called when the UI is loaded. I tried to clean up things a bit in the latest commit as calling funbox.activate() with no arguments is not so clear to others on what it is actually doing.

PR #1194, opened 2021-04-06
Make default timer color "main" and full opacity

SeerLite:
So it looks better by default with most theme. With some themes the default
black+0.25 was really hard to see, or just looked really bad.

An alternative way of doing this would be to make the timer color part of the theme, so each theme chooses what color it should use (and then some can keep the default black+0.25 if it looks better)

## serika dark (default)
image
image

## terra
image
image

## matrix
image
image

## 80s after dark
image
image

## arch
image
image

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -77,11 +77,11 @@ let defaultConfig = {
   savedLayout: "default",
   confidenceMode: "off",
   indicateTypos: false,
-  timerStyle: "text",
+  timerStyle: "mini",
   colorfulMode: false,
   randomTheme: "off",
-  timerColor: "black",
-  timerOpacity: "0.25",
+  timerColor: "main",
+  timerOpacity: "1",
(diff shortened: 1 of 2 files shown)


Miodec:
I actually think opacity 1 would look nicer with timer color main.

SeerLite:
I agree that with 1.0 on most themes it looks fine, but I feel like it looks too excessive/saturated with the default serika dark theme:
image
(maybe it's just me though haha. go ahead and merge if it looks fine to you)

What do you think about 0.75?
image

Miodec:
Probably feels like this because of the big numbers. What if we make mini style the default, and then opacity 1?

SeerLite:
Yeah I how that looks! Updated it
image

SeerLite:
Whoops, forgot to clear and check that :s sorry
Thanks for merging

Issue #1183, opened 2021-04-05
Add keymap legend style


Miodec:
Allow lowercase uppercase and blank. Default lowercase.

seaerchin:
i'm interested in this issue - could i get some context on what you mean by keymap legend style?

Miodec:
Basically add an option to toggle between theese:
![image](https://user-images.githubusercontent.com/13181393/113893462-f67e8800-97be-11eb-8547-da3d54e95812.png)
![image](https://user-images.githubusercontent.com/13181393/113893497-fed6c300-97be-11eb-8dce-5618f838dc3c.png)
![image](https://user-images.githubusercontent.com/13181393/113893571-10b86600-97bf-11eb-87f5-2a03910afe71.png)

seaerchin:
is the css already done? i'm updating the related js files in settings but unsure if i need to do so for the sass file

Miodec:
You will need to add a config property, add buttons to the settings page, settings to the command line and edit the css.

seaerchin:
ran into a bit of a stumbling block - the html elements in the settings page aren't being updated by `updateButton`; i've created a `settingsGroup` to encapsulate the keyboard legend style settings and i'm not sure why this isn't being updated

seaerchin:
@Miodec fyi, this is fixed in #1200

t-e-r-m:
Yes!  Thanks a ton for implementing this!

PR #1200, opened 2021-04-07
Feat/keymap legend style

seaerchin:
## Problem 
This PR addresses #1183, which is that  users want to be able to have different keymap legends. however, there is no such setting available now. 

## Solution
This PR addresses this through 
1. adding settings 
2. adding config
3. updating the DOM upon selection 

## Screenshots 
**Settings menu view**
image

**Blank**
image

**Uppercase**
image

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -17,6 +17,7 @@ import * as PaceCaret from "./pace-caret";
 import * as UI from "./ui";
 import * as CommandlineLists from "./commandline-lists";
 import * as BackgroundFilter from "./custom-background-filter";
+import keymapLegendStyles from "../../static/keymap/legend/_list.json";
 
 export let cookieConfig = null;
 export let dbConfigLoaded = false;
@@ -87,6 +88,7 @@ let defaultConfig = {
   keymapMode: "off",
   keymapStyle: "staggered",
   keymapLayout: "qwerty",
+  keymapLegendStyle: "lowercase",
   fontFamily: "Roboto_Mono",
   smoothLineScroll: false,
   alwaysShowDecimalPlaces: false,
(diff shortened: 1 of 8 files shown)


Miodec:
I think the buttons in the settings page should be: `lowercase`, `uppercase`, `blank`.

seaerchin:
> I think the buttons in the settings page should be: `lowercase`, `uppercase`, `blank`.

this is as opposed to `lowercase`, `UPPERCASE` and ` ` correct? i thought it'd be a good thematic fit to have it that way but i'll change if it's an issue

Miodec:
Yeah thats what i mean. All other settings page buttons are lowercase too.

static/keymap/legend/_list.json
```diff
@@ -0,0 +1,5 @@
+{
+    "lowercase": "lowercase",
+    "uppercase": "uppercase",
+    "blank":  "blank"
+}
```
Miodec, line 5: Also i don't fully understand the separate file approach here. Seems a bit overcomplicated compared to other configs.

src/js/config.js
```diff
@@ -1264,22 +1266,41 @@ export function setKeymapMode(mode, nosave) {
...
+  }
+
+  // Update and save to cookie for persistence
+  $(".keymapLegendStyle").addClass(style);
+  config.keymapLegendStyle = style;
+  if (!nosave) saveToCookie();
+}
+
 export function setKeymapStyle(style, nosave) {
+  console.log("keymap style", style);
```
Miodec, line 1297: Please remove console log.

static/keymap/legend/_list.json
seaerchin: i wanted a single source of truth throughout the application for the keymap legends settings but if you feel that it's unneeded/too complicated, i'll remove it!

Miodec: We don't do this for other settings so I think I would prefer it to be continuous across all settings.

seaerchin: sure, i'll change this

PR #1208, opened 2021-04-08
Added more quotes, did some refactoring, and updated README

UnrealApex:
Added quotes from the books Ready Player One, Ready Player Two, Holes, Fahrenheit 451, and The Outsiders.
@Miodec I would like to bring something to your attention. Most of quotes in the json file are formatted as 
```json
{
"text": "",
"source": "",
"id": ,
"length": 
},
```
(lines 1-30154)
but lines 30155-32525 are formatted like this
```json
{
"text": "",
"source": "",
"length": "",
"id": 
},
```
. Which format do you think the quotes should have? If you wish, I could refactor it, but it would take a few weeks because there are about 2,000 lines I will need to modify.


Miodec:
The order of properties completely doesn't matter

UnrealApex:
I would also like to get your opinion on having swear words in quotes. Some of the quotes that were added have swear words in them. Just to be safe, I left only the first letter and asterisks for the rest of the letters.

Miodec:
Things like hell, shit and other should be fine, but lets not add any of the more hardcore ones.

PR #1229, opened 2021-04-10
Added lithuanian and lithuanian 1k

Brachyurus:
I have no idea what I am doing but I'm trying to add lithuanian language


UnrealApex (changes requested):
This change cannot be accepted because you have not included a json file with words from the language. To get this change merged, create a file called `lithuanian.json` and `lithuanian_1k.json` in `your_fork/static/languages/`and add the most commonly used words from the language to the files(200 most common for `lithuanian.json` and 1000 most common for `lithuanian_1k.json`).

PR #1241, opened 2021-04-11
Fixed lithuanian word list.

Brachyurus:
Removed capitalized words.
Expanded list to 220 words.


UnrealApex (commented):
Looks good, but I am unsure if the adding an additional 20 words would be fine because most of the language files only have the top 200 words. You could add them to the `lithuanian_1k.json` file if it is not filled already. Check with @Miodec.

Brachyurus:
Alright, I'll make it 200.

UnrealApex (approved):

PR #1249, opened 2021-04-13
Lithuanian quote list.

Brachyurus:
Is there a file or a folder which I also have to edit to add new quotes file?


Brachyurus:
Added a quote and fixed curly brackets.

PR #1255, opened 2021-04-15
Increased font size of forgot password option

UnrealApex:
Increased the font size of the forgot password option so it is more visible.


static/index.html
```diff
@@ -3442,7 +3442,9 @@ <h1>delete account</h1>
             </div>
             <div class="login side">
               <div class="title">login</div>
+              <h5>
```
UnrealApex, line 3445: I am sure if this is the correct way to increase the font size in a div tag. Please share feedback  and/or documentation to clarify

Miodec:
Just increase the font size in css like this: 
https://github.com/Miodec/monkeytype/commit/66c02a80373de333a2fea81c116dd7e3ae6e4892

I've done this manually since its such a small change.

PR #1256, opened 2021-04-15
Completely updated version German_1k

Meimersdorfer:
I have completely updated the German  1k List from another source, which eliminates all the doubles and all words are in proper order now.


static/languages/german_1k.json
```diff
@@ -1,1006 +1,1007 @@
 {
   "name": "german_1k",
-  "_comment": "Sourced from: https://1000mostcommonwords.com/1000-most-common-german-words/.",
+  "_comment": "http://wortschatz.informatik.uni-leipzig.de/index_js.html",
```
UnrealApex, line 3: Leaving the source from where you have retrieved the words is fine, but I would suggest not including it.

UnrealApex (approved):

PR #1261, opened 2021-04-16
fix top filter does not show active settings between current and all

seba-nuo:
I noticed that in /account does not show what filter is active between "all" and "current settings"

Added a new group filter in account.js and index.html to track if the top filter "all" or "current settings" is active

Result


Miodec:
After pressing all, you can go into advanced, turn off some filters and the `all` button is still active. Same with current settings.

seba-nuo:
Ok, i will look at that.

PR #1275, opened 2021-04-19
Fixed missing trailing comma.

Brachyurus:


Miodec:
FYI: Trailing commas are the ones at the end of a list, which breaks JSON format. This is just a normal comma

PR #1278, opened 2021-04-19
Add simplified chinese

singularity23:
Add simplified chinese. 
_list.json updated.


UnrealApex (approved):

UnrealApex (changes requested):
All of the language json files have the top 200 words, I would suggest adding the extra ~300 to the `simplified_chinese_1k.json` file.

Miodec:
Same as #811. This will not work without a input system rewrite.

PR #1303, opened 2021-04-25
Add wordlist english24k

t-e-r-m:
source:
https://en.wiktionary.org/wiki/Wiktionary:Frequency_lists#Project_Gutenberg
"Frequency lists as of 2006-04-16"

Words from 10k-40k from the link above.  Words with punctuation and numbers, and words with length less than 4 have been removed.

24357 words.


UnrealApex (approved):

Miodec:
IMO rename to 25k.

t-e-r-m:
I've renamed it (It does have more than 24k words).

PR #1312, opened 2021-04-27
Ignoring Zen Trailing AFK #966

Estebene:
This commit addresses the improvement suggested in #966. 

This code works by storing the last time an alpha key or the spacebar was pressed. I decided not to count pressing backspaces or caps locks as 'typing' but this can be changed by adding `TestStats.updateLastKeypress()` to the handling functions in `input-controller.js`. It uses this to create a new, shorter test time which is then used to cut the critical arrays which are used to calculate stats.

### DIFF
--- a/src/js/input-controller.js
+++ b/src/js/input-controller.js
@@ -232,7 +232,6 @@ function handleSpace(event, isEnter) {
     TestUI.updateActiveElement();
     Funbox.toggleScript(TestLogic.words.getCurrent());
     Caret.updatePosition();
-    TestStats.updateLastKeypress();
     TestStats.incrementKeypressCount();
     TestStats.pushKeypressWord(TestLogic.words.currentIndex);
     // currentKeypress.count++;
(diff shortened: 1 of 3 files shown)


Miodec:
Did you test this with the local server running and did it save to the database? This could mess up the test duration verification (potentially) so just want to make sure that it's all good.

Estebene:
It did save the correct value to the database. I have noticed a slight issue where if you start and stop the test without typing you can get wpm of ~2000, however the test disregards these anyway as AFK detected.

Miodec:
How about making it so if the time from last keystroke is more than 5 seconds its not removed?

Miodec:
I just realised - this will not work for bailed out tests

Estebene:
Do you think it should work with bailouts in general or just bailouts in Zen mode?

Miodec:
image

Something is wrong. The wpm is using the spliced data, graph unspliced, time spliced, and afk time unspliced.

Estebene:
Bug has been fixed and commit was tested.

Miodec:
image

Looks like the chart data is spliced too far. There should be at least one point on the graph.

PR #1321, opened 2021-04-29
Fixed theme preview

Saint-dev:
Possible fix to get theme preview to work? Looks like hover function for items in theme list is not being called for some reason. Moving the call up a few lines of code seems to fix the issue. Maybe clearPreview function is doing something? I am still quite new to javascript so maybe I am missing something idk.

Addresses #1307

### DIFF
--- a/src/js/commandline.js
+++ b/src/js/commandline.js
@@ -43,11 +43,11 @@ function showFound() {
     try {
       $.each(list.list, (index, obj) => {
         if (obj.found) {
+          obj.hover();
           if (!/theme/gi.test(obj.id) || obj.id === "toggleCustomTheme")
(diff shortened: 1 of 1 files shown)


Saint-dev:
Simply changing Config.previewFontFamily(Config.fontFamily); to UpdateConfig.previewFontFamily(Config.fontFamily); in all instances seems to fix the exception. This allows obj.hover(); to be called so we dont have to move obj.hover(); at all. I will mark this ready for review to fix this bug, but in the future may re-write how this is being handled :)

Miodec:
Yeap, looks like a mess up on my part when i was rewriting the codebase to use modules.

PR #1322, opened 2021-04-29
Added 3 layouts: ISRT, Colemak-Qix, and ColemaQ

humptyhump:
Added three new layouts to `layouts.js`: ISRT, Colemak-Qix, and ColemaQ


NotGate:
This added the angle-modded versions of the layouts, which could confuse people wanting to try out the layout, since they probably aren't expecting that: 
image

PR #1333, opened 2021-05-01
Added Discord tag copy error handling logic and added links in table of contents

UnrealApex:
Added logic to handle the event of which username copy is not possible


static/privacy-policy.html
```diff
@@ -57,9 +57,18 @@
     </style>
     <script defer>
       function copyUserName() {
-        navigator.clipboard.writeText("Miodec#1512");
-        // alert("Copied To Clipboard!");
+        if (true) {
+          navigator.clipboard.writeText("Miodec#1512");
+          // alert("Copied To Clipboard!");
```
UnrealApex, line 62: I would like your opinion on making the username copy a bit more verbose because you don't exactly get any verification that the username is copied when you click it until you paste it.

Miodec: You would need to create a separate script file instead of keeping it in the HTML and then import the Notifications module I think

UnrealApex: I'll try that.

UnrealApex: I need a little help.

UnrealApex: For now, I'll just leave it as a TODO and work on it when I do a little bit more learning and research.

UnrealApex: @Miodec 
For now do you think I should just have an alert and console trigger when a username copy is successful?

Miodec:
Yeah ive got no clue how to make it use the Notifications module - this browserify thing is confusing. Alert is fine for now.

PR #1336, opened 2021-05-01
Added third lithuanian word list with 3k words.

Brachyurus:


UnrealApex (approved):
Looks all good. I didn't spot any errors in the json formatting, but @Miodec you might want to double check for any missing commas or double quotes for the entries that I might have missed.

static/languages/lithuanian_3k.json
```diff
@@ -0,0 +1,3011 @@
...
+	"aplink",
+	"architektūra",
+	"arčiau",
+	"atgaivinti",
+	"atrama",
+	"biologinis",
+	"demokratinis",
+	"dėžutė"
+  ]
+}
```
UnrealApex, line 3011: This is a very minor change, but I would suggest leaving one new line at the line of the file.

UnrealApex (changes requested):

PR #1344, opened 2021-05-03
Added Ruby language.

Sixeight:
Add Ruby [keywords](https://docs.ruby-lang.org/en/3.0.0/doc/keywords_rdoc.html) and bultin libraries w/o exceptions.


UnrealApex (approved):
Looks all good.

PR #1360, opened 2021-05-05
Remove offensive word from English 25k

semilin:
This should be a self-explanatory change.

### DIFF
--- a/static/languages/english_25k.json
+++ b/static/languages/english_25k.json
@@ -8599,7 +8599,6 @@
     "cashed",
     "kama",
     "rumba",
-    "niggers",
     "satyr",
     "wineglass",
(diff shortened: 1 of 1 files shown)


UnrealApex (approved):

UnrealApex:
Thank you for submitting this. I looked at the blame for the file, and it appears @t-e-r-m made the commit including this. @t-e-r-m if this was a mistake, please make sure to review your changes thoroughly before attempting to make a PR. Language of this nature is not appropriate for a workspace environment and should never merged. @Miodec I think we should secure the process of merging PRs a bit more. I would suggest requiring 2 people to review the file before it is allowed to be merged to prevent future incidents like this.

PR #1372, opened 2021-05-07
Improved custom prompts.

te-lang-wakker:
This commit introduces a few changes to the selection of custom filtered datasets:

- The "ok" button has been separated into "add" and "set"; the former doesn't clear the text field, making it easier to combine datasets.

- A few functions have been switched around to fix the loading icon and to circumvent some finnicky behaviour.

- The forEach loop has been replaced with a join; this looks a bit cleaner and is generally more performant.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -203,8 +203,8 @@
           (separated by spaces)
         </div>
         <div class="tip">
-          "Set" replaces the current selection with the selected language,
-          "Add" appends the selected language to the current selection
+          "Set" replaces the current custom word list with the filter result, "Add"
+          appends the filter result to the current custom word list
         </div>
         <i
           class="fas fa-fw fa-spin fa-circle-notch hidden loadingIndicator"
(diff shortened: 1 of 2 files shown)


static/index.html
```diff
@@ -205,7 +205,8 @@
         <i
```
Miodec, line 205: I think there should be a second `.tip` element that explains the difference between `set` and `add`

te-lang-wakker: I added the following tip to the same branch:

> "Set" replaces the current selection with the selected language, "Add" appends the selected language to the current selection

Not sure if I should open another pull request, so let me know.

static/index.html
```diff
@@ -202,10 +202,15 @@
           Use the above filters to include and exclude words or characters
           (separated by spaces)
         </div>
+        <div class="tip">
+          "Set" replaces the current selection with the selected language,
+          "Add" appends the selected language to the current selection
```
Miodec, line 207: "Selection" might not be clear to new users. I would say `replaces the current custom word list with the filter result` and `appends the filter result to the current custom word list`

te-lang-wakker: Done, I rephrased the tip.

PR #1373, opened 2021-05-07
Add missing "on" word to blind mode setting

ShawnNitke:
Simply added the missing word "on" to the index.html for the "blind mode" setting:

image


Miodec:
Not a bug. Button is 'blind'

ShawnNitke:
Ah, clever.

PR #1376, opened 2021-05-07
Renamed C# from csharp to c# and added Java quote

UnrealApex:
Changes the name of the C# files and added a Java quote. In the future more could be added, but the process of doing so is very tedious(needing to add the line breaks and tabs).


static/quotes/code_java.json
```diff
@@ -0,0 +1,17 @@
+{
+  "language": "code_javascript",
```
Miodec, line 2: `javascript` > `java`

Miodec (changes requested):

static/quotes/code_java.json
UnrealApex: Forgot to remove that after I copied the Javascript file thanks for pointing that out.

PR #1377, opened 2021-05-07
Made small changes to the PR template

UnrealApex:
Adding a language or a theme? Make sure to edit the `_list.json` file and add the `language.json` or `theme.css` as well, otherwise, it will not work!


static/languages/_groups.json
```diff
@@ -156,7 +156,7 @@
     "languages": [
       "code_python",
       "code_c",
-      "code_csharp",
+      "code_c#",
```
Miodec, line 159: Be careful about mixing in changes from other PRs. If you create a new branch for each Pr you could avoid having this happen.

UnrealApex: Thanks for the tip, I'll try that more often.

PR #1380, opened 2021-05-08
adding initial list of java quotes

mtthwn:
Adding a few quotes to the `code_java.json` file. Quotes were mainly pulled from [programming-idioms](https://programming-idioms.org/about#about-block-cheatsheets), but I've also included a small `enum` example that I made.

### DIFF
--- a/static/quotes/code_java.json
+++ b/static/quotes/code_java.json
@@ -1,0 +1,47 @@
+{
+  "language": "code_java",
+  "groups": [
+    [0, 100],
+    [101, 300],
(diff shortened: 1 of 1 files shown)


UnrealApex (changes requested):
I think it would be better to include the Java boilerplate code, however it would make it harder to format in json and is very tedious. If you aren't able to do so, its fine.

mtthwn:
Do you mean the stuff like the import statements and whatnot? I don't mind adding them to a least a few of them if you want

PR #1383, opened 2021-05-08
Added basic preset setting functionality

Estebene:
I have added basic preset setting functionality and have decided to put it on the repo for feedback on the fundamental features. This is still a work in progress, there is bound to be many bugs.


Miodec:
@Estebene the presets are not added to the command line. Please have a look at that while I investigate why certain settings are not applied correctly.

PR #1394, opened 2021-05-09
Encode tests in URL.

te-lang-wakker:
Custom tests are now encoded in the URL when applied for easier sharing:
http://monkeytype.com/?custom=Lorem|ipsum|dolor|sit|ameta
The maxUrlEncodeLength prevents the creation of URLs rejected by the browser or server.

This is a feature I'm personally quite fond of with other typing sites, but I get that you might not want go this route. If so, I hope you'll consider allowing users to create their own custom challenges instead.

### DIFF
--- a/src/js/popups/custom-text-popup.js
+++ b/src/js/popups/custom-text-popup.js
@@ -7,6 +7,7 @@ import * as WordFilterPopup from "./word-filter-popup";
 
 let wrapper = "#customTextPopupWrapper";
 let popup = "#customTextPopup";
+let maxUrlEncodeLength = 10000;
 
 export function show() {
   if ($(wrapper).hasClass("hidden")) {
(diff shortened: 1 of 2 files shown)


Miodec:
This might as well be expanded to all modes. So add things like `?time=60`, `?words=10|punctuation` and so on.

te-lang-wakker:
> This might as well be expanded to all modes. So add things like `?time=60`, `?words=10|punctuation` and so on.

Sure, sounds good. If we're going that route I'd like to make a single function responsible for applying all settings on page load instead of mutilating the codebase. Is there a file in particular you'd like this function to reside in? If not, I'll just add it to new JavaScript file.

Miodec:
Feel free to add a new module. Don't forget to add it to the `gulpfile`

CrumblyLiquid:
Doesn't seem there has been any activity regarding this feature. Is there any progress? If not I'll gladly work on this.

te-lang-wakker:
I pretty much finished encoding every setting within the URL a few days
after I started this issue, but other things got in the way before I could
polish it to be production-ready.

Feel free to take whatever code I already pushed back then.

PR #1410, opened 2021-05-12
hide replay on screenshot, gave better name to screenshot button

lukew3:


src/js/test/test-ui.js
```diff
@@ -184,6 +184,7 @@ export function colorful(tc) {
 }
 
 export function screenshot() {
+  $("#resultReplay").addClass("hidden");
```
Miodec, line 187: It looks like you're hiding the result but then not bringing it back when the screenshot is complete / errors out?

lukew3:
Made the requested change. Also added a function that reverts the page to what it was before the screenshot to remove duplicate code

Miodec:
Nice, thanks

Issue #1411, opened 2021-05-12
Accent button text color doesn't flip for light/dark


rustom:
**Describe the bug** 

Button text color needs to update properly based on theme accent colors. 

**To Reproduce** 

1. Set theme to something with light accent colors (e.g. `miami nights` or `terra`)
2. View accent color buttons in settings

**Expected behavior** 
The text in the colored buttons should be dark instead of light when the button color is light. 

**Screenshots** 
![Screen Shot 2021-05-12 at 4 17 58 PM](https://user-images.githubusercontent.com/45826892/118045517-9fc81900-b33d-11eb-9a53-c8895b2e2efd.png)

**Desktop:** 

- OS: MacOS Big Sur
- Browser: Chrome
- Version: 1.9

PR #1413, opened 2021-05-13
Fixed issue with accent colors

Estebene:
#1411

### DIFF
--- a/src/sass/style.scss
+++ b/src/sass/style.scss
@@ -2736,7 +2736,7 @@ key {
 
     .button.danger {
       background: var(--error-color);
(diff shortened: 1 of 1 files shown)


Estebene:
Screenshot 2021-05-13 222435

I have noticed the fix does look pretty bad with themes like menthol.

Miodec:
Yeah it seems like this is not the way. I've tried setting the background to `colorful error color` and the text to `text color`, which works for most.

Miodec:
Did a little hack, because i can't come up with anything else than redoing all themes JUST so they can work for 3 buttons.

PR #1414, opened 2021-05-13
Added click on account graph to scroll down to the test

Estebene:
Had a go at implementing #1409. There are still a few issues with this implementation, namely, you have to click exactly on the dot which can be difficult/frustrating. I thought about if the user could select the tooltip/dialogue box instead. Also, the selected test is given a class to show the user that it was the one they selected but this class in never removed which may become confusing.

### DIFF
--- a/src/js/account.js
+++ b/src/js/account.js
@@ -271,7 +271,7 @@ function loadMoreLines(lineIndex) {
     }
 
     $(".pageAccount .history table tbody").append(`
-    <tr class="result-row" id="result-${i}">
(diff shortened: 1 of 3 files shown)


Miodec:
Im not sure how click events are handled by chart js exactly so not sure about the clicking issue.

As for the class never being removed, how about instead of adding a class to the row just animate it with jquery just to highlight it for a second, and then go back to normal?

Miodec:
Ready to merge?

PR #1429, opened 2021-05-14
Replace curse words with sfw words

UnrealApex:
Adding a language or a theme? For languages, make sure to edit the `_list.json`, `_groups.json` files, and add the `language.json` file as well. For themes, make sure to add the `theme.css` file. It will not work if you don't follow these steps!


Miodec:
I don't think this change is needed. Plus it would be super annoying to type.

UnrealApex:
No worries, thank you for sharing your thoughts on this.

Issue #1369, opened 2021-05-06
Repeat test should enable pace caret with the wpm of the previous test.


Miodec:
It should also remember the previous state of the pace caret settings so it can revert back.

ivan-pop:
Can you add option in settings to disable this feature?
I usually play with master difficulty or accuracy limit and pace caret is annoying when i just want to improve accuracy and dont care about wpm.

Miodec:
Added `repeated pace` setting. Will be live soon.

PR #1432, opened 2021-05-15
Added repeat caret pace

Estebene:
Added #1369. This will work with both failed and successful tests that are repeated. However it will use the previous value with restarted quotes with repeat quotes in typing.

P.S It is a bit of a mess.


Miodec:
So far from my testing the pace caret is not enabled when repeating a test

PR #1440, opened 2021-05-17
Added repeat mode to settings

Estebene:
This works just as before except you have to enable it in settings first. I wonder if you want to make the mode just take the wpm of the previous test regardless of whether it was repeated. The pace caret mode in the settings is now quite long but I wasn't sure what to do about that so I left it.

### DIFF
--- a/src/js/commandline-lists.js
+++ b/src/js/commandline-lists.js
@@ -615,13 +615,6 @@ let commandsPaceCaret = {
       },
     },
     {
-      id: "setPaceCaretRepeat",
-      display: "repeat",
-      exec: () => {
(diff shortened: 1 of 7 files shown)


Miodec:
So, i think if the users can select whether this setting is on or off we could make it so it always sets the pace caret to the last wpm (if the test is repeated) no matter if pace caret is already enabled. For this the repeat pace caret would need to be its own setting - not part of the pace caret group.

In short:
- new setting
- if enabled always set to repeated pace no matter if pace caret is already enabled (and then go back to the previous setting, either pace caret off or on)

PR #1457, opened 2021-05-22
Added connection checker

### LINKED ISSUE #1456: Adding a connectivity check to Monkeytype
It would be nice to have a way to check if you had connection to Monkeytype to prevent changes that don't get put in the database,

UnrealApex:
@Miodec could you provide some feedback? I don't have that much experience with multiple scripts and wasn't sure where to load it so it isn't loaded anywhere in the repo. 

Resolves #1456


Miodec:
Is this necessary? The cloud functions will just error out. Or will this work as soon as you go offline? Also please use `Notifications`, not `alert`.

UnrealApex:
> The cloud functions will just error out. 

If a user's network refuses to send data(or just sends very small packets) to their computer but the network is still connected there is no error for the cloud functions. An example of this could be with internet parental controls. A user's device would be still connected to the network, but the network refuses to serve the device. I also think my implementation provides more insight to what the problem is than
`Error`
`Could not save result.` 
`Error: internal`

>  Also please use `Notifications`, not `alert`.

Thanks for the tip 👍 , I'll try that.

UnrealApex:
I honestly don't think this is really necessary.

PR #1467, opened 2021-05-25
fixed keymap not highlighting in simon says

### LINKED ISSUE #1435: Letters not highlighted at beginning when using simon says funbox.
**Describe the bug** 

When using the simon says funbox in tribes no letter is highlighted on the keyboard. It will show up once a random letter is typed. 

**To Reproduce** 

1. Go to tribes on the monkeytype dev website
2. Go into a private room and select the simon says funbox
3. Add someone else to the room
4. Start and see that no letter is highlighted on the on screen keyboard

**Expected behavior**

A letter should be highlighted from the beginning. 

**Desktop:** 

- OS: [Windows 10]
- Browser [Chrome]
- Version [Dev] - Is this monkeytype version?

Saint-dev:
Fixes #1435 

As mentioned in the issue...

$(".active-key").removeClass("active-key"); in config.js is being called after active key is set in keymap.js so when the test starts there is no key highlighted.

After looking in the call stacks for these functions I am still unsure why this is happening but manually calling       TestUI.showWords(); in funbox.js to call the highlighting function seems to fix everything as we are essentially calling the highlighting function in keymap.js AFTER its been set via funbox.js

### DIFF
--- a/src/js/test/funbox.js
+++ b/src/js/test/funbox.js
@@ -137,6 +137,7 @@ export async function activate(funbox, mode) {
       );
       UpdateConfig.setKeymapMode("next");
       Settings.groups.keymapMode.updateButton();
+      TestUI.showWords();
       TestLogic.restart();
     }
 


src/js/test/funbox.js
```diff
@@ -137,6 +137,7 @@ export async function activate(funbox, mode) {
       );
       UpdateConfig.setKeymapMode("next");
       Settings.groups.keymapMode.updateButton();
+      TestUI.showWords();
       TestLogic.restart();
```
Miodec, line 141: I dont think this is the correct fix because `showWords` is called inside `TestLogic.restart()`

Miodec:
Fixed in 974088926fbe42de1c7c82a1a8902a103c18b91f

PR #1475, opened 2021-05-27
Mongo

lukew3:
This is the almost complete mongo version of the project. Known errors and things that still need to be done can be found in the `backend/mongo-todo.md file`. I'm sharing this so that others can try out, test, and develop with this version.

Limited summary of what I did:
Firebase cloud functions were moved to api routes inside of an express.js server. Server interacts with the local Mongo database through mongoose. Mongoose models are in the `backend/models` folder with subschemas inside the subschemas folder. Server interacts with the frontend through `/api/` routes. Everything that could be copy/pasted from the old code was, so that new bugs could be avoided. User specific api routes require the user to provide a jwt. This jwt is automatically passed via `axiosInstance`, which also refreshes the token when it expires. Where user data used to be retrieved with `firebase.auth().currentUser`, it is now retrieved with `DB.currentUser()`, which fetches user data from localstorage. 

I'm going to keep working on this and will give an update if I think that it's completely ready. In the mean time, feel free to ask me anything.


Miodec:
Damn, this is huge. Don't even know where to start. Some initial things:
- Not sure what you mean by 'create an admin panel to make use of the analytics data'? This is using Google Analytics, not something I built myself. Or do you mean something completely different?
- The user transfer could be done all at once as I've got access to password hash parameters. They could be exported and potentially implemented into the new system? When I was initially thinking about mongo I thought about maybe leaving the authentication to firebase and only move the storage and database stuff to mongo. Would that make sense?

lukew3:
> Not sure what you mean by 'create an admin panel to make use of the analytics data'? This is using Google Analytics, not something I built myself. Or do you mean something completely different?

I was under the impression that you wanted the project to get rid of all Google/firebase dependencies, and was thinking that analytics would have to move too. Could put analytics back to the way they were.

> The user transfer could be done all at once as I've got access to password hash parameters. They could be exported and potentially implemented into the new system? When I was initially thinking about mongo I thought about maybe leaving the authentication to firebase and only move the storage and database stuff to mongo. Would that make sense?

I'd suggest just switch to mongo completely if password hashes can be moved immediately. Authentication using the new backend can be just as secure as firebase authentication. I'm not sure how firebase authentication would connect to the new database, but it could be done, likely adding latency. However, firebase can be useful because signing in with Google and validating with email are built in, removing the need for extra configuration. I'm split on which direction to take here.

Miodec:
> I was under the impression that you wanted the project to get rid of all Google/firebase dependencies, and was thinking that analytics would have to move too. I was planning on moving analytics data to the mongodb and creating a custom analytics dashboard. Could put analytics back to the way they were.

Having google analytics in is fine.

I think we can keep Firebase Auth in and somehow marry them together.

Miodec:
Alright, found a way to marry Firebase auth and Mongo.

On the client, get the auth token before every request:
```js
const idToken = await firebase.auth().currentUser.getIdToken() 
```

On the server, install the Firebase Admin package, and just decode the token:
```js
const tokenDecoded = await admin.auth().verifyIdToken(req.headers.authorization.split(" ")[1])
  console.log(tokenDecoded)
```

This will give us this: https://firebase.google.com/docs/reference/admin/node/admin.auth.DecodedIdToken

Which has all the necessary data.

lukew3:
Alright, nice. I'm checking it out right now

lukew3:
Almost ready for deployment! The changes to firebase authentication have been made, and I've fixed all of the bugs that I have found. The last thing I have to do is get the discord bot to be able to interact with the new database. Feel free to test and ensure that everything works as expected.
Once you're ready, here's how deployment should go:
* Create a linux vps using a service like [Linode](https://www.linode.com/), [Digital Ocean](https://www.digitalocean.com/products/droplets/), [Vultr](https://www.vultr.com/), etc.
* Install mongo and monkeytype+requirements on server
* Deploy the express api to `api.monkeytype.com`
  * [Here's a good video tutorial](https://www.youtube.com/watch?v=oykl1Ih9pMg), and it's companion [written tutorial](https://gist.github.com/bradtraversy/cd90d1ed3c462fe3bddd11bf8953a896)
* Move data from firebase to mongodb with the `migrate.js` node script
* deploy the main site as you normally would on firebase.

Miodec:
You're actually insane, I gotta have a look at it soon!

lukew3:
Alright, thanks. Are you planning on moving the discord bot over to the new server? I have to decide whether the bot will interact with the database directly or via api. If you're paying to host the bot right now you probably should, but if you're not it might be better just to leave it where it is.

Miodec:
Currently its hosted via AWS and fits within the free tier

Miodec:
Could you add some steps to the `contributing` file on how to connect Compass to the local database?

Miodec:
Big issue with `migrate.js` - if it fails midway the data is pretty much cut in half. Im constantly getting stopped by schemas.

Miodec:
Also, `migrate` after running through 300 users did not create any of them, or at least i cant see it

image

lukew3:
I'm not sure what the issue is. Maybe ram is becoming full? Have you refreshed mongodb-compass?

Miodec:
Oh yeah i didnt refresh it

lukew3:
I'd start by making sure that schemas are correct. I'm not sure why it's crashing. My best guess is a memory issue, but I'm not sure where

Miodec:
Problem is Firestore is completely schema less so there can be documents that might have missing fields (mostly documents that are old)

Miodec:
Also is it ok if the migration runs asynchronously? Or maybe should each user wait for the previous user to be done transfering?

lukew3:
It should be fine the way that it is now, but you can try to see if it helps

Miodec:
I cant seem to come up with a way to just ignore a schema error and keep going. I really dont want to stop the migration once it runs because of what i mentioned before - there is no way to see where it stopped in a previous run.

lukew3:
Try putting a try catch around newUser.save() and log user id and error if failed. If that doesn't work, you might have to put a try catch around every time that something is pushed to the user newUser object and when the newUser object is created.

Miodec:
Would you be available to chat on Discord in the #development channel? I tried a try catch but it doesnt seem to be working.

lukew3:
Yeah

PR #1494, opened 2021-05-31
Feature/min acc grace

Otard95:
### Checklist
- [x] If my PR is a new language or theme I modified the appropriate files to incorporate the language or theme
- [x] I understand that the maintainer has the right to reject my PR and it may not get accepted.

### Note

This is actually a feature request, I apologize if making a PR for this is inappropriate, and if so I'll be glad to create a discussion first.

### What and why

So I'm not great at typing, but I'd like to get better, which is why I started using monkeytype. But I wanted to make sure I didn't pick up any bad habits, so I want my min accuracy set to 95%. The issue with that is that if I make a mistake at the start I'm immediately dead. This feature tries to mitigate that by reducing the impact of errors at the start of each test.

### How

This PR as well as adding the setting and its corresponding frontend and command line entry, modifies the way the players accuracy is calculated in `test-timer.js` by conditionally calling a new function in `test-stats.js` if the setting is on.
https://github.com/Miodec/monkeytype/blob/15bbef6260b0eae49949379fbffaf7ec2d6402e6/src/js/test/test-timer.js#L41
This new function will multiply `accuracy.incorrect` with a fraction based on the `minAccCustom` config value and the total keystroke count before completing the accuracy calculation as usual.  
The new funktion:
```js
export function calculateAccuracyWithGrace() {
  const graceCharacters = Config.minAccCustom / 2.5;
  const grace =
    Math.min(accuracy.correct + accuracy.incorrect, graceCharacters) /
    graceCharacters;
  const incorrect = accuracy.incorrect * grace;
  return (accuracy.correct / (accuracy.correct + incorrect)) * 100;
}
```
The precise calculation feels ok to me at this point, its not helping to much, just enough, but this is just my thoughts, so I'd be happy to hear your thoughts too.

### Testing

I have been using this a good bit locally to tweak the function, and everything seems to work as intended. However I've not written any tests or similar for the function. I did however do my best to keep to the same style and structure as the project.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,8 +7,7 @@ If your change is visual (mainly themes) it would be extra awesome if you could 
 <!-- pro tip: you can check checkboxes by putting an x inside the brackets   [x]  -->
 
 ### Checklist <!-- please check the items you have completed -->
-- [] I have read the [`Code Of Conduct`](/CODE_OF_CONDUCT.md)
-- [] I viewed the [`Contributing file`](/CONTRIBUTING.md)
+- [] I have read the [`CODE_OF_CONDUCT.md`](https://github.com/Miodec/monkeytype/blob/master/CODE_OF_CONDUCT.md) and the [`CONTRIBUTING.md`](https://github.com/Miodec/monkeytype/blob/master/CONTRIBUTING.md)
 - [] If my PR is a new language or theme I modified the appropriate files to incorporate the language or theme
 - [] I checked if my PR has any bugs or other issues that could reduce the stability of the project
 - [] I understand that the maintainer has the right to reject my PR and it may not get accepted.
--- a/src/js/commandline-lists.js
+++ b/src/js/commandline-lists.js
@@ -671,26 +671,6 @@ let commandsMinAcc = {
   ],
 };
 
-let commandsMinAccGrace = {
-  title: "Change min accuracy grace option...",
-  list: [
-    {
-      id: "setMinAccGraceOff",
-      display: "off",
-      exec: () => {
-        UpdateConfig.setMinAccGrace("off");
-      },
-    },
-    {
-      id: "setMinAccGraceOn",
-      display: "on",
-      exec: () => {
-        UpdateConfig.setMinAccGrace("on");
-      },
-    },
-  ],
-};
-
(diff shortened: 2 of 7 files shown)


Miodec:
Is it really necessary to scale the accuracy this way? Currently, the min accuracy is only active after the 3rd word. We could increase it to 5 or something, but at the same time, you could just lower the setting to a value more comfortable, and work your way up.

Otard95:
> Currently, the min accuracy is only active after the 3rd word.
Correct me if I'm wrong but I believe that's only for `minWpm`. I feel like I've been failed on the first word before. But if that is the case, or its not but rather is a bug, I'll concede that this feature might be unnecessary.
 
> Is it really necessary to scale the accuracy this way?
It might be a little overly complex. The thing I do like about it is that it scales to the `minAccCustom` setting.

> you could just lower the setting to a value more comfortable, and work your way up
Again you might be right. I don't know the best way to improve, other than lots of practice. But I am afraid of building a bad habit of speed at the cost of accuracy.

Miodec:
> Correct me if I'm wrong but I believe that's only for `minWpm`. I feel like I've been failed on the first word before. But if it is the case that you are not failed until the 3rd word, or if stumbled onto a bug, I'll concede that this feature might be unnecessary.

You're right, its only for wpm. Try stashing your changes and changing the if statement in `test-timer.js:90` to this:
```js
if (
        (Config.minWpm === "custom" &&
          wpmAndRaw.wpm < parseInt(Config.minWpmCustomSpeed) &&
          TestLogic.words.currentIndex > 3) ||
        (Config.minAcc === "custom" &&
          acc < parseInt(Config.minAccCustom) &&
        TestLogic.words.currentIndex > 3)
      ) {
```

This should disable `minAcc` for the first 3 words. Let me know how it feels.

Otard95:
Not failing until the 3rd word is a much simpler solution and feels fairly similar. I think that's the better solution.

Should I create a new PR for that since it becomes a fix rather than a feature? Then we can close this and keep things nice and tidy.

Miodec:
Doesn't matter really. You can open a new one or edit this one.

Otard95:
Ok that should be it!

UnrealApex (approved):

PR #1504, opened 2021-06-03
Add Monkey Power (Power Mode - Easter-Egg)

Otard95:
### Description
Adds a mode where sparks fly and the whole page shakes with power as you type!
MonkeyPower in action

This mode is inactive and hidden in the settings until the user finds and activates it form the command line.  
Any text shown to the user is on purpose not as explicitly informative as normal, I took the liberty of (trying) to make it a bit fun.

Also there are 3 level of intensity (and it can obviously be turned off).  
 lvl 1 - sparks with color of your caret  
 lvl 2 - spark of all random colors  
 lvl 9001 - Its over 9000! Colorful sparks and shaking as you type!

### Checklist 

Resolves discussion #1461

### Testing

No explicit testing other than lots of use trying all the use-cases I can think of.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -109,7 +109,7 @@ const refactoredSrc = [
   "./src/js/input-controller.js",
   "./src/js/route-controller.js",
   "./src/js/ready.js",
-  "./src/js/power-mode.js",
+  "./src/js/monkey-power.js",
 
   "./src/js/account/all-time-stats.js",
   "./src/js/account/pb-tables.js",
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -109,6 +109,7 @@ const refactoredSrc = [
   "./src/js/input-controller.js",
   "./src/js/route-controller.js",
   "./src/js/ready.js",
+  "./src/js/power-mode.js",
 
   "./src/js/account/all-time-stats.js",
   "./src/js/account/pb-tables.js",
--- a/src/js/commandline-lists.js
+++ b/src/js/commandline-lists.js
@@ -1293,6 +1293,37 @@ let commandsCopyWordsToClipboard = {
   ],
 };
 
+let commandsMonkeyPowerLevel = {
+  title: "You control the power",
+  list: [
+    {
+      id: "monkeyPowerLevelOff",
+      display: "off",
+      exec: () => UpdateConfig.setMonkeyPowerLevel("off"),
+    },
+    {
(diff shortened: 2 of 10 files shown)


Miodec:
In my opinion dont add it to the settings page. Just add one hidden position in the command line `Monkey Power Level` which has options: `off, high, ultra, >9000`

Miodec:
Also how about another option where sparks are based on the caret color and screen shake is on?

Otard95:
> Also how about another option where sparks are based on the caret color and screen shake is on?

Actually was going to do that. But didn't want another lvl in the settings menu. But if we just move all settings to command line, that's a non issue.

I'll move the settings to the command line.

Otard95:
- Moved all settings to command line
 - Added power level with regular sparks but with shake
 - Renamed all power levels 
   - off - Power is completely contained
   - mellow - Sparks of your cursors color or red when making mistakes
   - high - Colorful sparks
   - ultra - Normal sparks with shake
   - \>9000 - Full power, colors and shake

Miodec:
I had to make some changes because the cloud function that was saving the config did not like the `>` character. Also I removed the `unlocked` property, because it was redundant (off works the same way).

As for rgb theme, I'm not sure if you can get the current state of a css animation with js, but feel free to attempt it after you wake up. Ill merge this in the meantime.

UnrealApex:
I was thinking of making the Easter Egg's options simpler. Maybe just three options? That's up to you.

Miodec:
Also shake is causing scroll bars to show up.

UnrealApex:
The option to enable power mode is visible without searching for it. I think it would be a better Easter Egg if it was hidden by default like the carrot caret Easter Egg in the settings. @Miodec @Otard95  what do you think?

### Current
power mode

### Desired
https://user-images.githubusercontent.com/79121360/120721035-ebc03680-c481-11eb-93c9-07145e02d127.mp4

Miodec:
Yeah, it should be hidden. Ill do that tomorrow

Otard95:
I can agree that the `unlock` setting is redundant. It was necessary when there was UI elements. As for the `level` setting what the actual value is is irrelevant but i think still displaying texts to the user instead of number would look better.

On the  theme front its using the `ThemeColors.caret` and `ThemeColors.error` values. So if either of those in the RGB theme is a non hex or `rgb(...)` color i think it'll have issues.

PR #1520, opened 2021-06-05
Add Urdu

BalkoBalkho:
Added Urdu language

- Updated _list.json and _groups.json

I hope everything is alright.


victorbnl:
There are around 450 words, where there should be 200. I don't know if this is a problem though, but your pull request might not be accepted because of this

Miodec:
I think i would like all the base word lists be 200 words in length for consistency. You could find 500 more words and create a second, 1k version.

BalkoBalkho:
Alright, I have trimmed it to have 200 words.

Miodec (approved):

PR #1578, opened 2021-06-26
More Quotes

UnrealApex:
This is a super ambitions PR, trying to add maybe a few thousand quotes.


Miodec (commented):
Thats a lot of empty quotes

UnrealApex:
Going to enter values into the json soon, just quickly copied and pasted the key value pairs to save time.

Miodec:
Tested locally? Does it load quotes without json errors?

Miodec (approved):

UnrealApex:
> Tested locally?

I currently don't have a machine that I run Monkeytype on. I do have a gaming laptop I can use(not my own but a "family laptop"), but I'm not allowed to clone any repos on it other than my own. Other than that, most of my commits are done on a school chromebook via [Gitpod](https://www.gitpod.io/) as my school administrators have disabled the installation of linux apps thus I am unable to install tools I usually use such as VS Code, Git, etc.... Gitpod includes npm and node, but when trying to sign in after installing Firebase, I am unable to because my school admins have also disabled the sign in of non school domain Google accounts(For GitHub and other services, I just sign in with my personal email).

Issue #1570, opened 2021-06-23
Min burst setting


Miodec:
### Discussed in https://github.com/Miodec/monkeytype/discussions/1564

<div type='discussions-op-text'>

<sup>Originally posted by **NotGate** June 22, 2021</sup>
This would go next to `min wpm` and `min acc` in the settings:
If you type a single word under this speed, the test fails. 
Even if someone is typing 60 wpm, if each word is typed above the burst minimum (ex: 300 wpm), the test keeps going.
This should only consider raw wpm.

When set, this could have an option called `count slow words as missed words`, which wouldn't fail the test if you typed the word below the `min burst` but would allow you to `practice missed words` for what you're slow *and* inaccurate at. 

I think this feature would be amazing for developing muscle memory for 'word-by-word' typing and for practicing good habits for higher speeds. </div>

Miodec:
Added

PR #1579, opened 2021-06-26
Added min burst

yzAlvin:
### Description
I have attempted to add min burst.

* Added toggle live burst speed from command line and settings page
* Added fail if burst speed is lower than custom speed, can set in command line and settings page
* Burst speed is in stats page after the test, but does not have any value

The calculateBurst() method in test-stats.js just needs to be modified to return the burst speed and everything should work. A method proposed by Miodec in Discord:
Track the timestamp the word was started and finished
Then calculate the wpm based on how many letters in that word + space

Hopefully someone can use this PR to finish this feature?

### Checklist 

Addresses #1570


Miodec:
Ideally it would be best to keep the code for burst tracking in a separate file, something like `burst-tracker.js`

PR #1606, opened 2021-07-08
Add weakspot mode.

shishberg:
### Description
Keeps track of a per-character "weakness" score, basically a moving average of the spacing duration with a penalty for incorrect letters. The weakspot funbox mode uses this to bias towards words with lots of weak letters.

There are a few hardcoded parameters here that I tweaked a bit until it seemed right to me, but could use some work. LMK if there's a better place to put them.

I've only tested it with English; it should work with other scripts but might behave a little weirdly if there are a lot more characters. (It wouldn't break, it just might not make obviously sensible word choices.)

It doesn't persist weakness scores at the moment. Having used it for a bit I actually like it that way, it gives a feel of working on particular things for one session but being able to reset it.

### Checklist 

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -168,6 +168,7 @@ const refactoredSrc = [
   "./src/js/test/layout-emulator.js",
   "./src/js/test/poetry.js",
   "./src/js/test/today-tracker.js",
+  "./src/js/test/weak-spot.js",
   "./src/js/replay.js",
 ];
 
--- a/src/js/input-controller.js
+++ b/src/js/input-controller.js
@@ -23,6 +23,7 @@ import * as Focus from "./focus";
 import * as ShiftTracker from "./shift-tracker";
 import * as Replay from "./replay.js";
 import * as MonkeyPower from "./monkey-power";
+import * as WeakSpot from "./weak-spot";
 
 $("#wordsInput").keypress((event) => {
   event.preventDefault();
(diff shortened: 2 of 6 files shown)


Miodec:
I think it would be great if it had its own module for starters

Miodec:
Ill check it out in detail tomorrow

shishberg:
> I think it would be great if it was in its own module (seperate file) for starters

Yeah makes sense, I'll move it.

Miodec:
Yeah i think it looks good so far. Just need to move it to a module and its good to merge

shishberg:
PTAL

Miodec:
Nicely done. I will merge it for now to see what people think. With the feedback, we can adjust the weights. Could you just explain real quick how the weights will affect this system? Also feel free to update this if you find a different way of scoring words (some of the comments mentioned taking wpm into consideration)

Miodec (approved):

shishberg:
> Could you just explain real quick how the weights will affect this system?

Sure.

* `adjustRate` changes how quickly it 'learns' scores - very roughly the score for a char is based on about the last `1/adjustRate` times it was seen. Larger numbers adjust faster. Should be between 0 and 1.
* `wordSamples` affects how much it biases word choices towards weak letters. Larger = more weak keys.
* `defaultScore` is the score that every char starts on. The ideal value would be the average spacing in milliseconds (hence the WPM comment) but there's no easy way to know that at the start... Having a fixed default value seems to work okay even if it's way off, it converges pretty quickly.
* `incorrectPenalty` is the score penatly (milliseconds) for getting a letter wrong.

This might be worth putting in a comment...

PR #1644, opened 2021-07-19
Use Notification instead of alert on successful clipboard write

UnrealApex:
Updated copy username function to use the notifications module instead of the alert function. @Miodec any feedback as I'm not sure if I wrote this correctly?


Miodec:
As far as i know this will not work because the privacy policy is not 'browserified', meaning the notifications module is unavailable from that file

PR #1661, opened 2021-07-23
Add GitHub workflow to lint and auto-format PRs

SeerLite:
### Description
Adds a workflow that on every PR **merge** to **any** branch, runs prettier (`pretty-quick`) to fix the new commits. This should finally fix the issues with formatting coming from PRs made through the GitHub UI or other means that skip the pre-commit hook.

After merging, you should run
```
npx prettier --write .
```
on all active branches to fix all leftover formatting issues.
After this, no more badly formatted code should ever reach master again, which means no more noise diff on unrelated commits :)

### Checklist 
- [X] I have read the [`CODE_OF_CONDUCT.md`](https://github.com/Miodec/monkeytype/blob/master/CODE_OF_CONDUCT.md) and the [`CONTRIBUTING.md`](https://github.com/Miodec/monkeytype/blob/master/CONTRIBUTING.md)
- [X] I checked if my PR has any bugs or other issues that could reduce the stability of the project
- [X] I understand that the maintainer has the right to reject my contribution and it may not get accepted.

Marking as draft until input rewrite is merged so I don't have to fix new formatting conflicts >:)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,46 @@
+name: Lint and auto-format PRs
+
+permissions:
+  contents: write
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  lint_and_format:
+    name: Lint and auto-format
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout previous commit and then pull, to fetch everything in between
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.base.sha}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+      - run: |
+          git pull origin ${{github.event.pull_request.head.ref}}
+          git checkout ${{github.event.pull_request.head.ref}}
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v2.3.0
+        with:
+         node-version: 14
+         cache: npm
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run ESLint
+        run: npx gulp lint
+
+      - name: Run Prettier
+        run: npx pretty-quick --branch ${{github.event.pull_request.base.sha}}
+
+      - name: Commit and push
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          git add --update
+          git commit -m "Auto-format with Prettier" || :
+          git push origin HEAD:${{github.event.pull_request.head.ref}}


ssantichaivekin:
The way I usually see people set this up is, they will have Github Actions run prettier against the new code just to make sure that the contributor has already run prettier before creating their pull requests. The Github Actions will reject the changes (by showing that some tasks fail) but will not commit on the contributor's or maintainer's behalf. Note that if any of the steps in the CI return a non-zero value, the step is considered failed.

Here are some examples: [vscode](https://github.com/microsoft/vscode/blob/14de627cf237d4c6d9375fa8d411432eae63858a/.github/workflows/ci.yml#L251), [three.js](https://github.com/mrdoob/three.js/blob/940816f6a1a3fe7a0fc703b8242eb086eebb00f1/.github/workflows/ci.yml#L31)

Miodec:
> The way I usually see people set this up is, they will have Github Actions run prettier against the new code just to make sure that the contributor has already run prettier before creating their pull requests. The Github Actions will reject the changes (by showing that some tasks fail) but will not commit on the contributor's or maintainer's behalf. Note that if any of the steps in the CI return a non-zero value, the step is considered failed.

Yeah this sounds like a better solution.

SeerLite:
I thought about doing that too but since there seems to be a lot of people that have contributed static files like languages and quotes (and even code sometimes) from envs without the pre-commit hook like the GitHub UI, I concluded that automating the process and running `pretty-quick` on master was better.

Otherwise contributors for small changes or language additions would have to set up their env to run it locally or check the output of the workflow and try to fix it manually from the UI.
Or, @Miodec could manually run Prettier on the branch when the workflow fails (?).

So yeah what do you think?

ssantichaivekin:
@SeerLite If you really want it that way, I would suggest that you tell Github Actions to commit the linted changes to the branch of the pull request, rather than on master.

SeerLite:
You're right. I thought it'd just make it harder for the contributor since they'd have to `pull` from the branch which might be unexpected, but now I realize if they're making the PR from the GitHub UI in the first place they won't have to deal with any of that.

.github/workflows/main.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Run Prettier
+        run: npx pretty-quick --branch ${{ github.event.pull_request.base.sha }}
+        
+      - name: Commit and push
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          git add --update
+          git commit -m "Auto-format with Prettier" || :
+          git push origin HEAD:${{ github.event.pull_request.base.ref }}
```
ssantichaivekin, line 39: You might have to authenticate with a GitHub account that has a repository write permission before you can run `git push`.

ssantichaivekin:
@SeerLite There's an additional concern that I just thought about: the bot has to be authenticated (i.e, logged-in to your account) to make changes to the specified branch.

If you want the bot to make changes to the master branch:
 - The bot has has to authenticate with a user that has a master branch write permission.

If you want the bot to make changes to a pull request in another repository
- The bot has has to authenticate with a user that has a repository write permission.
- The user who opened the pull request has to check the [Allow edits by maintainer](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) box .

.github/workflows/main.yml
SeerLite: That's set up automatically with the `checkout` action. I have tested it and it works fine. There also are examples of committing and pushing in the [checkout repo](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token).

SeerLite:
> If you want the bot to make changes to the master branch:
> 
> * The bot has has to authenticate with a user that has a master branch write permission.
> [...]
> * The bot has has to authenticate with a user that has a repository write permission.

Should be set up automatically via the [checkout action](https://github.com/actions/checkout#whats-new):

> * Script authenticated git commands
>    * Auth token persisted in the local git config

As for:
> * The user who opened the pull request has to check the Allow edits by maintainer box .

That should just cause the `push` command (or maybe even the whole `checkout` action) to fail and then we'd have an error like you suggested in the first place. And this would only happen if the PR isn't already auto-formatted.

SeerLite:
Updated it. I fixed some stuff like e.g. now it's using `npm ci` instead of `npm install`, basing myself off of the linked examples.

Also added an ESLint job. The whole workflow should fail if ESLint fails, and then it's just up to the contributor making the PR to fix it.
If ESLint succeeds, the Prettier workflow should run and if auto-formatting is performed, it's commited and pushed to the PR branch.

Miodec:
I was gonna say, what if prettier fails for whatever reason. Is the pr merged? What happens?

SeerLite:
@Miodec After my changes today it doesn't run on merges anymore but on PR creations/updates and it all happens inside the PR.

Summary:
1. Someone creates or updates a PR.
2. Action runs ESLint on the code of the PR.
    * If it succeeds, it continues with step 3.
    * If it fails, it aborts the Action and marks the PR as Failed. Probably sends an e-mail to the PR author but I'm not sure.
3. Action runs Prettier on all commits on the PR.
    * If it finds no need to auto-format, nothing happens.
    * If it succeeds, a new commit is made on the PR called "Auto-format with Prettier".
    * If it fails, it marks the PR as Failed.

ssantichaivekin:
@SeerLite I think you could test the that GitHub Actions config file you wrote does what you expect it to do.
1. Create a new private repository.
2. Copy [the files in the branch](https://github.com/SeerLite/monkeytype/tree/prettier-workflow) to the new repository.
3. Push the code to GitHub.
4. Create a pull request in that repository to see whether the workflow runs the same way you expect it to.

I have looked at the code, and I think there are a few comments:
1. I'm not sure why you are running both ESLint and Prettier. They seem to be a similar program. `eslint --fix` would probably cover all functionalities offered by prettier.
2. The two jobs will run concurrently in two different machines. If either fails, GitHub will say "some checks failed".
3. In step 3.2 you mentioned that if prettier succeeds it will create a new commit and push the new commit to the branch of the pull request. However, you need permission to write and push a commit. Github Actions by default do not have such permissions. This will lead to the push failing with 401: Unauthorized.

SeerLite:
> I'm not sure why you are running both ESLint and Prettier. They seem to be a similar program.

That would be the same reason Mio is running both of them (ESLint on building and Prettier on commit), which I assume is what is said in the [Prettier FAQ](https://prettier.io/docs/en/comparison.html):

> In other words, use **Prettier for formatting** and **linters for catching bugs!**

And I'm pretty sure (no pun intended) that Prettier does a lot more than just check that the code works. It is "An opinionated code formatter", after all.

Miodec:
> I'm not sure why you are running both ESLint and Prettier. They seem to be a similar program.

To my knowledge ESLint doesn't enforce any specific formatting rules like Prettier does.

ssantichaivekin:
@Miodec I think you can add specific rules to the .eslintrc file, but I'm not very experienced in javascript, so I think I should leave this discussions to you guys.

SeerLite:
I originally based myself on the linked three.js CI so made linting and formatting separate jobs (though it seemed kinda weird). Now I realized it's a waste of resources and it's better to just put them both in the same job, like the linked vscode CI file.

Miodec (approved):

PR #1710, opened 2021-08-06
Add Github theme (light, dimmed and dark variants)

itscarlosecp:
Added Github Theme in three variants (light, dimmed and dark)

### Description

These themes are using [Primer's color system](https://primer.style/css/support/color-system), the same used in the Github Theme extension for VSCode.

#### Variants
- Light

github_light

- Dimmed

github_dimmed

- Dark

github_dark

### DIFF
--- a/static/themes/_list.json
+++ b/static/themes/_list.json
@@ -573,5 +573,20 @@
     "name": "material",
     "bgColor": "#263238",
     "textColor": "#80cbc4"
-  }
+  },
+	{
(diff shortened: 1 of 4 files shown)


Miodec:
Sorry but i don't think im going to merge this one as there are very similar themes already available. Themes that look very close:

- paper
- nord 
- midnight and arch

I want to limit themes that are similar because I feel like there is already too many themes.

PR #1722, opened 2021-08-15
Pseudolang funbox mode - gibberish, but more wordlike

shishberg:
### Description

**NOTE:** I'm not sure whether this is a good idea. It seemed fun so I thought I'd share it, but it's totally up to you whether to do anything with it.

Generates random words using ngram distributions from the current language. The result is gibberish (I originally called it "gibberish" before I realized that was already the name of a funbox mode) but with similar patterns to real language.

Screen Shot 2021-08-15 at 2 17 14 pm
Screen Shot 2021-08-15 at 2 17 14 pm
Screen Shot 2021-08-15 at 2 17 23 pm
Screen Shot 2021-08-15 at 2 17 39 pm

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -169,6 +169,7 @@ const refactoredSrc = [
   "./src/js/test/poetry.js",
   "./src/js/test/today-tracker.js",
   "./src/js/test/weak-spot.js",
+  "./src/js/test/wordset.js",
   "./src/js/replay.js",
 ];
 
--- a/src/js/test/test-logic.js
+++ b/src/js/test/test-logic.js
@@ -525,7 +525,7 @@ export async function init() {
       });
     } else {
       for (let i = 0; i < wordsBound; i++) {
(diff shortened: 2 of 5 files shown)


Miodec:
How do you actually generate this? I don't see a bigram list or anything like that

shishberg:
> How do you actually generate this? I don't see a bigram list or anything like that

The `WordGenerator` constructor generates an ngram list from the language.

Miodec:
How about we rename this funbox to `pseudo_lang`, `pseudolang` or `pseudo_language` - your choice.

shishberg:
Done.

Miodec:
Also, have you tested it past 100 words? After 100 the word generation uses the function `addWord`

shishberg:
I think so. Let me check...
...
Yep. Tested with 200 words, seems fine.

Miodec (approved):

PR #1727, opened 2021-08-16
Chaos Theory

mTopy:
### Description

Added Chaos Theory Theme
----
Uses a dark background and pink main color with many flipped words and letters, modifies the default caret with an image from the set and adds a logo to the theme picker.

Screen Shot 2021-08-16 at 10 23 14 AM


mTopy:
The set will go into IC soon btw

Miodec:
Not really sure about merging this one, Luna theme seems pretty similar.

I've already turned down some themes due to similarity - I feel like there is too many themes already.

dotnickmk:
I think Luna is actually pretty different in color, features, and its overall execution.

The Chaos Theory theme has several complementing colors that are featured on the menu icons which Luna does not have. There are other features in the Chaos Theory theme such as the reversing of incorrect letters and letters in the banners (which plays into one of the kits of the keycap set), as well as the flipping animation of the banners upon typing which is not seen in the Luna theme. Generally, the colors of the current theme are not as purple as the Luna theme is as well.

In my opinion, this theme is pretty unique to anything I've seen on Monkeytype so far that goes above simply reskinning the template with colors. I understand the hesitancy to accept new themes given the number of current themes, there are definitely a lot available currently. But I think this adds something people haven't seen before on this website, and is something new and immersive for people to try out. 

I would really appreciate your reconsideration of this.

- Nick

mTopy:
Attached a clip trying to show off some of the animations and reversed text which help make the theme unique.


https://user-images.githubusercontent.com/71908482/129640010-228d555b-1933-45bf-98c0-9b974849c45f.mov

Miodec:
Okay, i will merge this in if you get this set into the GB stage

PR #1736, opened 2021-08-18
Ctrl/Cmd+Shift+P as alternate way to open command line

### LINKED ISSUE #1731: Add a config item to change what key/key combination brings up the command line
Add a swtich `esc | ctrl/cmd + shift + p`

Some users might want to use the vscode like shortcut.

UnrealApex:
Added ability to open the command line with ctrl+shift+p. This change allows Monkeytypers to use the command line while in full screen. Previously, when attempting to use the command line in full screen, pressing `esc` would just exit full screen and the command line would not open. 

**note: code is untested as my Gitpod container autodeletes Firebase for some reason, need to change the Docker image cause I really don't want to install, login and reconfigure Firebase every time.** Also note that keyCode is now [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) so might want to remove and refactor that in the future.

Closes #1731


UnrealApex:
Do you think I use refactor it to use [`event.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) instead?

Miodec:
This change makes it so you have to hold esc to see the command line.

Ctrl shift p and cmd shift p don't work.

UnrealApex:
> Ctrl shift p and cmd shift p don't work.

Working, try testing this code
```javascript
var map = [];

// too lazy to get the actual config object so just created a test one here
var Config = {
	"swapEscAndTab": false
};

onkeydown = onkeyup = function(e) {
  e = e || event; // to deal with IE
  map[e.keyCode] = e.type == 'keydown';
  if (event.keyCode == 27 || map[17] && map[16] && map[80] || (event.keyCode == 9 && Config.swapEscAndTab)) {
		if (event.keyCode == 27) {
    	console.log("Open command line with Escape");
    } else if (map[17] && map[16] && map[80] ) {
    	console.log("Opened command line with Ctrl + Shift + P");
    } else if (event.keyCode == 9 && Config.swapEscAndTab) {
    		console.log("Opened command line with tab ");
    
    }
  }
}

```

Miodec:
Esc triggers twice - on key up and key down.

Cmd shift p doesn't work.

Ctrl shift p works. 

If you press cmd + shift + p, it doesn't work, but then if you press ctrl + shift (without p) it still triggers.

Miodec:
So, i just had a go at this... why not do something like this? 

```js
document.addEventListener('keydown', (e) => {
  if(e.key.toLowerCase() == "p" && (e.metaKey || e.ctrlKey)){
    e.preventDefault();
    if(e.shiftKey){
      console.log('command line');
    }
  }
})
```

UnrealApex:
Sure, that would work.

UnrealApex:
Anything else this PR is missing/needs improvement on?

Miodec (approved):

PR #1745, opened 2021-08-21
remove diacritics from arabic_10k.json & add more qouts

Mustafa-Asaad-iq:
1- remove diacritics from `arabic_10k.json` :
It is not used much only in sensitive texts (poetry, Bible, Qur'an , etc..) and in the case of fast writing if you are typing 100wpm
it will be nearly 60wpm with diaritics :(
NOTE: there is no diacritics in `arabic.json`
Actually in the future I will add  somthing like `arabic-diacritics.json` in case any one want to use them 

2- add more 9 qouts


Miodec (approved):

ZER0-X:
I've thought about creating a `arabic_10k-diacritics.json` but @Miodec said he was thinking about adding a button for this like the 'numbers' button for example. I don't know if he gave up on the idea or not.
https://github.com/Miodec/monkeytype/discussions/1423#discussioncomment-738549

Miodec:
Ive just added a no accent/no diacritis mode thing, check it out here https://github.com/Miodec/monkeytype/commit/c063012e9e148dc35517c6880e8fc2c69a7b4dba

ZER0-X:
I've added the Arabic diacritis to the lazy mode.
 #1856

Mustafa-Asaad-iq:
> I've added the Arabic diacritis to the lazy mode list.
>  #1856

should I return the old 10k_arabic file now?
also do you think we should add diacritis to.`arabic.json` ?

Miodec:
Probably

ZER0-X:
> should I return the old 10k_arabic file now?
> also do you think we should add diacritis to.`arabic.json` ?

You added some new words so if you want you can use [mishkal](https://pypi.org/project/mishkal/) in Python to add the diacritis to the new words, it's smart.

In some words in `arabic.json` we need to replace the character "ا"with some 'أ', 'إ' and 'آ'  then we could use [mishkal](https://pypi.org/project/mishkal/) for every word.

If you don't have time or you don't know how to use Python don't worry, as soon as I have some time I will make a script for that.

Mustafa-Asaad-iq:
> > should I return the old 10k_arabic file now?
> 
> You added some new words so if you want you can use [mishkal](https://pypi.org/project/mishkal/) in Python to add the diacritis to them, it's a smart software.
> 
yes I will use mishkal but not python library I will just use the website 

> > also do you think we should add diacritis to.`arabic.json` ?
> 
> For some words in `arabic.json` we need to replace the character "ا"with some 'أ', 'إ' and 'آ'  then we could use [mishkal](https://pypi.org/project/mishkal/) for all of them.
> 
sure it is small file I will do it 

> If you don't have time or you don't know how to use Python don't worry, as soon as I have some time I will make a script for that and you can work in `arabic.json`.

I have time but not much
I know little bit about python but I will just use the website
I will tell you if I had any problems with the website to create the python script

Mustafa-Asaad-iq:
Hi I was working on `arabic.json` and when  I enable lazy mode there's nothing happens
diacritics are still (lazy mode doesn't remove them) and with no errors so what should I do know?

Miodec:
Try again - the accents were added to the wrong file.

ZER0-X:
This is my script: https://gist.github.com/ZER0-X/980a4426390e635a3ace4f23eefb6bce

I used it in: #1878 

If you want to use it, make like this:
```
$ python monkeytype_add_arabic_diarcitis.py <input file> <output file>
```

PR #1746, opened 2021-08-21
Update pull_request_template.md

UnrealApex:
### Description

Used a better word for what a PR does when it closes an issue.


Miodec:
I meeean, i dont know if this tip makes sense because they would use the . shortcut *before* making a PR.

UnrealApex:
Well, that does make sense, but hey, easter egg.

Miodec:
How about just include both

UnrealApex:
Sure

Miodec (approved):

PR #1749, opened 2021-08-23
Add Irish quotes

Moilleadoir:
Adding Irish quotes file

### DIFF
--- a/static/quotes/irish.json
+++ b/static/quotes/irish.json
@@ -1,0 +1,743 @@
+{
+  "language": "english",
+  "groups": [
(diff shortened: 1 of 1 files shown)


static/quotes/irish.json
```diff
@@ -0,0 +1,743 @@
+{
+  "language": "english",
```
Miodec, line 2: Change to irish

Moilleadoir: Yikes! Fixed.

Miodec (approved):

PR #1752, opened 2021-08-23
Update turkish.json words.

Yultax:
Add 600+ new Turkish common words.


Yultax:
Added 600+ new words, and removed the "devam etmek" item because it consists of 2 words so i don't know that is it right to keep it in "words" category?

PR #1754, opened 2021-08-23
Added a delimiter switch

### LINKED ISSUE #1728: Custom mode delimiter switch
Current custom mode breaks up everything on space. This means that if the user wants some words to always appear together ( like, I don't know, "big house" ) the space will make it impossible, and randomise those words separately. 

Adding a switch on the custom text popup, that toggles between a space delimiter and a pipe delimiter would allow users to randomise phrases (for example `big house|test|increase|other`). With this, the phrase `big house` will never be generated on their own.

Adding support for this could also allow phrases to be used in language files, for languages that have phrases where having the words be separate would not make much grammatical sense.

pran01:
Adding a new feature, Delimiter Switch in custom text mode.
image
image

### Description

Added a switch on the custom text popup, that toggles between a space delimiter and a pipe delimiter. It would allow users to randomise phrases (for example "big house|test|increase|other"). With this, the words "big" and "house" will always occur together.

Closes #1728


Miodec:
One thing i found:

- Even if the delimiter is set to `|`, the words filter still inserts words separated by space, making them not randomised.

Miodec:
Also, after seeing this i think i would prefer the delimiter selection to be below, like other settings

image


Something like:

(checkbox) Pipe delimiter
Change how words are separated. Using the pipe delimiter allows you to randomize groups of words.

pran01:
> One thing i found:
> 
> - Even if the delimiter is set to `|`, the words filter still inserts words separated by space, making them not randomised.

That's not what's happening. Tell me if i am wrong. The words seperated by space come alone only in the last. That is to take care of words limit.
If you say you want 25 words, its gonna give you 25 words, but if the last phrase contain 3 words, it will only give those words that will make it 25.
Is this not acceptable?

pran01:
> Also, after seeing this i think i would prefer the delimiter selection to be below, like other settings
> 
> image
> 
> 
> Something like:
> 
> (checkbox) Pipe delimiter
> Change how words are separated. Using the pipe delimiter allows you to randomize groups of words.

I can do that.

Miodec:
> That's not what's happening. Tell me if i am wrong.

Set the delimiter to `|`. Then open the words filter menu, and click `set` - the words are still using space as a delimiter.

pran01:
> > That's not what's happening. Tell me if i am wrong.
> 
> Set the delimiter to `|`. Then open the words filter menu, and click `set` - the words are still using space as a delimiter.

Yes, Sorry I didn't see the wordFilter function, I didn't realise delimiter would have been used there as well. I will edit it there as well, and also make it into a check box instead of switch. Are you okay with what I said though? If the words appear in the last, it can't exceed the word limit, so phrases might be incomplete in the last word of the paragraph to respect the word limit.

Miodec:
Yeah, cutton off phrases to fit in the word limit is fine.

pran01:
I have done both of them, but as I am not fully aware of all the functionalities of the website, there might be some other places as well, where delimiter might have been used, but I am not aware. So, if there is any other place, please inform me.

Miodec:
I *thiiiink* the only other place might be challenge loading.

Check what happens when a user switches the delimiter to `|` and then loads a challenge from the command line.

historybandgeek:
I had two thoughts -
1) Making phrases visible (mentioned in the issues thread as well): highlighting or underlining the phrase would help the learner see "and he said" instead of "and" "he" "said" etc..
2) Would the TTS function say the whole phrase or still one word at a time?

These are just optional, but my thoughts as one who plans on using this feature. Thanks for your great work!

Miodec:
1. Not sure how the ui for that would work
2. Nah, i dont think so

pran01:
> 1. Not sure how the ui for that would work
> 2. Nah, i dont think so - single word is fine

I agree. I think its enough that those words come together, its not important that it looks like a phrase or just like other words.

Miodec:
> I opened the command line from esc, and applied modes. 

The command line will have a "Load Challenge" option. If not you need to make sure to pull the latest code from remote

pran01:
> > I opened the command line from esc, and applied modes.
> 
> The command line will have a "Load Challenge" option. If not you need to make sure to pull the latest code from upstream

okay I tried some challenges. They seem to work fine to me. I guess the delimiter only has a value in custom mode.

Miodec (approved):

Miodec:
Alright, looks good. Nice one.

PR #1771, opened 2021-08-28
Added tooltip to submit review button

UnrealApex:
Added tooltip to submit review button. @Miodec I was wondering about your thoughts on replacing the icon for the submit button with one of these so it is more intuitive:
[Sign in alt](https://fontawesome.com/v5.15/icons/sign-in-alt?style=solid)
[check-circle](https://fontawesome.com/v5.15/icons/check-circle?style=solid)


static/index.html
```diff
@@ -201,7 +201,12 @@
             <i class="fas fa-star"></i>
           </div> -->
         <div class="icon-button submitButton">
-          <i class="fas fa-chevron-right"></i>
+      <span
+          aria-label="Submit review"
+          data-balloon-pos="up"
+        >
+        <i class="fas fa-chevron-right"></i>
+      </span>
```
Miodec, line 209: i think the tooltip should be on the whole button, not just the icon inside it

Miodec (changes requested):

Issue #1755, opened 2021-08-23
No accent mode


Miodec:
Add a dictionary for each language of the accented characters and their normal counterparts.

With this mode enabled, the accented characters should be replaced with normal ones.

Im not sure what happens with PB's when this mode is enabled. Disabling accents will make everything considerably easier, so PB's might have to be disabled. Open to suggestions on how to approach this.

noitatum:
Aren't PB's tracked separately per dictionary? Why would it matter if you get a faster time using this mode, isn't that kind of the point?
Currently in Spanish for example, accentuated characters are counted as one, even though they are obviously two key presses (And even if that wasn't the case the accent key press is the slowest on the keyboard by far). Getting a PB in Spanish ends up being more RNG than say English. If you get a bad stride of accentuated words you will get a worse than average score. Playing in this mode will be way less frustrating, even if it's technically "butchered" Spanish.

Miodec:
> Aren't PB's tracked separately per dictionary?

Yes they are, but the idea is that with this mode you *don't* have to duplicate every list on the website to remove all the accents. So, both no accent and accent Spanish lists would be referenced as "Spanish" in the config.

noitatum:
> > Aren't PB's tracked separately per dictionary?
> 
> Yes they are, but the idea is that with this mode you _don't_ have to duplicate every list on the website to remove all the accents. So, both no accent and accent Spanish lists would be referenced as "Spanish" in the config.

If that's the case then I believe it would still be nice to have PB's in a separate "no accent" leaderboard.

Miodec:
Added lazy mode

PR #1778, opened 2021-08-28
Removed diacritics from Albanian Words

deniscerri:
Edited both Albanian and Albanian_1K by removing diacritics. 

Most the population doesn't use them while writing so it was best to replace them with normal letters. :) 
That's about it.


Miodec:
I think this should be part of #1755, instead of modifying the wordlist for everyone.

PR #1779, opened 2021-08-29
Add exit condition for username loop

alex-1121:
### Description

When user uses Google Sign In to login for the first time, he is prompted to choose a username. If prompt is closed (cancel pressed), request is sent with `{name: null}`. Check fails and prompt reappears. User is now stuck in this loop, unless a valid username is provided. There might be reasons why user would choose not to proceed with this, so let's provide an option not to. If prompt is closed, break the loop and sign out.

### DIFF
--- a/src/js/account-controller.js
+++ b/src/js/account-controller.js
@@ -103,6 +103,7 @@ export async function signInWithGoogle() {
 
         if (name == null) {
           AccountController.signOut();
+          $(".pageLogin .preloader").addClass("hidden");
           return;
         }
 
--- a/src/js/account-controller.js
+++ b/src/js/account-controller.js
@@ -2,6 +2,7 @@ import * as Notifications from "./notifications";
 import * as UpdateConfig from "./config";
 import * as AccountButton from "./account-button";
 import * as Account from "./account";
+import * as AccountController from "./account-controller";
 import * as CommandlineLists from "./commandline-lists";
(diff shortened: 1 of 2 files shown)


Miodec:
This is intended. An account needs to have a name.

alex-1121:
I think there's still something wrong with UX.

I have multiple Google accounts, one of them I use for Monkeytype. I accidentally logged in with a wrong one, realized it after seeing username prompt, began looking for a way to sign out, and got myself locked out for a day.

Something that wasn't mentioned initially, is that it is pretty easy to reach RateLimit in that loop. RateLimit is 60 per hour for `/user/checkName`, but it is just 3 per day for `/updateName`. If user comes from login page, `/checkName` is queried, but if in an effort to get rid of that prompt user refreshes page, queries are sent to `/updateName` from there on. And it takes just 3 queries to reach RateLimit.

That's why to me it seems best to let user sign out by pressing Cancel on username prompt. There is a legit usecase where user wouldn't like to proceed with registration. Other solution could be to do something about that very low RateLimit.

Miodec:
Right, in that case instead of signing out I think it should also delete the newly created account document in the database.

Ill update the limits on the entpoints.

Miodec:
Nevermind, the database document isn't even created at that point in time

Miodec (approved):

PR #1786, opened 2021-08-30
added Persian layout and Lalezar font for RTL languages

rendinjast:
### Description

- add  Persian (standard) layout 
- font support for RTL languages ([Lalezar Font](https://github.com/BornaIz/Lalezar))


src/sass/style.scss
```diff
@@ -2154,6 +2155,7 @@ key {
         .word {
           //flex-direction: row-reverse;
           direction: rtl;
+          font-family: "Lalezar" !important;
```
Miodec, line 2158: Why force the font?

rendinjast: I was trying some different font and forgot to remove that. My bad.

PR #1800, opened 2021-08-31
Fix bug in update name time limit

mrbrianevans:
I noticed that the rate limit for changing name on account was incorrectly set to 30 days in seconds, but Date.now() returns milliseconds, so the limit would actually be 0.03 days instead of the intended 30 days.

The only thing I've changed is multiplied the delay by a thousand so that its in milliseconds.

```
30 * 24 * 60 * 60 * 1000
>> 2592000000
```


Miodec:
Good catch

PR #1804, opened 2021-09-01
Increase roundness

MysteriousLog6:
image
Rounded                                        vs               Non-Rounded

Makes it look a bit more modern , can be reduced to 0.5 instead of 0.75 , 0.5 is more subtle.


Miodec (approved):

Miodec:
Well go with 0.5

MysteriousLog6:
Sure!

PR #1817, opened 2021-09-06
added 3 more file for italian language

marcoferrari2409:
### Description

Closes #


marcoferrari2409:
the update 8222117 is not valid, it was an old one

Miodec:
I think 660k is a bit much. I would remove that one.

Miodec:
You forgot to remove the 660k from groups and the file itself.

Miodec (dismissed):

PR #1822, opened 2021-09-08
new theme, pritu_palnight

pritudev:
### Description

Closes #


Miodec:
Please include a screenshot when adding themes. Also you forgot to add it to the list file.

Issue #1832, opened 2021-09-10
Pressing the escape key when the page is starting up will cause things to break


uShldGetCeleste:
**Describe the bug** 
If you press the escape key while monkeytype is starting then things will break, usually just visually.

**To Reproduce** 

1. Open the site
2. Before stuff finishes loading press the escape key. The earlier that you press it the more it will be broken

**Expected behavior** 
Page should load properly

**Screenshots** 
![image](https://user-images.githubusercontent.com/68669265/132906885-deb663a3-c592-4bdc-8be2-01571f052e8c.png)
![image](https://user-images.githubusercontent.com/68669265/132906913-9a1d0386-ca4f-48f5-a000-911c81da2528.png)
![image](https://user-images.githubusercontent.com/68669265/132906965-59df6814-e982-4478-a772-829834d62ffa.png)

**Desktop:** 

- OS: Windows 10
- Browser Chrome 
- Browser Version 93.0.4577.63

sphhax:
The escape key in browsers haults all requests so, until the site loads enough to register the escape key as the console button then I am not sure much more can be done. 

The only possible fix would be to put a function at the top of index.html so we can try and load is asap but honestly, this is just normal functionality.

rasulkamolov:
Its been fixed

Miodec:
I keep trying to replicate this and I cant seem to be able to get it.

sphhax:
@Miodec im only able to do it as soon as I refresh the page or by simulating a slower internet connection. The problem likely is that on slower connections it's easier to press escape before the site loads enough to disable it.

UnrealApex:
Try reloading and spamming escape.

UnrealApex:
> @Miodec im only able to do it as soon as I refresh the page or by simulating a slower internet connection. The problem likely is that on slower connections it's easier to press escape before the site loads enough to disable it.

I agree, most computers would be able to load Monkeytype fast enough that these requests can be made before the site is ["interactible"](https://developer.mozilla.org/en-US/docs/Glossary/Time_to_interactive) by the user.

UnrealApex:
Maybe at the top of `index.html` we could add this: 
```html
<script>
document.addEventListener("keydown", function(event) {
  event.preventDefault();
);
</script>
```

sphhax:
[...]

PR #1840, opened 2021-09-12
fixed esc actually stopping requests on slow connections

sphhax:
### Description

Added an escape button blocker much earlier in the code to prevent it stopping requests on slower connections

Closes #

#1832


static/index.html
```diff
@@ -1,6 +1,11 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <script>
+      document.addEventListener("keydown", function(e) {
+         if (e.keyCode == 27) e.preventDefault();
```
UnrealApex, line 6: `event.keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). Use `event.key` instead.

UnrealApex (changes requested):

static/index.html
sphhax: done

Miodec (approved):

PR #1856, opened 2021-09-14
Added Arabic support in lazy mode.

ZER0-X:
I added the support for Arabic in lazy mode, but can i use unicode hex value (like `\u064b`) in `_accents.json` ? Sinse Arabic diacritis don't appear in an organized manner in the text editor.


Miodec (approved):

Miodec:
I just realized I left in the `_accents` file that is no longer used - the accents should be added to `lazy-mode.js`. I'll move them.

PR #1858, opened 2021-09-14
Corrections to British English

Banzobotic:
### Description

I have removed many incorrect changes that were made with the list of differences between the normal english and british english. I have also corrected changes where appropriate and also added mom --> mum

There were a few changes that I have not made yet but I would like people's opinions on:
- While gramme is usually considered to be the british spelling of the word, gram is more commonly used
- While sulphur is generally consider to be the british form, sulfur is used by IUPAC and also many institutions such as most (if not all) uk exam boards

Also many words seem to be North America only with no direct parallel in british english, in the future will there be a way of just removing these from the word selection?

In situations where I found that multiple spellings of a word were valid in both american and british english, I tried to use the spelling already in use by the english language.


Miodec (dismissed):

Miodec:
> While gramme is usually considered to be the british spelling of the word, gram is more commonly used

I think more common is better.

> While sulphur is generally consider to be the british form, sulfur is used by IUPAC and also many institutions such as most (if not all) uk exam boards

Let's stick with sulfur then.

Banzobotic:
I made it so that it now uses sulfur and gram.

Miodec (approved):

PR #1864, opened 2021-09-15
Added words to british english and removed hyphens

Banzobotic:
### Description

This pr adds all the words from the english 10k changes list in #958 and also removes the hyphenated words from the british english list. 

I think that while the hyphens are more correct, they are also more correct in american english, so the british english list is not the place to make this change. Although I think that it is best to stick with no hyphens because hyphens are not fun to type. 

I have also removed some British variants of words from american english lists although this is by no means thorough, the best way to get rid of these may be to create a script that uses the british english words in the list and searches for them in the american english lists, but I have no experience working with json. 

I will look into merging the 450k list from #958 with the current one but the words are placed in the list ["british", "american"] which is the wrong order, is there a way of easily flipping this around? It will take a lot of work to check through the 450k list however so it will be quite some time before I am able to get a finished list from it even once I have it flipped around.


Miodec:
Here is the flipped list you're after 
[changed.txt](https://github.com/Miodec/monkeytype/files/7172755/changed.txt)

Miodec (approved):

Banzobotic:
Can it be done with the formatting the same as the other british english list?

Banzobotic:
Also, if I were to replace an america only word with "" would it make it not show up at all or would it make it show up as an awkward double space section?

Miodec:
I think it would result in a double space, so best to not leave empty words in there.

PR #1866, opened 2021-09-16
completed lazy mode for chinese pinyin

Miigon:
### Description

Lazy mode accents are not complete and lacks some of the characters in chinese pinyin(namely `ǎ`,`ě`,`ǐ`,`ǒ`,`ǔ`,`ü`). This PR added those.

`ü` is typically typed as `v` in chinese keyboard layout so it's mapped as so.


src/js/test/lazy-mode.js
```diff
@@ -1,9 +1,10 @@
...
-  ["éèêëẽęę́ēę̄ė", "e"],
-  ["íìîïĩįį́īį̄", "i"],
-  ["óòôöøõóōǫǫ́ǭ", "o"],
-  ["úùûüŭũúūů", "u"],
+  ["áàâäåãąą́āą̄ǎ", "a"],
+  ["éèêëẽęę́ēę̄ėě", "e"],
+  ["íìîïĩįį́īį̄ǐ", "i"],
+  ["óòôöøõóōǫǫ́ǭǒ", "o"],
+  ["úùûüŭũúūůǔ", "u"],
+  ["ü", "v"], // as in chinese pinyin
```
Miodec, line 7: Problem is that `ü` is already replaced in the line above, so this will not work. I would need to come up with some sort of per language override, or completely reword lazy mode

ZER0-X: It would be great if there were levels too, not just `on` or `off`.

Miodec:
Alright, i added an accent override support in e3c8177802b31d0965f92383377d4038523caa9c.

What you need to do is add an 'accents' key to the languages that you want to override the accents in, for example:

```js
{
  "name": "spanish",
  "leftToRight": true,
  "accents": [
     ["letters to replace", "character"]
  ],
  "words": [
    "como",
    "su",
    "que",
    "él",
    "era",
```

Miigon:
Just realized language-specific override accent list would completely override the global accent list, not just part of it.😅

This should do the job.

Miodec (approved):

PR #1867, opened 2021-09-16
Remove extra space in "Sign In" link

driazati:
### Description

HTML treats newlines as whitespace and collapses it + the spaces down to a single space which has the underline. This fixes this by tightening up the `<span>`s

image

to

image

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -1140,9 +1140,7 @@
                 <canvas id="wpmChart"></canvas>
               </div>
               <div class="loginTip">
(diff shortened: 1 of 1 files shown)


UnrealApex (approved):

Miodec:
There is a very high possibility that its Prettier that inserts that break, and this PR will end up getting reversed when Prettier runs, but I guess we will see

Miodec (approved):

PR #1872, opened 2021-09-16
get and execute config actions from object instead of ifs

MarksASP95:
Discalimer: this is my first time contributing to open source code, so any feedback or observation is appreciated.

Just made a little refactoring in the config.js file. Some stuff was being executed depending on the mode using if statements, which I believe are unnecesary and could be improved by refactoring them into an object in which each key is a mode and each value is a function of the actions taken. Then the mode action saved in the `configAction` variable and is then executed.

### DIFF
--- a/src/js/config.js
+++ b/src/js/config.js
@@ -228,67 +228,78 @@ export function setMode(mode, nosave) {
   config.mode = mode;
   $("#top .config .mode .text-button").removeClass("active");
   $("#top .config .mode .text-button[mode='" + mode + "']").addClass("active");
-  if (config.mode == "time") {
-    $("#top .config .wordCount").addClass("hidden");
-    $("#top .config .time").removeClass("hidden");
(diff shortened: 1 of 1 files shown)


Miodec:
I'm not sure if this is necessary? What's the advantage of doing things this way?

MarksASP95:
It's just minor refactoring and it and this allows the code not to go through all ifs in case it's zen, and makes it easier to add new modes. I know it's not critical for performance, it's just a recommendation that looks cleaner (in my opinion)

Miodec:
> allows the code not to go through all ifs in case it's zen

Don't you have to do pretty much the same, looking for the key:value pair for zen?

I don't know. I don't feel like I would do something like this myself, and besides there are many examples in the codebase that look exactly the same, and changing just one of them picked at random seems more confusing to me. Sorry, I think ill just keep it the way it is for consistency

PR #1881, opened 2021-09-18
Added Custom Color Inputs that you can Paste into

### LINKED ISSUE #432: replace color inputs with js ones that support copy and paste

Estebene:
### Description
This is a attempt to resolve #432 by adding text inputs and buttons to the custom color menu. This is a draft since I have not tested it and there is still some functionality I would like to add (like other color format paste as-well). Feedback on the premise of this solution would be much appreciated.

Closes #432

### DIFF
--- a/src/js/settings/theme-picker.js
+++ b/src/js/settings/theme-picker.js
@@ -65,15 +65,18 @@ export function refreshButtons() {
 
 export function setCustomInputs() {
   $(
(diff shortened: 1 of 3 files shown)


Miodec:
Looks pretty good so far. Couple things:

- Picker shows up in top left of the page 
image


- I kinda miss the colorful pickers. Maybe do something like this? 
image

PR #1887, opened 2021-09-19
Chaos Theory Theme

mTopy:
### Description

Uses a dark background and pink main color with many flipped words and letters, modifies the default caret with an image from the set and adds a logo to the theme picker.
[Link to GH post](https://geekhack.org/index.php?topic=114642.0)

use for post


Miodec:
Thanks, congrats on the gb

PR #1897, opened 2021-09-22
Expose layouts globally to make editing layouts easier.

angularorbit:
This would greatly simplify, e.g. an extension to add a new customized layout.

### Description

Basically exposes the layouts list as a global variable.

### DIFF
--- a/src/js/layouts.js
+++ b/src/js/layouts.js
@@ -1,4 +1,4 @@
-const layouts = {
+const layouts = window.layouts = {
     default: {},
     qwerty: {
         keymapShowTopRow: false,


Miodec:
Im not sure whats the benefit of this

angularorbit:
It is meant to allow someone to add his own layout via a browser extension by modifying the `layouts` object, which would otherwise be difficult to extract using js, as it seems the whole of monkeytype.js is inside a webpack closure, and the layouts variable is inaccessible (without a debugger). I have successfully been able to modify the layout by adding a break point at the declaration and then exposing the variable. But I still don't know how it is possible to do that automatically via an extension, Thus I put the pull request...

Miodec:
But, why would you want to modify the layout at runtime? Why not just add it to the layouts file?

angularorbit:
Because I haven't set up a dev server in my machine.

Miodec:
You don't even need a local dev server to create layouts, just edit the layouts file. Also, setting up local dev server is not difficult, I don't think exposing layouts globally is needed.

PR #1916, opened 2021-09-29
Create japanese_hiragana.json

sashinexists:
A list of common japanese words written in Hiragana only. This works well with the Mozc and KKC layouts and means conversion isn't necessary, therefore all of Monkey Type's accuracy features can be used.

### Description

Closes #


Miodec:
Duplicate of #1919

PR #1917, opened 2021-09-29
Create japanese_katakana_common.json

sashinexists:
A list of commonly used loan words borrowed from different languages written in Japanese. All of them are written in katakana. This should work well with the Japanese (Kana) layout

### Description

Closes #


Miodec:
Duplicate of #1919

PR #1925, opened 2021-10-01
Separate Arabic words with diacritics 

darkstego:
### Description

Recent changes have made it impossible to practice Arabic without diacritics. Since most of the Arabic typed is without diacritics (or in some cases minimal diacritics), these recent changes severely reduces the usability of the site for what most Arabic users want to practice. This change splits the word list into those with and without diacritics. This gives the user the flexibility to decide how they want to practice.


ZER0-X:
Turn on the `lazy mood`, no need for separate list.

darkstego:
That is

PR #1926, opened 2021-10-01
added b-w theme and edited _list.json

AadhavanPL:
added b-w theme. edited _list.json and added b-w.css


Miodec:
I think its too similar to `dark` theme.

PR #1944, opened 2021-10-04
Fix quote search popup width

CrumblyLiquid:
Makes the width of the quote search responsive to the width of the screen

This is how it looks currently:
image

This is how it would look:
image

(The window is the same size across the screenshots.)

### DIFF
--- a/src/sass/popups.scss
+++ b/src/sass/popups.scss
@@ -400,7 +400,7 @@
     padding: 2rem;
     display: grid;
     gap: 1rem;
-    width: 1000px;
+    width: 80vw;
     height: 80vh;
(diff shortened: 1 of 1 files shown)


src/sass/popups.scss
```diff
@@ -400,7 +400,7 @@
     padding: 2rem;
     display: grid;
     gap: 1rem;
-    width: 1000px;
```
Miodec, line 403: this will cause the popup to be extremely wide on wide monitors

Miodec (changes requested):

Miodec (approved):

PR #1966, opened 2021-10-07
Make Tab and Esc behavior consistent even with swap

SeerLite:
Edit: This PR also refactors and cleans up some messy code I left in after input rewrite

---

When typing in zen mode or a quote that contains a tab character:
#### No swapEscAndTab
* **Tab**: Inputs tab character
    * **Shift + Tab**: Selects restart button (or restarts with quick tab)
* **Escape**: Opens command line

All good.

#### With swapEscAndTab
* **Tab**: Opens command line
* **Escape**: Inputs tab character
    * **Shift + Escape**: Selects restart button (or restarts with quick tab)

Makes sense, it literally "swaps" all of their behavior, but is unintuitive.

#### No swapEscAndTab
Same as before.

#### With swapEscAndTab
* **Tab**: Inputs tab character
    * **Shift + Tab**: Opens command line
* **Escape**: Selects restart button (or restarts with quick tab)

Main functionality is swapped, but Tab still inserts a tab character. More intuitive.

### DIFF
--- a/src/js/input-controller.js
+++ b/src/js/input-controller.js
@@ -570,12 +570,15 @@ function focusRestartOrRestart(event) {
     }
 
     if (Config.quickTab) {
-      if (event.shiftKey) {
-        ManualRestart.set();
+      if (event.key === "Tab") {
+        if (event.shiftKey) {
+          ManualRestart.set();
+        } else {
+          ManualRestart.reset();
+        }
       } else {
-        ManualRestart.reset();
+        ManualRestart.set();
       }
-
       if (
         TestLogic.active &&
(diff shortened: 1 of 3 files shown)


SeerLite:
I was also thinking about replacing "swap tab and esc" with a "restart mode" setting with only these options:
* **Manual**: no quick tab and no swap.
* **Quick Tab**: quick tab but no swap.
* **Quick Esc**: quick tab and swap.

Escape can't be used to cycle elements (we can't emulate what Tab does, at least not easily), so its only functionality when swapping without quick tab is to focus the restart button anyway.

What do you think?

PR #1968, opened 2021-10-08
Add .nvmrc file and updated Node.js installation instructions

mtthwn:
### Description

`nvm` is a tool used for handling multiple versions of Node.js on a user's machine. Adding an `.nvmrc` file allows contributors using `nvm` to quickly switch to the version of Node.js that Monkeytype supports, and also warns them if they do not have that version installed. This is a relatively small change, but can make development on Monkeytype easier for Node.js developers who contribute to multiple projects.

The proposed standardized version of Node.js in the `.nvmrc` file is set to v14, but please let me know if the team prefers to use a different one for development with Monkeytype.


Miodec:
What about specifying the npm version?

PR #1971, opened 2021-10-09
Fix results page stats css for mobile

lukew3:
### Description
I changed css media queries on smaller screens to display wpm and acc stats side by side instead of on top of each other. Used code that already existed for doing this at a larger size, which was not copied for smaller sizes for some reason.

Before:
Screenshot from 2021-10-09 04-28-40
After:
Screenshot from 2021-10-09 04-27-54

Also, note that there are still some issues with mobile ui that should probably be fixed, but I just wanted to get this one commit in. For example, test type selection is cut off and buttons below results don't use all the space they are afforded.

### DIFF
--- a/src/sass/media-queries.scss
+++ b/src/sass/media-queries.scss
@@ -293,22 +293,6 @@
       }
     }
   }
-  #middle {
-    #result {
-      grid-template-areas:
-        "stats stats"
-        "chart chart"
-        "morestats morestats";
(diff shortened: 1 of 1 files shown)


lukew3:
Edit: realized the fix doesn't require me to paste already written css. Just needed to remove a couple of lines that were overwriting the correct css.

PR #1976, opened 2021-10-09
Add .editorconfig

SeerLite:
### Description
EditorConfig is a configuration file to standardize file format settings across editors. Most code editors have the ability to use it natively while others have community plugins.

* [VSCode plugin available here](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)
* [Plugins for other editors available here](https://editorconfig.org/#download)

See the [EditorConfig homepage](https://editorconfig.org/) for more information.

For now I just added configuration for indentation size and format, and I think that's pretty much all we're gonna need. They are the only things I had to constantly set up when opening my editor to work on Monkeytype, and with EditorConfig it's automatic :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*.{html,js,css,scss,json,yml,yaml}]
+indent_size = 2
+indent_style = space


SeerLite:
No, this sets editor configuration.

If an user's editor uses tabs by default, and these tabs are e.g. 6 characters long visually, it will look wrong when editing, even when indenting correctly.

Prettier auto-formats a bunch of other stuff but does that at commit time. It can't solve the above.

EditorConfig will, and it's a standard used in a lot of popular projects.

SeerLite:
Added yaml files (for `.github` configs)

PR #2025, opened 2021-10-21
Settings from URL

CrumblyLiquid:
This PR builds from what was done in #1394 and also probably closes #1394.

You can add parameters to URL that if recognized will be applied to settings.

The `?` separates settings from the rest of the URL and `&` separates settings from each other. Each setting has key as its name followed by `=` and the desired value.
Example URL might look like this `https://monkeytype.com/?mode=words&words=25&difficulty=normal` or like this `/?mode=custom&custom=Hi|hi|hi!`.

### DIFF
--- a/src/js/url.js
+++ b/src/js/url.js
@@ -288,13 +288,10 @@ export function loadFromUrl() {
 
   window.history.replaceState(null, null, url); // update the URL with new params
   // ui.js:129 prevents params from being shown for longer time
-  console.log(url);
 
   let appliedSettings = 0;
   // Search for and apply settings
   Object.keys(Settings).forEach((name) => {
-    console.log(name);
-    console.log(Settings[name].options);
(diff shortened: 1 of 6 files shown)


CrumblyLiquid:
This is not yet ready for merge since options don't load for `language` and `theme` settings.

src/js/url.js
```diff
@@ -0,0 +1,316 @@
...
+      1: [0.0, 2.0], // Brightness
+      2: [0.0, 2.0], // Saturation
+      3: [0.0, 1.0], // Opacity
+    };
+  }
+
+  set(val) {
+    let arr = val.split(this.sep);
+    // input array is incomplete
+    if (arr.lenght !== this.values.lenght) return false;
```
MichaelDeutschCoding, line 159: typo "length"

PR #2027, opened 2021-10-22
Reordered language list in alphabetical order in _groups.json

tyc45:
### Description

Closes #


Miodec:
I would prefer the languages to be kept in their current order - which is roughly by their popularity.

PR #2034, opened 2021-10-25
Fixed Errors

Estebene:
Just fixed a few errors that show up when running npm run start:dev


Estebene:
Ok, looks good

Estebene:
Sorry, it took so long to fix, but here it is. I tested it to make sure it works:
image

Just checking, we are meant to be removing normal spaces aswell right?

Miodec:
> Just checking, we are meant to be removing normal spaces aswell right?

No, if you check the original function it replaces characters based on key-value pairs. This means you would add the \u codes to the object as keys and normal spaces as values.

PR #2062, opened 2021-11-05
Incement accuracy even for trailingComposeChars

pooooooooods:
Found the root cause for https://github.com/Miodec/monkeytype/issues/2057

Even though the bug is fixed with the change from Mio, typing \` will still keep the accuracy as 100 whereas a wrong character will drop it to 0. This fix will make sure \` will also be counted during calculating accuracy.

### DIFF
--- a/src/js/input-controller.js
+++ b/src/js/input-controller.js
@@ -398,15 +398,15 @@ function handleChar(char, charIndex) {
     char +
     TestLogic.input.current.substring(charIndex + 1);
 
+  MonkeyPower.addPower(thisCharCorrect);
+  TestStats.incrementAccuracy(thisCharCorrect);
+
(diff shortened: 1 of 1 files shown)


Miodec (approved):

SeerLite:
Oops, should have commented earlier on this one, sorry.

This PR will break accuracy calculation.
Easy way to reproduce is having a custom test with text "ésto" and then typing it correctly. The `trailingComposeChars` (aka the accent `´` will lower the accuracy since `"´" != "é"`

@Miodec please revert this one. The right way to fix it is to just display 0% instead of the real acc when it's NaN.

Sorry again for not responding and reviewing this earlier.

Miodec:
Thanks @SeerLite

Issue #306, opened 2020-08-21
Add Wikipedia Mode


madnight:
I would like to type Wikipedia articles. This mode should be like the quote mode, but instead of quotes it fetches the texts from Wikipedia.

As for the technical aspect, I could either imagine to do a live query to the Wiki API to obtain the texts or to use one of the text-only Wiki dumps.

rolfvreijdenberger:
+1 and abstract it away so it would be possible to integrate more data providers. how-to-type.com has an affiliate program with amazon.com to type quotes.

UnrealApex:
Might be possible accessing Wikipedia articles using the Wikipedia api or via web scraping. 
https://stackoverflow.com/questions/7185288/how-to-get-wikipedia-content-using-wikipedias-api
https://stackoverflow.com/questions/53127383/how-to-pull-data-from-wikipedia-page
https://www.freecodecamp.org/news/scraping-wikipedia-articles-with-python/

Liquidmantis:
@Miodec I actually came to open an issue for the Amazon affiliate link.  Plenty of times doing quote typing exercises I'm intrigued by the quote, look at the source, and I'm surprised it's not a link.

TheSmallTeaBoi:
Hey, any news on this? I think this is an awesome idea!

Jonny-exe:
I would be happy to work on this.

PR #2069, opened 2021-11-07
Added wikipedia mode to funbox

Jonny-exe:
### Description

Added wikipedia mode to funbox. In this mode you get random sentences from wikipedia articles, using wikipedia's [API](https://www.mediawiki.org/wiki/API:Main_page).

Closes #
#306

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -179,6 +179,7 @@ const refactoredSrc = [
   "./src/js/test/test-config.js",
   "./src/js/test/layout-emulator.js",
   "./src/js/test/poetry.js",
+  "./src/js/test/wikipedia.js",
   "./src/js/test/today-tracker.js",
   "./src/js/test/weak-spot.js",
   "./src/js/test/wordset.js",
--- a/src/js/test/test-logic.js
+++ b/src/js/test/test-logic.js
@@ -30,6 +30,7 @@ import * as Replay from "./replay.js";
 import axiosInstance from "./axios-instance";
 import * as MonkeyPower from "./monkey-power";
 import * as Poetry from "./poetry.js";
+import * as Wikipedia from "./wikipedia.js";
 import * as TodayTracker from "./today-tracker";
 import * as WeakSpot from "./weak-spot";
 import * as Wordset from "./wordset";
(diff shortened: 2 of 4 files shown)


Miodec:
That was quite fast, testing now.

Miodec:
So, biggest issue i see so far is that if the section is very short, and your mode is set to time 30 for example, after running out of Wikipedia words it will just revert to adding random words.

Writing it just now i realised that it will fail regardless of how long the Wikipedia text is after 100 words, because it doesn't work with the `addWord` function. I remember i ran into this issue when trying to fix the poetry funbox. Im not sure whats the approach here.

Jonny-exe:
I hadn't considered that. Maybe loading another text whenever the user is close to finishing? Or maybe just adding the same text from the start? I'll try to think of something. I'll also have a look at the addWord function and try to make it work with wikipedia sections.

Jonny-exe:
I'm not able to reproduce the error with `addWord`. I have tried: 
1. Going to funbox, wikipedia.
2. Selecting time mode
Whenever the section is to short random words are added until the time is finished.

What's the expected behaviour?

Miodec:
Well, i dont know what should be the expected behavior. It doesn't make sense to have a shorter text than how long the test is. And reverting to random words seems... not like the solution. Im not sure what to do here, and the poetry funbox.

Jonny-exe:
Would loading a new Section be a viable solution?
Instaed of loading random words you could load a new section whenever the user is close to finishing.

Miodec:
Hm.. Maybe it could work.

Jonny-exe:
I changed it, so it will fetch a new section whenever the user has 20 words left. Also (in words mode), at the beginning it will fetch as many sections as it needs until it has reached the desired words amount.
I think it is a viable solution, but tell me what you think. If it is ok, I can do the same for poems.

thuongtruong1009 (approved):
Well done. Keep contributing.

Miodec:
image

Broke on a 30 second test when the wiki article was short.

Miodec:
Adding more words works well. Im still thinking how it would perform with slow internet, but i guess its fine for now. I think for time mode i would always make it generate at least 100 words so it fills the screen correctly.

image

If you're feeling extra awesome, you could also make this change for the poetry funbox, but its up to you.

Miodec (approved):

Miodec:
Nice work

PR #2071, opened 2021-11-08
Remove VERY inappropriate russian text

DanInSpace104:
### Description

Removed very inappropriate quote from russian.json


austintraver:
When I saw this, at first, I was skeptical of what constitutes "VERY inappropriate", so I took a look at the diff. Once I found the removed passage, I translated the Russian text into English and... 

😦

Good catch.

Miodec:
I need to find who the heck added that. Thanks

PR #2075, opened 2021-11-09
Resolve Miodec/monkeytype#2074

austintraver:
### Description

Users requesting to use the font `Comic Sans MS` would instead be presented with `Helvetica` instead. 

---

Resolves Miodec/monkeytype#2074


Miodec:
Thats intended, its a designer joke.

PR #2082, opened 2021-11-12
Patch 1

samuelxyz:
### Description

Adds more layouts


Miodec:
Dang, seems like a lot of really obscure layouts? How does it look on the settings page? Isnt it too busy now? Maybe we need to add some sort of grouping

thuongtruong1009 (dismissed):
🔥

samuelxyz:
> How does it look on the settings page? Isnt it too busy now?

It's still much less than the number of themes, so by that comparison I think it's not too bad.

Miodec:
Alright, thanks

PR #2088, opened 2021-11-15
Introduce auto switching between light and dark theme

### LINKED ISSUE #1629: autoswitch between light and dark theme

ashutoshc8101:
### Description

Closes #1629

Introduces auto-switching between light and dark mode as per system theme and also provides button in settings to toggle this feature.

Video:

https://user-images.githubusercontent.com/24855641/142440340-9d5f3a21-63c5-42ae-ab3d-3b97d5c45ad9.mp4

### DIFF
--- a/src/js/commandline-lists.js
+++ b/src/js/commandline-lists.js
@@ -664,6 +664,29 @@ let commandsFlipTestColors = {
   ],
 };
 
+let commandsAutoSwitchTheme = {
+  title: "Auto switch theme accroding to system",
+  configKey: "autoSwitchTheme",
+  list: [
+    {
+      id: "setAutoSwitchingOff",
+      display: "off",
+      configValue: false,
+      exec: () => {
+        UpdateConfig.setAutoSwitchTheme(false);
+      },
+    },
+    {
+      id: "setAutoSwitchingOn",
+      display: "on",
+      configValue: true,
+      exec: () => {
+        UpdateConfig.setAutoSwitchTheme(true);
+      },
+    },
+  ],
+};
+
 let commandsSmoothLineScroll = {
   title: "Smooth line scroll...",
   configKey: "smoothLineScroll",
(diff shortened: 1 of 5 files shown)


Miodec:
Good start. For the settings page I'm thinking a section that has an on off button for enabling the auto switching, and then a drop down with all the themes.

ashutoshc8101:
@Miodec Please take a look.

Miodec:
Couple things:

- I dont see a way of selecting which theme is for light mode and which one is for dark mode?
- The theme buttons are duplicated
image
- Looks like the setting is not saved to config, it doesn't persist after refreshing the page - it resets to off

ashutoshc8101:
@Miodec Thanks for the response.

I have made some progress and have a few questions.

> Looks like the setting is not saved to config, it doesn't persist after refreshing the page - it resets to off
> It would be nice to have a notification if the theme switches during runtime

Done.

I am unable to reproduce the second bug. Can you share a video of this happening?

> I don't see a way of selecting which theme is for light mode and which one is for dark mode?

Do we have similar dropdowns present in the application, If yes, can you reference that?
Knowing this will be helpful.

Miodec:
> Do we have similar dropdowns present in the application, If yes, can you reference that?
> Knowing this will be helpful.

Custom > change > words filter > language dropdown.

> I am unable to reproduce the second bug. Can you share a video of this happening?

Ill see if it happens again after you recent changes

Miodec:
Yeah, double theme buttons are fixed. For the general usage I'm thinking if the setting is enabled:
- Load the light/dark theme on page load according to the selected themes in the drop down. 
- Allow the user to change theme as usual - this doesn't affect the new config.
- If the system makes a theme change and the website picks up the event, change the theme if the test is NOT active. If it is active, stop that change until the test is completed.

deter0:
What does the "Only works on Windows and Mac" mean? `Window.matchMedia` is supported on almost all browsers ([Compatibility](https://caniuse.com/matchmedia)) And `prefers-color-scheme` works on most main-stream browsers just excluding a few. ([Compatibility](https://caniuse.com/prefers-color-scheme))

ashutoshc8101:
> Any update?

I was a bit busy for the past few weeks. I will definitely complete this and push new changes by 31 Dec.

Miodec:
Any update? Should i continue this?

PR #2102, opened 2021-11-19
fixed length type for english quote

whorfian:
A quote had a different type for its `length` field than the other quotes. This broke a parser I was writing.


Miodec:
Nice catch

Miodec (approved):

PR #2117, opened 2021-11-23
Updated contributors list with nov 2021 data

webtax-gh:
### Description

I noticed that there was a disparity between the website's list of contributors and github's list. As such, I have updated the data using the github api.


Miodec (approved):
Thanks

PR #2122, opened 2021-11-25
Added Phantom Theme

justussoh:
### Description

Added GMK Phantom Color Theming 

SS of New Theme
phantom

Not sure if this is too close to Future Funk so added it for comparison
ff
 
Purples are a tad bit darker, main text in yellow instead of off white.

### DIFF
--- a/static/themes/_list.json
+++ b/static/themes/_list.json
@@ -623,5 +623,10 @@
     "name": "tiramisu",
     "bgColor" : "#CFC6B9",
     "textColor":"#C0976F"
+  },
+  {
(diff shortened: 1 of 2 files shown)


Miodec:
I would make the purple darker, and switch the text color to white. The error color might be not really visible, especially with colorful mode or flip test colors mode. 

Please check the guidelines: https://github.com/Miodec/monkeytype/blob/master/CONTRIBUTING.md#theme-guidelines

Nugget1308:
Make the purple a little bit brighter.

PR #2142, opened 2021-12-03
Enable PWA

lukew3:
### Description
Website is now an installable Progressive Web App. I've only added the minimum necessary files to allow the site to be installed locally. Explanations for what I did and why can be found at https://mobiforge.com/design-development/pwa-minimus-a-minimal-pwa-checklist. 

Basically, the site just works outside of the browser, no extra functionality added. You can still type offline because the site files have been cached, but a red alert saying script error shows up on page load. Haven't really dug into it but it looks like the tests might be uploaded once connection is restored. 

Edit: I'm an idiot, offline doesn't work. I was running on localhost, so when I turned wifi off, it still worked because localhost doesn't depend on wireless connection.

Also, a better 512x512 icon should probably be added as well as more sizes, I just upscaled the favicon to make a single 512 icon.

#### Closes
#1018 and similarly #1051 #2036 #2126


Miodec:
So, how do you actually "install" it? Where can i find the function?

lukew3:
> So, how do you actually "install" it? Where can i find the function?

When you're viewing the website in chrome, there will be a group of buttons inside the url bar on the right side. The install button should be the leftmost buttoning this group. Clicking on this should allow you to install the app

UnrealApex:
PWAs are supported in other non chromium browsers like Firefox correct?

Estebene:
Not without a bit of work, you can install a program to do it.
https://techdows.com/2021/09/install-pwas-firefox-desktop.html

UnrealApex:
Most non technical users probably wouldn't do any extra steps to get it working so an actual native app alongside the website would also be useful.

Miodec (approved):
Thanks!

jweickm:
I had been enjoying the PWA feature of Monkeytype but recently I can no longer install it. Have tried it on Brave and Microsoft Edge browser on two different machines (win 10 and win 11). Has this feature been disabled? Other websites like https://superpwa.com/doc/test-pwa-on-desktop/ show the PWA install button normally.

Miodec:
> I had been enjoying the PWA feature of Monkeytype but as of recent I can no longer install it. Have tried it on Brave and Microsoft Edge browser on two different machines (win 10 and win 11). Has this feature been disabled? Other websites like https://superpwa.com/doc/test-pwa-on-desktop/ show the PWA install button normally.

Yeah, sadly this was causing many issues with caching. Even though fixes were being pushed to the server users would not see them. Ill probably come back to this some time

PR #2150, opened 2021-12-06
Fix Estonian 1k/10k resetting to English

### LINKED ISSUE #2143: Problem with estonian 1k and estonian 10k

- [X] I have cleared my cache

### Current Behavior

Everytime I do estonian 1k or estonian 10k typing test it automatically switches back to english after the test is done (when I press next test). Doesn't happen with regular estonian (the default one).

### Expected Behavior

I expect that everytime I do estonian 1k or estonian 10k typing test I can press next test and continue typing on estonian 1k or estonian 10k without it automatically switching back to english after the test is done.

### Steps To Reproduce

1. Click on the cog wheel
2. Scroll down and select either "estonian 1k" or "estonian 10k"
3. Do a typing test (doesn't matter what time you select)
4. After completing the typing test press "Next test" and you should see that it the text is in english

### Environment

- OS: Windows 7
- Browser: Google Chrome
- Browser Version: Version 96.0.4664.45 (Official Build) (64-bit)

### Anything else?

_No response_

### Config

_No response_

DanGonite57:
### Description

Closes #2143

### DIFF
--- a/static/languages/estonian_10k.json
+++ b/static/languages/estonian_10k.json
@@ -1,5 +1,5 @@
 {
-  "name": "estonian",
+  "name": "estonian_10k",
   "leftToRight": true,
   "bcp47": "et-EE",
   "words": [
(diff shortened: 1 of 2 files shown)


Miodec (approved):
Good catch

PR #2154, opened 2021-12-07
Added korean language

kckang1103:
### Description

Added korean.json, korean_1k.json and updated list and groups correspondingly.


troy51538:
ay zitch

kckang1103:
Tested locally and didn't have issues with input handling. Although with korean.js it had a similar issue with #2143. 

https://user-images.githubusercontent.com/53661230/144982811-d240f05f-26d9-467f-80b7-20998e12ab98.mov

Miodec:
Duplicate of #811, as mentioned there and looking at your video, 2 problems:

- The characters are flashing red even though you are typing it correctly
- The above causes the accuracy calculation to be incorrect

There needs to be a check if the key pressed is 'able to construct' the character that needs to be typed and react accordingly.

PR #2158, opened 2021-12-07
added GMK Mecha_01 theme

batsoup804:
Here's a photo of the Mecha_01 theme.

image

### Description

This theme is based off GMK Mecha_01. It's color scheme is Bright orange, Indigo and Chartreuse green. I included the orange in the icon buttons next to the logo, as well as the errors to give the letters some more pop. As the indigo is the base in the keycap set I made it the background. Since it is quite a light purple I decided to exclude the black modifier from the theme to make it a bit of a lighter theme.

### DIFF
--- a/static/themes/_list.json
+++ b/static/themes/_list.json
@@ -633,5 +633,10 @@
     "name": "modern_ink",
     "bgColor" : "#ffffff",
     "textColor":"#ff360d"
+  },
+  {
+    "name": "mecha_01",
+    "bgColor": "#7a05ff",
+    "textColor": "#a9a9a9"
   }
 ]
--- a/static/themes/mecha_01.css
+++ b/static/themes/mecha_01.css
@@ -1,0 +1,20 @@
+:root {
+  --bg-color: #7a05ff;
+  --main-color: #2aff19;
+  --caret-color: #2aff19;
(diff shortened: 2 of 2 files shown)


Miodec:
I think the purple seems brighter than the set. And im thinking if after making it a bit darker it could be too close to the `joker` theme...

batsoup804:
Joker is much much darker than mecha but I see your point.I'll make the change and we'll see.

Nugget1308:
Instead of using purple, maybe u should use black.

batsoup804:
> Instead of using purple, maybe u should use black.

image
not very attractive... besides purple is a huge part of the set so it's kind of moot without having the main color.

batsoup804:
Finally got back to this theme, sorry for the inactivity. Anyways I tried out the lighter purple and here's how it looks:
image

Miodec:
I dont know, i feel like the color scheme in general is too close to joker, and there isnt enough elements on the website to make the themes different enough from each other.

batsoup804:
sounds good, then I guess we can close Mecha-01.

batsoup804:
all right, sounds good. I guess we can close Mecha-01 with this then.

PR #2159, opened 2021-12-07
Add Spanish with no accent marks language

rodovelazquez-utn:
In the Spanish language, certain words need to have an accent mark over a vowel, for example in the word "también". The accented vowels look like this: á, é, í, ó, ú. In the Spanish keyboard layout we have a dedicated key that we have to press before a vowel in order to put the accent mark on it. So for the word "también" we have to type the sequence of keys " t a m b i ´ e n ". As you can probably tell, it is very common to miss that accent mark and just write "tambien" (with no accent mark). This is technically a typo, but in a non professional or casual context people won't mind about that mistake. That's why I took the spanish.json as a baseline and created the spanish_no_accents.json file. The difference is that the words don't have accent marks. The main reason for this change is that it's very frequent to miss the accent mark on a vowel while doing a typing test, and that counts as an error. I think that with this change, lazy Spanish people like me would be really happy!


Ferotiq:
Isn't this the same as **lazy mode**?

rodovelazquez-utn:
I didn't know that "lazy mode" existed. I tried it and it makes the same effect. Thanks for commenting!! @Ferotiq

PR #2167, opened 2021-12-10
Add the option to use MongoDB credentials

Ferotiq:
### Description

I have added the option to use MongoDB credentials to login to MongoDB. This is for people who have MongoDB locked with a password.

The new `.env` entries are:
1. DB_USERNAME
2. DB_PASSWORD
3. DB_AUTH_MECHANISM
4. DB_AUTH_SOURCE


Miodec:
This isn't really needed as you can just put the credentials into the URI

`mongodb://username:password@ip:port`

Worth adding a comment about it and adding it to the readme though.

Miodec:
As for the result repeating if you want it on a separate pr you need to work on it on a different branch because any changes you make on the same branch will be added to the PR.

I would prefer if PR's didn't have unrelated changes like this.

Ferotiq:
I figured that was the case after I saw the commit was added.

PR #2168, opened 2021-12-11
clean up on french and french 1k

druotoni:
### Description

Multiple words were add in a single Json word, like "vers le bas" instead of 3 differents entries : "vers", "le", "bas".
I also sort french.json and remove some duplicates


Miodec (approved):
Thanks

PR #2170, opened 2021-12-11
Add Retry Saving Result Button

Ferotiq:
Screenshot 2021-12-10 212829

This PR adds a button that will retry to save a result that has failed to save. The red redo button only shows up when a test has failed to save and will hide itself again when the result has saved. When it has succeeded in saving the result it should remove the click listener. I have tested this by disconnecting from the internet, doing a 10 word test, trying the button, reconnecting to the internet, and trying the button again.

There might be some bugs. For example, the button may activate twice. So this still might need some work.

Closes [#2166](https://github.com/Miodec/monkeytype/issues/2166)

### DIFF
--- a/src/js/test/test-logic.js
+++ b/src/js/test/test-logic.js
@@ -1271,10 +1271,10 @@ var retrySaving = {
   mode2: null,
   stats: null,
   consistency: null,
-  listenerAdded: false,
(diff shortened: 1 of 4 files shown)


Miodec (approved):
Thanks

PR #2176, opened 2021-12-13
Fix Image Issue

teddinotteddy:
### Description
Fixed image issue where mtsocial.png wouldn't show on Discord or Twitter.

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -68,7 +68,7 @@
     <meta property="og:title" content="Monkeytype" />
     <meta property="og:url" content="https://monkeytype.com/" />
     <meta property="og:type" content="website" />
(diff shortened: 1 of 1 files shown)


Miodec (approved):

Miodec:
Are you sure this is the correct path? Im not sure if 'static' is needed there?

teddinotteddy:
@Miodec It might not be needed, but I removed that anyway.

Miodec:
Might? So, is this even tested?

teddinotteddy:
@Miodec I'm so sorry. That was very poor choice of words on my part. I was rushing through this as I am in school right now but yes, I just double checked with the new path. It does indeed work. Sorry for the miscommunication.

Miodec:
Alright, sounds good

PR #2190, opened 2021-12-17
Added functionality so that you fail on incorrect opposite shift for master and expert mode

### LINKED ISSUE #2181: Opposite shift error does not fail test on master difficulty

- [X] I have cleared my cache

### Current Behavior

When doing a test on the 'master' difficulty, the test is expected to end instantly after any mistake made during the test. However, if an incorrect shift key is used and a typo is made, there are no repercussions until the when next key is typed, when the test fails. The user can use the wrong shift key as many times as they want and still pass the test as long as they eventually use the correct shift key.

### Expected Behavior

Test fails instantly on wrong shift key used

### Steps To Reproduce

difficulty > master
opposite shift mode > opposite shift
punctuation > on

Type the first character with the wrong shift key. This can be done infinitely many times as long as the correct shift key is used afterwards and the test will not fail.

### Environment

- OS: MacOS 12 Monterey
- Browser: Microsoft Edge
- Browser Version: 96.0.1054.43 (arm64)

### Anything else?

_No response_

### Config

_No response_

Estebene:
### Description

Closes #2181

This pr doesn't solve the problem in a particularly nice way because it may not be obvious to the user how they failed the test. I am not sure exactly how to solve this one however because it seems that the opposite shift functionality and the master/expert mode are at odds which each other. The idea of the opposite shift settings is that is doesn't type incorrect opposite shifts where as the idea of master mode is a focus on typing EVERY letter correctly. So idk, here's my attempt though.
image

### DIFF
--- a/src/js/input-controller.js
+++ b/src/js/input-controller.js
@@ -428,7 +428,7 @@ function handleChar(char, charIndex) {
     }
   }
 
-  if (!correctShiftUsed) return;
+  if (!correctShiftUsed && Config.difficulty == "normal") return;
 
   //update current corrected version. if its empty then add the current char. if its not then replace the last character with the currently pressed one / add it
   if (TestLogic.corrected.current === "") {
--- a/src/js/input-controller.js
+++ b/src/js/input-controller.js
@@ -428,7 +428,7 @@ function handleChar(char, charIndex) {
     }
   }
 
-  if (!correctShiftUsed && Config.difficulty == "normal") return;
+  if (!correctShiftUsed && Config.difficulty != "master") return;
 
   //update current corrected version. if its empty then add the current char. if its not then replace the last character with the currently pressed one / add it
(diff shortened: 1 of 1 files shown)


src/js/input-controller.js
```diff
@@ -428,7 +428,7 @@ function handleChar(char, charIndex) {
     }
   }
 
-  if (!correctShiftUsed) return;
+  if (!correctShiftUsed && Config.difficulty == "normal") return;
```
Miodec, line 431: I think this should return if difficulty is not master. Because currently, as far as i can understand, it will let incorrectly shifted letters through in expert mode, which kinda defeats the purpose of opposite shift.

Estebene:
Totally agree, Fixed

Miodec (approved):
Thanks

PR #2195, opened 2021-12-19
Completely Fix Wikipedia Spacing

### LINKED ISSUE #2112: issue with wikipedia mode

Ferotiq:
### Description

I don't know if what I'm doing isn't great practice but what I'm doing is not taking Wikipedia's way of turning the HTML into text and instead asking for the HTML (removing `explaintext=true` in the URL). Then I replace every instance where there's a new paragraph (`</p><p>`) with a space, and convert the HTML back into text. One problem I see with this is that if there's a pronunciation or a measure with a unit it will remove the space between it (eg. 12 km -> 12km) . This does consistently add the space between paragraphs and does not confuse acronyms, though.

Closes #2112 and extends on PR [#2193](https://github.com/Miodec/monkeytype/pull/2193)

### DIFF
--- a/src/js/test/wikipedia.js
+++ b/src/js/test/wikipedia.js
@@ -30,7 +30,7 @@ export async function getSection() {
       rej(randomPostReq.status);
     }
 
-    const sectionURL = `https://en.wikipedia.org/w/api.php?action=query&format=json&pageids=${pageid}&prop=extracts&exintro=true&explaintext=true&origin=*`;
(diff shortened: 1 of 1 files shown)


druotoni:
Hi !
I also worked on the funbox : https://github.com/Miodec/monkeytype/pull/2195
Parsing this wikipedia text will be challenging.

Miodec (approved):
Thanks

PR #2198, opened 2021-12-20
wikipedia funbox with current group language

druotoni:
### Description

The wikipedia funbox is a great feature but it's only for english language.
I added a support for spanish, french, german, portuguese, italian and dutch.

The language group is associate with the appropriate TLD for the wikipedia request.
I haven't included other language because I'm not sure if the wikipedia text is correct or "typable" in these langages.

Maybe it could be helpfull to add this tld information in the _groups.js structure in the futur, instead of a big switch in the fonction. But this data in not really usefull outside of the wikipedia funbox.

I removed the "remove non ascii" part because it deletes lot of diacritics.
In my opinion, it's better to have to skip sometime than have to type non-sense words. 
sectionText = sectionText.replace(/[\u{0080}-\u{10FFFF}]/gu, "");

An alternative could be cleaning the wikipedia input with some specific regex for each language to match the corpus. 

Closes #


Miodec (approved):

Ferotiq:
This overrides the whitespace fix, removing `explaintext=true&` will fix that, else lines 82-83 are redundant.

Miodec:
Whoops, will fix.

PR #2204, opened 2021-12-22
Added italian theme

Alenygam:
### Description

Added the Italian Theme

### Screenshots

image
image

### DIFF
--- a/static/themes/_list.json
+++ b/static/themes/_list.json
@@ -1,5 +1,10 @@
 [
   {
+    "name": "italian",
+    "bgColor": "#009344",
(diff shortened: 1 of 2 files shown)


Horix21 (approved):

Miodec:
I think the colors dont work well together. Plus, I don't want to add themes like this, because then people will ask for all sorts of themes that feature their home country's flag. Sorry.

ryujinscales:
> I think the colors dont work well together. Plus, I don't want to add themes like this, because then people will ask for all sorts of themes that feature their home country's flag. Sorry.

True

PR #2207, opened 2021-12-25
Added some English quotes

xia0:
### Description

Closes #


UnrealApex (changes requested):
You are missing the quote `length`(how long in characters each quote is) property in the entries you have added.

xia0:
> You are missing the quote `length`(how long in characters each quote is) property in the entries you have added.

Thanks for the heads up.

Miodec (approved):

Miodec (approved):

PR #2220, opened 2021-12-28
Add scrolling to top after test restarts

Ferotiq:
### Description

With a large font size or a high word count, the page would go down to the bottom upon restart, this was very annoying for me. So, I added a simple line to scroll to the top of the page after the test restarts.


Miodec (approved):
Thanks

PR #2228, opened 2021-12-29
chore: remove create directory from contributing guide

SethFalco:
### Description

Removes that the `credentials` directory must be created, as this is no longer the case.
Just noticed that I forgot to remove that. ^-^'


UnrealApex (approved):

Miodec (approved):
Thanks

PR #2231, opened 2021-12-30
Fix retry button duplication issue

Ferotiq:
### Description

Fixes the duplication issue that would happen to your result if you used the retry button.


amine-abdelli (approved):

Miodec (approved):
Thanks

PR #2232, opened 2021-12-31
German Quotes fix

Meimersdorfer:
Hey Mio,
 sorry to bother you with this shit again, but in my hollidays I now got the time to look through the entire code and tried to find all the mistakes still missing. As I am actually not that good with grammar ( for a native atleast) I don't know for sure whether or not I got them all now.

But I hope I got them all now, or at least there are like one or two minor ones left and I hope the code is still working.

Happy new year if you got to see it then otherwise get good into the new year.

Thanks for all the work you do with the website.

### DIFF
--- a/static/quotes/german.json
+++ b/static/quotes/german.json
@@ -1100,19 +1100,19 @@
       "id": 182
     },
     {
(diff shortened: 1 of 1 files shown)


Miodec (approved):
Thanks

PR #2233, opened 2021-12-31
New quotes for javascript

ZaninAndrea:
### Description
I added 40 new quotes for the language `code javascript`, I also fixed a couple quotes that had the length field set incorrectly. All the new quotes come from MIT licensed code.


Miodec (approved):
Thanks!

PR #2235, opened 2021-12-31
update sw.js to properly cache and request

lukew3:
### Description

The previous version of the service-worker did not cache any resources other than `index.html`. This version checks where the request is asking for data from and decides to request normally if accessing an api serving dynamic data and caches data returned from routes that are not an api serving dynamic data.


Miodec (dismissed):
Thanks!

PR #2237, opened 2022-01-01
Added google analytics to no cache list

lukew3:
Fixed google analytics requests being cached when they shouldn't be. If in the future, you want to see which responses are being saved to cache, you can go to the `Storage` tab of the Firefox developer tools and select `Cache Storage` -> `https://monkeytype.com` -> `sw-cache`
Screenshot from 2022-01-01 15-30-52

### DIFF
--- a/static/sw.js
+++ b/static/sw.js
@@ -13,7 +13,7 @@ self.addEventListener("install", (event) => {
 self.addEventListener("fetch", async (event) => {
   const host = new URL(event.request.url).host;
   if (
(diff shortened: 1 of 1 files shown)


Miodec (approved):
Thanks. What this breaking analytics? Why was this needed?

lukew3:
> Thanks. What this breaking analytics? Why was this needed?
Every event sent to google analytics was cached. Analytics was working, it was just filling up the cache with requests.

PR #2238, opened 2022-01-02
Add language code_bash

em1lyy:
### Description

This language file includes standard POSIX shell keywords, common GNU
bash extension keywords and some commonly used shell commands.


Miodec (approved):
Thanks

PR #2240, opened 2022-01-03
Correct button text animations

lukew3:
## Description
Css properties of button text were set to `transform: 0.25s`. This is invalid and `transition: 0.25s` is likely what the code was intended to say.


Miodec (approved):
Good catch.

PR #2241, opened 2022-01-03
Replaced call to nonexistent function Misc.countAllKeys

lukew3:
### Description
Load result filters calls a nonexistent function Misc.countAllKeys, causing it to error every time. I'm not completely certain what countAllKeys did but my best guess is that it counts the number of keys in a json object. I replaced the function call with code that counts the number of keys in the object, fixing the issue.


Miodec (approved):
Thanks

PR #2242, opened 2022-01-03
Clear sw-cache if localhost or new version

lukew3:
### Description
Because `sw.js` doesn't check if it's cache is equal to the server's files, cache is not updated when a new version of the site's files are released. Fixed this by clearing cache on page load if working on localhost or 127.0.0.1 and clearing cache if a new version was released on Github.


Miodec (approved):
Thanks

PR #2243, opened 2022-01-03
remove css/media-queries.css

lukew3:
### Description
`index.html` was trying to get the `css/media-queries.css` stylesheet, which does not exist. I removed the line from `index.html` that was requesting it.


Miodec (approved):
Thanks

PR #2251, opened 2022-01-04
Implemented screenshot fix

### LINKED ISSUE #2054: Screenshot is not taken wide enough

- [X] I have cleared my cache

### Current Behavior

The screenshot button takes a screenshot that is not of the correct width, which removes a part of the content on the right and looks bad.

image

### Expected Behavior

I would expect the screenshot to be of the correct width and capture all of the content.

### Steps To Reproduce

1. Complete a test
2. Attempt to create a screenshot
3. Paste the screenshot somewhere to observe it

### Environment

- OS: MacOS 11.4
- Browser: Chrome
- Browser Version: Version 95.0.4638.54 (Official Build) (x86_64)

### Anything else?

_No response_

- [ ] I have checked incognito mode

### Config

_No response_

bragefuglseth:
### Description

I increased the default screenshot padding in test-ui.js. This was done to patch the issue with the screenshot content exceeding the width of the image. I also made the html2canvas element's position get calculated by two padding variables.

Closes #2054

(This is my first pull request with code to a public repository ever, so please tell me if I need to do anything different) :)

### DIFF
--- a/src/js/test/test-ui.js
+++ b/src/js/test/test-ui.js
@@ -241,14 +241,18 @@ export async function screenshot() {
   let src = $("#middle");
   var sourceX = src.position().left; /*X position from div#target*/
   var sourceY = src.position().top; /*Y position from div#target*/
-  var sourceWidth = src.width(); /*clientWidth/offsetWidth from div#target*/
-  var sourceHeight = src.height(); /*clientHeight/offsetHeight from div#target*/
+  var sourceWidth = src.outerWidth(
+    true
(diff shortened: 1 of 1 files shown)


Miodec:
I'll be honest, just increasing the padding on the screenshot seems like a quick and dirty fix, rather than finding out why the dimensions are wrong. I've changed the function that gets the width and height to see if that helps too.

Miodec (approved):

Issue #2182, opened 2021-12-15
Add custom layout and remove some of the more obscure layouts from the website


monkeytypegeorge:
Add a popup menu to let the user define a custom layout to be used with the emulator and keymap. Also, check the database for how many users use each layout, and remove the least used ones.

Estebene:
Is the proposed solution #316?

Miodec:
> Is the proposed solution #316?

No

PR #2264, opened 2022-01-06
Add dvarf keyboard layout

O-X-E-Y:
### Description

I created a keyboard layout based on Dvorak (right hand homerow and r, a and e and some stray keys I could keep in the same positions), Whorf (left hand index is a mirror of Whorf's) and Dwarf (vowel columns) which I think is pretty good. It'd be great if it could be on here :)

### DIFF
--- a/src/js/layouts.js
+++ b/src/js/layouts.js
@@ -899,7 +899,7 @@ const layouts = {
             " "
         ]
     },
-    octa8: {
+    dvarf: {
         keymapShowTopRow: false,
         keys: [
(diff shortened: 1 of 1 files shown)


Miodec:
Did you make this layout? Does it have any users? Because as per #2182 im going to be removing the lesser used layouts and adding a custom layout option - im not sure if it makes sense to merge this if im going to remove it soon.

O-X-E-Y:
I made it and am using it yeah, haven't been able to shill it to other people yet :p a custom layout option sounds great though, when that comes out I'll probably use that anyway. Thanks!

PR #2265, opened 2022-01-06
updated colemak dhv

oatmealsecurity:
updated bottom row of Colemak-DHv layout to accurately reflect the current layout as the previous iteration was not finalized


Miodec (approved):
Thanks

PR #2267, opened 2022-01-06
Add a Beach Theme 🏝️

teddinotteddy:
### Description

Added a beach/island aesthetic theme.

Before a test:
Screen Shot 2022-01-06 at 4 05 03 PM

In the middle of a test:
Screen Shot 2022-01-06 at 4 05 15 PM

At the end of a test: 
Screen Shot 2022-01-06 at 3 52 51 PM


Miodec:
Text color needs to be as close to white or black as possible. It can have a little bit of color in it, but mainly monochrome.

teddinotteddy:
@Miodec Do you have any suggestions to what I could change the color too? I really like the color that I chose but I will gladly change it to whatever is necessary.

teddinotteddy:
@Miodec Made those changes here is what it looks like
New Look
.

PR #2290, opened 2022-01-12
Add More Theme Documentation

teddinotteddy:
I felt as there wasn't much documentation for people who are completely new to coding but they wanted to add themes. So I created THEMES.md which explains to new coders how to add themes. I feel like some more stuff could be added so let me know if there is anything I should change or add. I also offered a little bit of information for forking repos and whatnot.


Miodec:
I will say it reads quite weird. Its aimed at the more inexperienced users of GitHub, but the content of it describes forking, then making a pull request before making any changes. I think it should be reordered.

Also it would be a nice addition to add screenshots that highlight which color changes which elements.

teddinotteddy:
Alright I made those changes.

PR #2291, opened 2022-01-12
Adds rsthd layout

AnthonyW90:
### Description

Adds the RSTHD keyboard layout as described here: https://xsznix.wordpress.com/2016/05/16/introducing-the-rsthd-layout/


Miodec:
Its an ergodox layout. Monkeytype doesn't support multiple thumb keys. Sorry.

PR #2292, opened 2022-01-12
Add Ducky Frozen Llama Theme

teddinotteddy:
I added a theme meant to resemble the Ducky One 2 mini Frozen Llama keyboard.
20190822174945b9ll89bt71

Here is what the theme looks like:
Screen Shot 2022-01-12 at 4 27 20 PM
Screen Shot 2022-01-12 at 4 27 04 PM


Miodec (approved):
Thanks

PR #2299, opened 2022-01-13
Fix APT layouts

Apsu:
Remove outdated and little-used APTx
Update APT to current version in author's repo
Add row-stagger angle-mod variant


Miodec (approved):
thanks

PR #2312, opened 2022-01-15
Offline mode (with localStorage snapshot)

lukew3:
### Description

Set monkeytype into offline mode if getDataAndInit fails. Tests that cannot be uploaded are saved to localStorage and updated once monkeytype is started without offline mode. This commit differs from a pr I opened earlier today because now dbSnapshot is saved in localStorage, making it accessible offline.


Miodec:
I feel like Misc mainly holds helper functions. I think `OfflineMode` should be its own file.

lukew3:
> I feel like Misc mainly holds helper functions. I think OfflineMode should be its own file.

It can be. It's just a pretty insignificant amount of code. Actually all the code in there is doing is just setting and getting a single variable. What do you think about storing and accessing the offlineMode variable from sessionStorage?

src/js/account.js
```diff
@@ -50,13 +55,27 @@ export async function getDataAndInit() {
...
+    );
+    sessionStorage.setItem("offlineMode", true);
+    if (DB.getSnapshot()) {
+      $("#menu .icon-button.account .text").text(DB.getSnapshot().name);
+    }
+    //$("#top #menu .account").css("opacity", 1);
+    if (UI.getActivePage() == "pageLoading") UI.changePage("");
+    return;
 
+    /* The following is old code that can probably be deleted, unless connection there are other errors other than no connection where users shoudl sign out
```
lukew3, line 71: What should happen with this code? I don't think that there's a reason to sign the user out when they have no connection.

Miodec: How are you going to authenticate the user if you are offline though

lukew3: If you are offline, you use the saved user and save results to localstorage. The next time they are authenticated, the results are added

Miodec: Are you testing this? Because thats what i get 
image
image

I dont think there is a way to login without internet...

lukew3: Is that when you try to login?

Miodec: Yes, wifi disabled

lukew3: I didn't really consider logging in because most people who are using it offline have used it before when they did have internet access. There probably should be an error handler that tells the user that they can't go to the sign up/login page because they are offline

lukew3: Alright, user can't attempt to login in offline mode now

CameronCT:
The only concern with this would likely be the security of it and of course the modification of local storage to get an illegitimate result. Are there any plans to encrypt or modify the data to at least help that?

Miodec:
Yeah, after some thought - keeping the results locally could be a problem.

PR #2322, opened 2022-01-19
Extend code_python.json (dunder methods and libraries)

ASMaksimov2007:
### Description

In this PR I added some new words to Python code file, including dunder magic methods and some useful libraries that are often imported.


static/languages/code_python.json
```diff
@@ -3,6 +3,24 @@
...
+    "__floordiv__",
+    "__truediv__",
+    "__mod__",
+    "__pow__",
+    "__lt__",
+    "__le__",
+    "__eq__",
+    "__ge__",
+    "__ne__",
+    "__gt__",
```
Miodec, line 23: Are those needed? Are they being typed that often in Python? I feel like these dont' translate well into a typing test, just because how awkward it is to type them, but I dont do much Python so Im not sure.

ASMaksimov2007: I think some of them are very nesessary to include, for example, __init__ and __repr__. Others are not so popular, so I will remove them if you want.

Miodec (approved):
Thanks

PR #2323, opened 2022-01-19
added tashkeel stripped virsion for arabic files

mosaleh52:
The Arabic files available are with tashkeel which is kind of symbols that are used to recognize words but not used in every day so I made a script that is removing tashkeel for those whom want to practice without it and add the stripped files with change in list and groups files
``` python

from pyarabic.araby import strip_tashkeel
import json
file = 'arabic.json' #file name with assumtion that you are runnig code in same folder
target_file  = file.replace('.json','tashkeel_stipped.json')
print(target_file)
with open(file,'r+') as f:
    data = json.load(f)
    data['name'] = target_file.replace('.json','')
    for idx,word in enumerate(data['words']):
        data['words'][idx] = strip_tashkeel( data['words'][idx])
with open(target_file,'w') as tf:
    json.dump(data,tf,indent=4,ensure_ascii = False)
    print('done')
```


Miodec:
Have you checked if `lazy mode` doesn't achieve the same effect?

mosaleh52:
> Have you checked if `lazy mode` doesn't achieve the same effect?

That is true, I have not noticed that ```lazy mode``` is doing this thanks for the note

PR #2339, opened 2022-01-23
Update britishenglish.json

Marken-Foo:
Correct British spellings of "armour" and "labelling"

### Description

Closes #


Miodec (approved):
Thanks

PR #2340, opened 2022-01-23
Fix bug that prevented sw cache from being deleted

lukew3:
### Description
Multiple things are occuring in this pr, all related to the service worker and caching system
*  Added `event.waitUntil(self.skipWaiting());` and `event.waitUntil(self.clients.claim());`
   * These functions stop old service workers and allow the new service worker to take control
     * https://stackoverflow.com/questions/41009167/what-is-the-use-of-self-clients-claim 
*  Placed delete caches inside of "activate" event.
   *  Caches should be removed and clients.claim() should be run when the sw is activated (first loaded)
* Open cache by name before checking and setting cache for better precision
* Removed "monkeytype.com" and "localhost:5000" from nocache list.
  * Now that sw.js is updated after each build and new files are requested, we don't have to request new ones if there was a bad file in an older version
* Stopped updating /static/sw.js on compile.
  * When gulp compile was run, the sw.js file in the static folder was updated. When watching files for updates in dev mode, updating static/sw.js causes the program to rebuild infinitely.
  * You could create a simple gulp task to copy sw from dist to static and add to build series if you think it's important to commit, but I don't think it's necessary

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -281,8 +281,7 @@ task("updateSwCacheName", function () {
         `const staticCacheName = "sw-cache-${dateString}";`
       )
     )
-    .pipe(dest("./dist/"))
-    .pipe(dest("static/"));
+    .pipe(dest("./dist/"));
 });
 
 task(
--- a/static/sw.js
+++ b/static/sw.js
@@ -34,19 +34,20 @@ self.addEventListener("fetch", async (event) => {
     event.respondWith(fetch(event.request));
   } else {
     // Otherwise, assume host is serving a static file, check cache and add response to cache if not found
-    let cache = await caches.open(staticCacheName);
     event.respondWith(
-      cache.match(event.request).then(async (response) => {
-        // Check if request in cache
-        if (response) {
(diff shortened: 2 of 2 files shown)


Miodec:
Im getting a lot of
```js
sw.js:38 Uncaught (in promise) DOMException: Failed to execute 'respondWith' on 'FetchEvent': The event handler is already finished.
    at http://localhost:5000/sw.js:38:11
```
when refreshing the page. (but its not showing up at all when refreshing with clearing cache `ctrl + shift + r`)

Another thing I found is that the cache (at least on localhost) holds about 20 files. After recompiling and refreshing the page, the old cache is cleared correctly, but the new one only holds one item - is that okay? (It fills itself back up after another refresh)

Old cache before refreshing
image
New cache after refreshing
image

lukew3:
You can't fix the second issue you mentioned without loading the service worker alone and then loading index.html and it's required files (https://stackoverflow.com/a/61432538/11073182).

Miodec:
Works great, thanks.

Miodec (approved):

PR #2351, opened 2022-01-25
Add Grad Theme

akhilesh-balaji:
Added the “grad” theme:

image


Miodec:
I would say its too similar to themes `vscode`, `drowning`, `midnight` and `arch`. Sorry.

PR #2357, opened 2022-01-26
Refactor cron jobs

Bruception:
### Description

- Separated cron job definitions from server.
- Restructured how cron jobs are configured.
- Rewrote/simplified the leaderboard update job logic.


backend/jobs/updateLeaderboards.js
```diff
@@ -0,0 +1,49 @@
...
+
+  await LeaderboardsDAO.update("time", leaderboardTime, "english");
+
+  const top10AfterUpdate = await mongoDB()
+    .collection(`leaderboards.english.time.${leaderboardTime}`)
+    .find()
+    .limit(10)
+    .toArray();
+
+  const newRecordHolder = top10AfterUpdate.find((user, index) => {
```
Bruception, line 24: The old logic found the first mismatched `uid` and then checked if it was a recent record.
Exiting the search even if the first mismatch was not recent.

The new logic, will continue to try and find a new record even if it is not recent.
Not sure what the consensus on this new logic is. But I do think it makes a bit more sense.

If not, I'm okay with reverting back to the old logic.

backend/jobs/updateLeaderboards.js
```diff
@@ -0,0 +1,51 @@
...
+
+  await LeaderboardsDAO.update("time", leaderboardTime, "english");
+
+  const top10AfterUpdate = await mongoDB()
+    .collection(`leaderboards.english.time.${leaderboardTime}`)
+    .find()
+    .limit(10)
+    .toArray();
+
+  const newRecord = top10AfterUpdate.find((record, index) => {
```
Bruception, line 25: The old logic found the first mismatched uid and then checked if it was a recent record.
Exiting the search even if the first mismatch was not recent.

The new logic, will continue to try and find a new record even if a mismatch was not recent.
Not sure what the consensus on this new logic is. But I do think it makes a bit more sense.

If not, I'm okay with reverting back to the old logic.

Miodec: I guess this only fixes a rare occurrence where I remove a cheated score from the leaderboard, the whole thing shifts up and in the same update, someone got a top 10 score. Very rare but yeah, it's fine to keep this

Bruception: I’m trying to test the discord integration. Is there any additional setup required outside of what is specified in the documentation?

Miodec: I don't think so

Bruception: Is the current discord bot polling the production database for commands? If so, how can I test the discord bot locally? If not, could you point me to a resource where I can read more about the bot’s implementation?

Miodec: Yeah the bot is polling the database for commands. The bot doesn't have any docs im afraid, but the code is here https://github.com/Miodec/monkeytype-bot

backend/jobs/updateLeaderboards.js
```diff
@@ -0,0 +1,53 @@
...
+    top10BeforeUpdate.map((record) => {
+      return [record.uid, record];
+    })
+  );
+
+  await LeaderboardsDAO.update("time", leaderboardTime, "english");
+
+  const top10AfterUpdate = await getTop10(leaderboardTime);
+
+  const newRecords = top10AfterUpdate.filter((record) => {
```
Bruception, line 27: After thinking about this for a bit, I think it makes sense to announce all new records.

To detect a new record we only need to consider two possibilities:
- A user already in the top 10 improves their score
OR
- A new user enters the top 10.

We also check if the record is recent, in the case that a record is deleted due to cheating.

Once we have a list of new records, we perform a single insertion that includes all the new records instead of inserting each bot announcement individually.

Miodec: Well, as mentioned before - cases where there have been two changes in the top 10 in the same update are very very small, since updates happen once every 5 minutes and the top 10 is quite hard to get into. Still, could do that yeah.

> A user already in the top 10 improves their score

I agree, but only if that changes their position. I don't think its necessary to annouce that someone improved their record by 0.1 wpm but that didnt change their position in the top 10.

> We also check if the record is recent, in the case that a record is deleted due to cheating.

Thats not what that check is for. That check is there to not annouce old results in case a cheated score was removed.

> Once we have a list of new records, we perform a single insertion that includes all the new records instead of inserting each bot announcement individually.

I think it still needs separate command insertions since the bot will only handle one user at a time.

Bruception: > I agree, but only if that changes their position. I don't think its necessary to annouce that someone improved their record by 0.1 wpm but that didnt change their position in the top 10.<

Yes that was a typo, my code explicitly checks for a change in rank not WPM score.

> Thats not what that check is for. That check is there to not annouce old results in case a cheated score was removed. (because that would cause all positions to be shifted up one place)

Yes, my code will correctly handle this case. 
 
> I think it still needs separate command insertions since the bot will only handle one user at a time.

I was reading [here](https://github.com/Miodec/monkeytype-bot/blob/bc2c66dd595d3cc08583089bceff2458dbde7f0f/index.js#L339), and it seems like the bot pulls up to 10 commands every 30 seconds from the DB and pushes then into a queue for execution.
My code will write N bot commands to the DB for N users if they meet the criteria for a leaderboard record.

Miodec: Great, so is this PR ready?

Bruception:
Screen Shot 2022-01-28 at 12 43 36 PM
Here is my local testing with the discord bot.

backend/jobs/updateLeaderboards.js
Bruception: Yes, I added a screenshot that shows my tests.

Miodec (approved):
Nice work. Thanks.

PR #2367, opened 2022-01-29
Add initial report quote endpoint

Bruception:
### Description

Added initial report quote endpoint. Still trying to spec some of the implementation details.
We can use this PR (and any subsequent ones) as a way to discuss this feature further.


backend/api/routes/quotes.js
```diff
@@ -48,4 +50,24 @@ quotesRouter.post(
   QuoteRatingsController.submitRating
 );
 
+quotesRouter.post(
+  "/report",
+  RateLimit.quoteReportSubmit,
+  authenticateRequest,
+  requestValidation({
+    body: {
+      quoteId: joi.string().required(),
```
Bruception, line 59: Do quotes have a unique identifier? From what I can tell, quotes are identified by their id relative to their language.
Meaning we would need to keep a tuple to identify which quote a report is referencing.

backend/api/routes/quotes.js
```diff
@@ -48,4 +50,24 @@ quotesRouter.post(
...
+      quoteLanguage: joi.string().required(),
+      reason: joi
+        .string()
+        .valid("Grammatical error", "Inappropriate language")
```
Bruception, line 63: Open to anymore reasons we want to include.

backend/api/routes/quotes.js
```diff
@@ -48,4 +50,24 @@ quotesRouter.post(
...
+        .required(),
+      comment: joi.string().allow("").max(140).required(),
```
Bruception, line 65: Arbitrary maximum limit.

backend/middlewares/apiUtils.js
```diff
@@ -0,0 +1,32 @@
+const joi = require("joi");
+const MonkeyError = require("../handlers/error");
+
+function requestValidation(validationSchema) {
```
Bruception, line 4: Adding this handler so that we can standardize how future endpoints handle request validation.

backend/middlewares/rate-limit.js
```diff
@@ -68,6 +68,14 @@ exports.quoteRatingsSubmit = rateLimit({
   keyGenerator: getAddress,
 });
 
+// Quote reporting
+exports.quoteReportSubmit = rateLimit({
+  windowMs: 60 * 60 * 1000,
```
Bruception, line 73: This can be tuned. Open to suggestions.

package.json
```diff
@@ -62,6 +62,7 @@
     "gulp-replace": "^1.1.3",
     "helmet": "^4.6.0",
     "howler": "^2.2.1",
+    "joi": "^17.6.0",
```
Bruception, line 65: Joi is a great tool for schema validation.

backend/api/routes/quotes.js
Miodec: Yes. Could also build a string ‘language_id‘

backend/api/routes/quotes.js
Miodec: Im thinking something like "low quote quality" or something like that but I'm not sure how to word it. Basically for cases where the text is not really "quotable".

backend/api/routes/quotes.js
Miodec: Increase to 250

backend/middlewares/apiUtils.js
Miodec: Sounds good

backend/middlewares/rate-limit.js
Miodec: Lower to something just out of reach for a normal user. Probably lower the max to 250, but that also seems like a lot. Maybe lower the window too.. I'll think about it more

Miodec:
Also, could you mark this as a draft? Ill merge it once its all complete.

Bruception:
> Also, could you mark this as a draft? Ill merge it once its all complete.

I was thinking of putting out this feature in a series of PRs. That way each PR is more focused. The PRs would start with backend logic, and the final PR would be the frontend UI/logic.

Miodec:
> > Also, could you mark this as a draft? Ill merge it once its all complete.
> 
> I was thinking of putting out this feature in a series of PRs. That way each PR is more focused. The PRs would start with backend logic, and the final PR would be the frontend UI/logic.

Alright, sure. But i think marking as draft would still make sense until we resolve conversations. Easier for me to see whats going on at a glance i think.

backend/middlewares/rate-limit.js
Bruception: We can tune it as we build out the feature, and can test the full flow.

backend/api/routes/quotes.js
Bruception: Is your yes referring to the fact that quotes have an individual id, or that we need to keep a tuple to identity them?

backend/middlewares/rate-limit.js
```diff
@@ -70,8 +70,8 @@ exports.quoteRatingsSubmit = rateLimit({
 
 // Quote reporting
 exports.quoteReportSubmit = rateLimit({
-  windowMs: 60 * 60 * 1000,
-  max: 500 * multiplier,
+  windowMs: 30 * 60 * 1000, // 30 min
+  max: 250 * multiplier,
```
Miodec, line 74: After some thought, I would even go down to 50 reports every 30 minutes. I don't think users will report every quote and with that change, it still allows a report every 36 seconds. Could even experiment with making it even lower and seeing when users hit the limit.

backend/api/routes/quotes.js
Miodec: Quotes don't have a completely unique id. They will have to be identified by their language too. Thats why i mentioned that we could also combine them into a string `language_quoteid` and then split on `_`

Miodec (approved):
Looking good so far.

Miodec (commented):
Do you think its really necessary to validate the language like that? Im kinda thinking just checking that its a string is fine. And the front end will send the current language anyway, the user doesnt even get to change whats being sent.

Bruception:
> Do you think its really necessary to validate the language like that? Im kinda thinking just checking that its a string is fine. And the front end will only send the current language anyway, the user doesnt even get to change whats being sent.

I was thinking in the case that someone may maliciously craft a POST request. They might change the language to something that is not supported. I think being strict could potentially save us from any unforeseen issues. I'm fine with removing it tho.

Miodec:
> > Do you think its really necessary to validate the language like that? Im kinda thinking just checking that its a string is fine. And the front end will only send the current language anyway, the user doesnt even get to change whats being sent.
> 
> I was thinking in the case that someone maliciously crafts a POST request. They might change the language to something that is not supported. I think being strict could potentially save us from any unforeseen issues. I'm fine with removing it tho.

Alright, leave it. Then I guess it would be good to add this validation to the already existing endpoints at some point ;D

Bruception:
> > > Do you think its really necessary to validate the language like that? Im kinda thinking just checking that its a string is fine. And the front end will only send the current language anyway, the user doesnt even get to change whats being sent.
> > 
> > 
> > I was thinking in the case that someone maliciously crafts a POST request. They might change the language to something that is not supported. I think being strict could potentially save us from any unforeseen issues. I'm fine with removing it tho.
> 
> Alright, leave it. Then I guess it would be good to add this validation to the already existing endpoints at some point ;D

Thats the plan! lol

Miodec (approved):

Bruception:
Can we merge this into the master branch?

PR #2369, opened 2022-01-29
Caps lock

tricarbonate:
### Description

When pressing the Caps Lock key, the warning gets immediately shown, but when pressing again, the warning only gets hidden after another key press. This PR aims at hiding the caps lock warning immediately after caps lock is deactivated.


Miodec (changes requested):
This breaks the indicator on a Mac:
 - only shows up after another key is pressed after caps lock
 - goes away after caps lock is pressed again

PR #2392, opened 2022-02-01
Added DMG theme

Usman-Abubakr:
### Description

Added DMG theme. When typing is active, the menu buttons disappear to leave only one, the power indicator.

1 - home

Default color settings:
2 - default

Colorful mode (green errors referencing the screen):
3 - colorful


Miodec (approved):
Thanks

PR #2395, opened 2022-02-02
Update Configuration to handle arrays/ Add new db client

Bruception:
### Description

- Added additional logic to correctly handle arrays in the config.
- Added new Db client

### DIFF
--- a/backend/dao/configuration.js
+++ b/backend/dao/configuration.js
@@ -1,5 +1,5 @@
 const _ = require("lodash");
-const { mongoDB } = require("../init/mongodb");
+const db = require("../init/db");
 const BASE_CONFIGURATION = require("../constants/base-configuration");
 const Logger = require("../handlers/logger.js");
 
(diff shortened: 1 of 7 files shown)


backend/dao/configuration.js
```diff
@@ -84,10 +90,25 @@ class ConfigurationDAO {
...
     return this.configuration;
   }
+
+  static async pushConfiguration(configuration) {
+    if (this.databaseConfigurationUpdated) {
+      return;
+    }
+
+    try {
+      await mongoDB().collection("configuration").replaceOne({}, configuration);
```
Bruception, line 102: Changing this to replace since if there are properties in the live configuration that are not longer in the base, they will not get removed.

backend/init/mongodb.js
```diff
@@ -1,42 +1,7 @@
...
-      })
-      .catch((e) => {
-        console.error(e.message);
-        console.error("FAILED TO CONNECT TO DATABASE. EXITING...");
-        process.exit(1);
-      });
-  },
   mongoDB() {
-    return mongoClient.db(process.env.DB_NAME);
+    return db;
```
Bruception, line 5: Keeping this for backwards compatibility. We can slowly refactor code that uses the old client.

Miodec:
Whats the advantage of the new db client?

Bruception:
> Whats the advantage of the new db client?

It’s purely for refactor purposes. We could add some connection retry logic later on tho. I’m kind of leaning towards using a higher level abstraction from the native mongo driver tho. Using something like mongoose would better help with having clear schemas for our data model.

Miodec (approved):
Thanks

PR #2404, opened 2022-02-03
Added grabbing to the range input

clickers331:
Added the cursor styles for the hover and active states for the range inputs thumb like so:

```
&:hover{
   cursor:grab;
}
&:active{
   cursor:grabbing;
}
```


Miodec:
Thanks

Issue #2375, opened 2022-01-30
Add post request validation to all api endpoints


monkeytypegeorge:
- [x] Config endpoint
- [x] Leaderboard endpoint
- [x] Preset endpoint
- [x] Quotes endpoint
- [x] Result endpoint
- [ ] User endpoint

Bruception:
I can work on this.

PR #2417, opened 2022-02-04
Add results schema validation

Bruception:
### Description

- Added endpoint request validation to the result endpoints.

Related to: #2375


Miodec (approved):
Thanks

PR #2421, opened 2022-02-06
Fixed menu children properties

batsoup804:
##  PR solves:

This aims to fix the backgrounds of menu children. They should be differing in background color however because the css is incorrect it isn't working completely correctly.


Bruception:
Hi could you attach before and after fix screenshots?

Miodec:
The pr looks like this:
image

I tried fixing it by changing the text color:
image


But i dont like either i think. Ill just stick with the version that is live:
image

PR #2423, opened 2022-02-06
Add request validation to user endpoints

### LINKED ISSUE #2375: Add post request validation to all api endpoints

Bruception:
### Description

- Added request validation to user endpoints
- Re-wrote some of the endpoints to follow REST standards.

Closes #2375


src/js/account-controller.js
```diff
@@ -484,7 +484,7 @@ async function signUp() {
     //make sure to do clean up here
     if (createdAuthUser) {
       await createdAuthUser.user.delete();
-      axiosInstance.post("/user/delete", { uid: createdAuthUser.user.uid });
```
Bruception, line 487: Why were we sending `uid` as part of the response? We are reading directly from the decoded token. In prod, the decoded token wont be set based on the response body.

Miodec: Could be a leftover of something, could also be the fact that at the moment of this request being sent, the user doesnt have a token.

Would need to place an exception in there and test.

Bruception: > Could be a leftover of something, could also be the fact that at the moment of this request being sent, the user doesnt have a token/is not signed in.
> 
> Would need to place an exception in there and test.

Even if they are not signed in, they won't be able to pass authentication since the environment is not dev.

Miodec (approved):
Thanks

PR #2424, opened 2022-02-06
added snes theme

Usman-Abubakr:
### Description

Added a SNES theme. Gradient menu buttons, switching to grey when typing.

Home:
1 - home

Default settings:
2 - default

Colorful mode:
3 - colorful


Miodec (approved):
Thanks

PR #2440, opened 2022-02-07
chore: updated node/npm versions

CameronCT:
Updated node and npm versions to be `16.13.2` and `8.1.2` as per Miodec's request.


Ferotiq:
epic

Issue #2430, opened 2022-02-07
Convert backend to TS


monkeytypegeorge:

mrbrianevans:
I would like to work on this

Miodec:
> I would like to work on this

Awesome. Please start with a pr for compilation setup.

Bruception:
> Okay #2452 opened. Working on it

Can we do this after the backend overhaul is complete?

mrbrianevans:
@Bruception #2452 should really be merged as soon as possible. The longer its left seperate, the more out-of-date it gets. I converted all the `require`s to `imports` etc and successfully got it to compile to typescript and run. If it is merged, and then the backend is overhauled, at least the changes to backend will be using `import` and `export`, rather than just creating more files that need to be migrated. 

Are there any issues that anyone is aware of that are preventing the "Compile backend with Typescript" PR from being merged?

The way its implemented, all the files can stay `.js` for the moment, and still be compiled by typescript, as long as their using ES module import syntax. So type information doesn't need to be immediately added to the whole backend. The PR just adds compilation, but doesn't require files to be converted to typescript.

mrbrianevans:
Converting the backend to TS is now apart of the backend API overhaul. Compilation is set up on the `api-overhaul` branch (#2515). 

Converting any backend files to typescript should be done on the `api-overhaul` branch and NOT `typescript`. The `typescript` branch is for the frontend, and appears to be unmaintained.

(please could someone with permissions update the checklist in the issue description to tick off set up compilation)

Miodec:
> (please could someone with permissions update the checklist in the issue description to tick off set up compilation)

If i assign you to this issue do you then have permission to update the checklist?

mrbrianevans:
nope, still can't edit

Miodec:
Alright, ill mark it as completed then.

Bruception:
> Converting the backend to TS is now apart of the backend API overhaul. Compilation is set up on the `api-overhaul` branch (#2515).
> 
> Converting any backend files to typescript should be done on the `api-overhaul` branch and NOT `typescript`. The `typescript` branch is for the frontend, and appears to be unmaintained.
> 
> (please could someone with permissions update the checklist in the issue description to tick off set up compilation)

I was thinking to continue with the backend TS conversion once API overhaul is merged.

mrbrianevans:
[...]

PR #2452, opened 2022-02-09
Compile backend with Typescript

mrbrianevans:
## Part of #2430 (setting up compilation)
- installed typescript at the top level package.json (because both frontend and backend will be using typescript)
- added tsconfig in backend
- added build script to backend package.json
- fixed minor typescript compilation errors (eg calling `.toString()`)
- migrated to ES modules (changed to import export syntax)
- added typing for anticheat


mrbrianevans:
It cannot yet build, many compilation errors with JS code. 

I will work through them and resolve where possible. 

Some look like actual mistakes (too many arguments passed to function etc), some just need type assertions or conversions. 

Can someone explain to me how anticheat works? I'm not familiar with it and it doesn't seem to be in my git repo, but its imported somewhere and typescript can't find it. Is it in another repo or something? 
```ts
function validateResult(result): boolean
function validateKeys(result, uid): boolean
```

I can write a type declaration file for it if I know the inputs and outputs. Above is what I can tell from its usage.

Miodec:
Yes, anticheat is a private repo. Can you still implement without having the code? Do you need some info about it to be able to?

Ferotiq:
It might make sense just to use ts-node instead of building the backend.

mrbrianevans:
@Miodec 
> Yes, anticheat is a private repo. Can you still implement without having the code? Do you need some info about it to be able to?

I should be able to write a typescript declaration for it without having access to the source code. If you could just confirm the function signatures (ie arg and return types)?

Miodec:
> @Miodec
> 
> > Yes, anticheat is a private repo. Can you still implement without having the code? Do you need some info about it to be able to?
> 
> I should be able to write a typescript declaration for it without having access to the source code. If you could just confirm the function signatures (ie arg and return types)?

Both return a string. Result is an object object (you can see its structure in result-schema.js) and uid is a string.

mrbrianevans:
@Ferotiq what advantage would there be to running it using ts-node over first building and then running with node? As far as I know, ts-node adds runtime memory overhead, and doesn't seem to be superior other than being quick and convenient for development. Interested to hear any differing opinions though

package.json
```diff
@@ -33,7 +33,8 @@
     "eslint-plugin-require-path-exists": "1.1.9",
     "husky": "4.3.0",
     "prettier": "2.1.2",
-    "pretty-quick": "3.1.0"
+    "pretty-quick": "3.1.0",
+    "typescript": "^4.5.5"
```
Bruception, line 37: Use save exact

mrbrianevans: Okay. Not sure where's best to have typescript installed. In the root `package.json` since it is required by frontend and backend, or separately in each one?

Miodec: It would be nice to have it in one place but im not sure how the ts compilation will handle not having the typescript version defined (or it not being installed in those folders)

Bruception: I think it’s okay to install it in the root. The typescript cli can handle multiple ts configs in different folders.

backend/package.json
```diff
@@ -3,9 +3,12 @@
   "version": "1.0.0",
   "license": "GPL-3.0",
   "private": true,
+  "type": "module",
   "scripts": {
-    "start:dev": "nodemon --watch ./ ./server.js",
-    "start:dev-ts": "nodemon --watch ./ ./server.js"
+    "build": "tsc --build",
+    "start": "node --experimental-specifier-resolution=node ./build/server.js",
```
Bruception, line 9: Can we name these scripts such that they map to the ones that they are replacing?
Also if we are not longer using Nodemon, can we remove it as a dependency?

backend/package.json
```diff
@@ -30,5 +33,10 @@
     "string-similarity": "4.0.4",
     "ua-parser-js": "0.7.28",
     "uuid": "8.3.2"
+  },
+  "devDependencies": {
+    "@types/node": "^17.0.16",
```
Bruception, line 38: Save exact

backend/jobs/delete-old-logs.js
```diff
@@ -1,6 +1,8 @@
-const { CronJob } = require("cron");
-const { mongoDB } = require("../init/mongodb");
-const Logger = require("../handlers/logger");
+import { CronJob } from "cron";
+
```
Bruception, line 2: Nit: remove the spaces between the imports.

backend/middlewares/api-utils.js
```diff
@@ -1,6 +1,8 @@
-const _ = require("lodash");
-const joi = require("joi");
-const MonkeyError = require("../handlers/error");
+import _ from "lodash";
+
```
Bruception, line 2: Nit: too many new lines between imports

backend/jobs/update-leaderboards.js
```diff
@@ -1,6 +1,7 @@
-const { CronJob } = require("cron");
-const BotDAO = require("../dao/bot");
-const LeaderboardsDAO = require("../dao/leaderboards");
+import { CronJob } from "cron";
+
```
Bruception, line 2: Nit: too many new lines between imports

backend/api/controllers/new-quotes.js
```diff
@@ -1,48 +1,49 @@
-const NewQuotesDAO = require("../../dao/new-quotes");
-const MonkeyError = require("../../handlers/error");
-const UserDAO = require("../../dao/user");
-const Logger = require("../../handlers/logger.js");
-const Captcha = require("../../handlers/captcha");
+import NewQuotesDao from "../../dao/new-quotes";
+
```
Bruception, line 2: Nit: new line

Bruception:
Nit: Remove the new lines between imports.

backend/api/controllers/new-quotes.js
```diff
@@ -1,48 +1,48 @@
-const NewQuotesDAO = require("../../dao/new-quotes");
-const MonkeyError = require("../../handlers/error");
-const UserDAO = require("../../dao/user");
-const Logger = require("../../handlers/logger.js");
-const Captcha = require("../../handlers/captcha");
+import NewQuotesDao from "../../dao/new-quotes";
+const { get, add, approve: _approve, refuse: _refuse } = NewQuotesDao;
```
Bruception, line 2: I don't think we should do this destructing.  Just import the NewQuotesDao.

backend/dao/result.js
```diff
@@ -1,7 +1,9 @@
-const { ObjectID } = require("mongodb");
-const MonkeyError = require("../handlers/error");
-const { mongoDB } = require("../init/mongodb");
-const UserDAO = require("./user");
+import Mongo from "mongodb";
+const { ObjectID } = Mongo;
+import MonkeyError from "../handlers/error";
+import db from "../init/db";
+
```
Bruception, line 5: Missed one

backend/handlers/logger.js
```diff
@@ -1,17 +1,15 @@
-const db = require("../init/db");
+import db from "../init/db.js";
```
Bruception, line 1: Remove .js extension

backend/init/db.js
```diff
@@ -1,4 +1,4 @@
-const { MongoClient } = require("mongodb");
+import MongoClient from "mongodb/lib/mongo_client.js";
```
Bruception, line 1: remove .js

backend/types/express.d.ts
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line
+declare namespace Express {
+  export interface Request {
```
Bruception, line 3: I think instead of creating an express types declaration we can do something like:

`declare namespace MonkeyTypes { interface Request extends Express.Request } ...`
Also I don't think we have to use the export keyword here.

Bruception:
Closing this because this is addressed on another PR.

PR #2456, opened 2022-02-09
Add Initial API Client

Bruception:
### Description

- Defined the frontend API client, currently only GET requests are being routed through it.
- Keeping axios as the HTTP client library for now. Will make this change later.
- The client will get updated whenever the new API is rolled out.


Bruception:
Some of the TypeScript definitions need to be ironed out before this can be merged.

PR #2457, opened 2022-02-10
Banana Theme 🍌

averyl2020:
add banana theme.
Screen Shot 2022-02-09 at 6 07 49 PM
Screen Shot 2022-02-09 at 6 08 01 PM


teddinotteddy (approved):

Miodec:
I think the yellow is too saturated, and the scheme in general is too close to honey. Sorry

PR #2519, opened 2022-02-16
Added support for the Lua and Luau languages.

howmanysmall:
### Description

I added support for the [Lua](https://www.lua.org/) and [Luau](https://luau-lang.org/) languages.


frontend/static/languages/code_lua.json
```diff
@@ -0,0 +1,58 @@
...
+    "nil",
+    "not",
+    "or",
+    "repeat",
+    "return",
+    "then",
+    "true",
+    "until",
+    "while",
+
```
Bruception, line 27: Why these new lines?

frontend/static/languages/code_luau.json
```diff
@@ -0,0 +1,58 @@
...
+    "local",
+    "nil",
+    "not",
+    "or",
+    "repeat",
+    "return",
+    "then",
+    "true",
+    "until",
+    "while",
```
Bruception, line 26: Same as above

frontend/static/languages/code_lua.json
howmanysmall: Oh, that's just me separating them based on what they'd be categorized as. First group is keywords you cannot use as a variable name, second one stuff that you can.

Bruception: > Oh, that's just me separating them based on what they'd be categorized as. First group is keywords you cannot use as a variable name, second one stuff that you can.

Ah, I think the prettier job will fail if you include those. It’s okay to remove them.

Miodec:
Ill just merge this, prettier will remove the new lines.

PR #2522, opened 2022-02-17
Ape migration p2

Bruception:
### Description

- Migrated the leaderboards endpoints to use ape.
- Initiated some backend types.


Miodec (approved):

frontend/src/scripts/elements/leaderboards.ts
```diff
@@ -280,18 +280,19 @@ async function update(): Promise<void> {
...
-    return response.data;
-  });
+  const responses = await Promise.all(leaderboardRequests);
+
+  const failedResponse = responses.find((response) => response.status !== 200);
+  if (failedResponse) {
+    Notifications.add(
+      "Failed to load leaderboards: " + failedResponse.message,
+      -1
+    );
```
Miodec, line 290: return here so the rest of the leaderboard loading doesn't cotinue.

Miodec (changes requested):

frontend/src/scripts/elements/leaderboards.ts
Bruception: Missed this, thanks.

PR #2526, opened 2022-02-18
A new minimalistic theme

Fluffy-Cookie:
### The screenshots for extra awesomeness!

Screenshot 2022-02-18 151941

Screenshot 2022-02-18 152415

### The charcoal and pure white go perfectly with each other

Screenshot 2022-02-18 152814

### DIFF
--- a/_list.json
+++ b/_list.json
@@ -1,0 +1,741 @@
+[
+  {
+    "name": "dark",
+    "bgColor": "#111",
+    "mainColor": "#eee"
+  }
+  ,{
+    "name": "muted",
+    "bgColor": "#525252",
+    "mainColor": "#B1E4E3"
+  }
+  ,{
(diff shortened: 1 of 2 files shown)


Miodec:
Sorry, the color scheme is pretty much the same as the `paper` theme.

Fluffy-Cookie:
Hmm, I did notice that thought that punch was different, anyway I am going to see this as just a prototype maybe I will do one more changes and BTW I would like to know how we create custom colors for the menu buttons
Screenshot 2022-02-19 115809

It was hard NVM we can make a better one!

Miodec:
Here is the theme you are looking at: 

https://github.com/Miodec/monkeytype/blob/master/frontend/static/themes/aether.css

Fluffy-Cookie:
Thanks, I'll look into it!

PR #2535, opened 2022-02-19
feat(cmd): Add command to choose a quote

kiriDevs:
### Description

This PR aims to add a command to the commandline that allows the user to pick a specific quote. When a quote is selected, the mode will be switched to 'custom' and the custom text will be set to the text of the respective quote.

The currently selected language will be considered in the options the user is given for selecting a quote in the submenu.

### Known Problems

- The options are only adapted to the language the first time the site loads
  (-> doesn't reflect language changes without reloading)
  **Requires generating the options before the submenu is opened - is that even currently possible in the command line? Also performance?**
- The site seems to become "laggy" the 2nd time a quote is selected?
- Whatever is happening here in the last element of the list
  image
  (Potentially the command line understandably just not being made for this many options?)

### More Information

When the first "known problem" is fixed, I assume that the same solution could be used to also consider the currently selected quote length when generating the quote options for the submenu.

### DIFF
--- a/frontend/src/scripts/elements/commandline-lists.ts
+++ b/frontend/src/scripts/elements/commandline-lists.ts
@@ -3183,7 +3183,7 @@ export const defaultCommands: MonkeyTypes.CommandsGroup = {
     {
       id: "quoteChoice",
       display: "Choose Quote...",
-      icon: "fa-bars",
+      icon: "fa-align-left",
       subgroup: quoteChoice,
     },
   ],
--- a/frontend/src/scripts/elements/commandline-lists.ts
+++ b/frontend/src/scripts/elements/commandline-lists.ts
@@ -2449,6 +2449,39 @@ const commandsMonkeyPowerLevel: MonkeyTypes.CommandsGroup = {
   ],
 };
 
+export const quoteChoice: MonkeyTypes.CommandsGroup = {
+  title: "Quote...",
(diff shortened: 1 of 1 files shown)


Miodec:
From initial testing:
 - This is getting very laggy when single list command line is enabled. I would only show the first 10 suggestions that match the search.
 - Its using custom mode instead of quote mode.
 - Still getting english quotes after switching to french langauge

kiriDevs:
> This is getting very laggy when single list command line is enabled. I would only show the first 10 suggestions that match the search.

Didn't really happen to me but I expected something like that. Good suggestion on the fix, would most likely also prevent the carnage at the end of the submenu.

> Its using custom mode instead of quote mode

Yep, I wasn't sure how I could set the quote without just putting the text into custom mode, so that was just my initial approach.

---
Also, a kind of meta question: If there even is such a "convention" here - do you want me to merge changes from here into my branch or should I rebase my branch instead?

Miodec:
Merge is fine

kiriDevs:
>  - Still getting english quotes after switching to french langauge

Yup, after testing this amso seems to also be caused by the options only being generated once at page load. However, `Config.language` seems to always be `english` in that scenario, so this seems to happen before the language is loaded.

Another reason that the commands have to be generated on-demand.

PR #2551, opened 2022-02-21
added "show average" setting

OctahedronV2:
### Description

I did the thing, I added the setting!!!


frontend/src/scripts/db.ts
```diff
@@ -313,11 +313,11 @@ export async function getUserAverageWpm10<M extends MonkeyTypes.Mode>(
 
     // Return the last 10 average wpm for quote if the current quote id has never been completed before by the user.
     if (count == 0 && mode == "quote") {
-      return Math.round(last10Wpm / last10Count);
+      return last10Wpm / last10Count;
     }
 
     // Return the average wpm of the last 10 completions for the targeted test mode.
-    return Math.round(wpmSum / count);
+    return wpmSum / count;
```
Miodec, line 320: If youre going to return unrounded data, you need to check where this funciton is used and round it there. Im pretty sure this breaks pace caret.

frontend/src/scripts/elements/commandline-lists.ts
```diff
@@ -358,6 +358,29 @@ const commandsLiveWpm: MonkeyTypes.CommandsGroup = {
   ],
 };
 
+const commandsShowAvg: MonkeyTypes.CommandsGroup = {
+  title: "Show Average...",
```
Miodec, line 362: Lowercase a

Miodec (changes requested):
Also, what happens if you try to use this while signed out?

frontend/src/scripts/db.ts
OctahedronV2: It doesn't break pace caret, I checced. I'm pretty sure if anything it just makes it more accurate lol. :))))))))

frontend/src/scripts/elements/commandline-lists.ts
OctahedronV2: ughhhhhhHHHHHhHHH

OctahedronV2 (commented):
love u mio

frontend/src/scripts/db.ts
Miodec: Please round

Miodec (dismissed):
Thank you mister octa

PR #2553, opened 2022-02-21
Complete Ape Migration

Bruception:
### Description
- Done


frontend/src/scripts/elements/notifications.ts
```diff
@@ -58,7 +58,7 @@ class Notification {
       cls = "bad";
       icon = `<i class="fas fa-fw fa-times"></i>`;
       title = "Error";
-      console.trace(this.message);
```
Bruception, line 61: Changing console.trace to error because it is making the dev console harder to use.

frontend/src/scripts/popups/quote-report-popup.ts
```diff
@@ -118,6 +117,7 @@ async function submitReport(): Promise<void> {
     captcha
   );
   Loader.hide();
+  grecaptcha.reset(CAPTCHA_ID);
```
Miodec, line 120: Wont this cause the captcha to immediately clear when you submit as opposed to it clearing when the popup is invisible?

Miodec (approved):

PR #2571, opened 2022-02-22
display 0 if the time typing is NaN, closes #2567

### LINKED ISSUE #2567: new accounts see NaN in time typing

samarmohan:
### Description

When creating a new account, the total time typing shows `NaN:NaN:NaN`. This PR _**should**_ fix it. See https://github.com/Miodec/monkeytype/issues/2567

Closes #2567

### DIFF
--- a/frontend/src/scripts/account/all-time-stats.ts
+++ b/frontend/src/scripts/account/all-time-stats.ts
@@ -14,9 +14,10 @@ export function update(): void {
     // let th = Math.floor(DB.getSnapshot().globalStats.time / 3600);
     // let tm = Math.floor((DB.getSnapshot().globalStats.time % 3600) / 60);
     // let ts = Math.floor((DB.getSnapshot().globalStats.time % 3600) % 60);
+    const x: number = snapshot.globalStats.time as number; 
     $(".pageAccount .globalTimeTyping .val").text(
       Misc.secondsToString(
-        Math.round(snapshot.globalStats.time as number),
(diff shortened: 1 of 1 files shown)


frontend/src/scripts/account/all-time-stats.ts
```diff
@@ -14,9 +14,10 @@ export function update(): void {
     // let th = Math.floor(DB.getSnapshot().globalStats.time / 3600);
     // let tm = Math.floor((DB.getSnapshot().globalStats.time % 3600) / 60);
     // let ts = Math.floor((DB.getSnapshot().globalStats.time % 3600) % 60);
+    const x: number = snapshot.globalStats.time as number; 
     $(".pageAccount .globalTimeTyping .val").text(
       Misc.secondsToString(
-        Math.round(snapshot.globalStats.time as number),
+        !isNaN(x)) ? Math.round(x) : 0,
```
Miodec, line 20: I think instead of this do a check if `(snapshot?.globalStats?.time ?? 0) === 0`,  if its true dont use `secondsToString` and just insert a `-` instead.

samarmohan: You're the lead so I probably shouldn't question it, but what does `snapshot?.globalStats?.time` contain if there is no data? 0 or `null`/`undefined`

Miodec: That's what the ?? is for. If the expression to the left of it is nullish it outputs the right side.

PR #2573, opened 2022-02-23
remove run-script-os and change the script back

Ferotiq:
### Description

Closes #


Ferotiq:
command prompt can only use double quotes so let's hope there's no conflicts with other shells on that.

Miodec (approved):

Issue #2431, opened 2022-02-07
Convert frontend to TS


monkeytypegeorge:
- [x] Setup compilation (@sondrekje) (#2464)
 - [x] Account (@Ferotiq) (#2494)
 - [x] Controllers (@Miodec, @DanGonite57)
 - [x] Elements (@Miodec)
 - [x] Observables (@Miodec)
 - [x] Pages (@Miodec)
 - [x] Popups (@Miodec)
 - [x] Settings (@rizwanmustafa, @Miodec) (#2499) (#2509)
 - [x] States
 - [x] Test (@Ferotiq, @Miodec) (#2540) (#2601)
 - [x] Root files (@Ferotiq, @Miodec) (#2539)

To allow multiple contributors to work on this, please only work on 1 task at a time.

Miodec:
# On hold by #2462

rizwanmustafa:
I want to work on settings.

Ferotiq:
Finally done

Miodec:
![ablobblewobble-blob](https://user-images.githubusercontent.com/13181393/160007299-32077ea2-3c86-4aae-9359-1f3c55a11470.gif)

PR #2575, opened 2022-02-23
Remove some `@ts-ignores` in frontend scripts

cbebe:
### Description

- Create interfaces for `String` prototype and the `global` object
- Fix function signature of `objectToQueryString`

Part of the little fixes in #2431.

### DIFF
--- a/frontend/src/scripts/index.ts
+++ b/frontend/src/scripts/index.ts
@@ -35,24 +35,26 @@ import "./popups/mobile-test-config-popup";
 import "./popups/edit-tags-popup";
 import * as Account from "./pages/account";
 
-type Global = typeof globalThis & MonkeyTypes.Global;
+type ExtendedGlobal = typeof globalThis & MonkeyTypes.Global;
 
-(global as Global).snapshot = DB.getSnapshot;
+const extendedGlobal = global as ExtendedGlobal;
 
-(global as Global).config = Config;
+extendedGlobal.snapshot = DB.getSnapshot;
 
-(global as Global).toggleFilterDebug = Account.toggleFilterDebug;
+extendedGlobal.config = Config;
 
-(global as Global).glarsesMode = enable;
+extendedGlobal.toggleFilterDebug = Account.toggleFilterDebug;
 
-(global as Global).stats = TestStats.getStats;
+extendedGlobal.glarsesMode = enable;
 
-(global as Global).replay = Replay.getReplayExport;
+extendedGlobal.stats = TestStats.getStats;
 
-(global as Global).enableTimerDebug = TestTimer.enableTimerDebug;
+extendedGlobal.replay = Replay.getReplayExport;
 
-(global as Global).getTimerStats = TestTimer.getTimerStats;
+extendedGlobal.enableTimerDebug = TestTimer.enableTimerDebug;
 
-(global as Global).toggleUnsmoothedRaw = Result.toggleUnsmoothedRaw;
+extendedGlobal.getTimerStats = TestTimer.getTimerStats;
 
-(global as Global).enableSpacingDebug = TestInput.enableSpacingDebug;
+extendedGlobal.toggleUnsmoothedRaw = Result.toggleUnsmoothedRaw;
+
+extendedGlobal.enableSpacingDebug = TestInput.enableSpacingDebug;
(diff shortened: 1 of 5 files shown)


Bruception:
Instead of defining our own partial types for the Firebase user object, do you think we can get firebase types setup?

Bruception:
> It's fixed now! It's a bit hard to commit when the pre-commit hooks fail on linting, though. Is there a way to bypass this without modifying the hooks themselves?

What lint errors are you getting?

package.json
```diff
@@ -10,7 +10,7 @@
     "deploy:live": "cd frontend && npm run deploy:live",
     "install:all": "sh ./bin/install.sh",
     "install:windows": ".\\bin\\install.cmd",
-    "lint": "eslint \"./backend/**/*.{ts,js}\" \"./frontend/src/scripts/**/*.{ts,js}\"",
+    "lint": "",
```
Bruception, line 13: Please revert this change.

cbebe: Yes, on it! Sorry about that

cbebe:
This really could've been a single commit if I wasn't being careless, hopefully my mistakes are squashed out 😂

cbebe:
All the lint errors were on the backend, which was about using `{}` as a type

Bruception:
> All the lint errors were on the backend, which was about using `{}` as a type

Haha yeah there is a PR addressing this up. Do you think you could complete adding explicit Firebase types in the rest of the source code?

frontend/src/scripts/index.ts
```diff
@@ -36,32 +35,43 @@ import "./popups/mobile-test-config-popup";
 import "./popups/edit-tags-popup";
 import * as Account from "./pages/account";
 
-// @ts-ignore
-global.snapshot = DB.getSnapshot;
+// May be pushed to test-timer when it gets converted to TypeScript
+type TimerStats = {
```
Bruception, line 39: Maybe we can move this here: https://github.com/Miodec/monkeytype/blob/master/frontend/src/scripts/types/types.d.ts

Bruception:
> 
> Sure thing. I don't have much experience with using Firebase on the browser but I'm pretty familiar with TypeScript.

It would just require declaring explicit Firebase types, i.e: https://github.com/Miodec/monkeytype/blob/master/frontend/src/scripts/ape/index.ts#L16

There current user is `any` type.

Bruception:
> > It would just require declaring explicit Firebase types, i.e: https://github.com/Miodec/monkeytype/blob/master/frontend/src/scripts/ape/index.ts#L16
> > In that line, current user is `any` type.
> 
> Should I annotate the Firebase types in this PR or create a separate one?

Do it in a separate PR.

frontend/src/scripts/index.ts
```diff
@@ -36,32 +35,24 @@ import "./popups/mobile-test-config-popup";
 import "./popups/edit-tags-popup";
 import * as Account from "./pages/account";
 
-// @ts-ignore
-global.snapshot = DB.getSnapshot;
+type Global = typeof globalThis & MonkeyTypes.Global;
 
-// @ts-ignore
-global.config = Config;
+(global as Global).snapshot = DB.getSnapshot;
```
Bruception, line 40: Instead of (global as Global), would defining something like:

```
const extendedGlobal = global as Global;
extendedGlobal.config = Config
...
```
work?

cbebe: Yep I think it would. Since `global` an object, that would basically just be like aliasing

Ferotiq: I suggested defining a namespace
```ts
namespace global {
  testProp: string;
}

global.testProp = “test”;
```
which I think works, I have done it before in the index.

Bruception: > 

I don't think this works: https://www.typescriptlang.org/play?#code/HYQwtgpgzgDiDGEAEBzANgewEYjUg3gFBJIAu0pACgE4YwBcSUp1AlsCgNyEC+hh6bLgB05ZjTpIAvEkA4BGNKBcAk5A

Bruception (approved):
LGTM IIW

frontend/src/scripts/index.ts
Ferotiq: It might have to use `declare` but it should go at the top of index.ts

PR #2576, opened 2022-02-23
Upgrade mongo library to 4.x.x

Bruception:
### Description

- Migrates db to ts
- Upgrades mongo library to 4.4.0
- Fixes a bug in the frontend.


backend/init/db.ts
```diff
@@ -16,9 +23,7 @@ class DatabaseClient {
       DB_NAME,
     } = process.env;
 
-    const connectionOptions = {
-      useNewUrlParser: true,
```
Bruception, line 20: These options are deprecated. Couldn't find options that replace these.

Miodec (approved):

backend/init/db.ts
Miodec: Thats ok, if its all working.

PR #2577, opened 2022-02-23
Add explicit Firebase types

cbebe:
### Description

- Add explicit types to Firebase `user` and `credentials` variables

Not sure if this is the optimal way to do this since we can get type
inference if we actually import `firebase` (not using `import type`).
Importing `firebase` might mess with how `monkeytype.js` is bundled,
though.

Part of #2431.


Ferotiq:
I'll just close because it'll be fixed in #2603 and there's been no progress on this.

PR #2583, opened 2022-02-24
Added code_matlab

iuuang:
### Description

Some basic MATLAB functions that are used when the program isn't crashing trying to do a sum. Will expand on this as my knowledge expands too.


frontend/static/languages/code_matlab.json
```diff
@@ -0,0 +1,72 @@
+{
+    "name": "code_MATLAB",
```
Bruception, line 2: Naming inconsistency.

Miodec (approved):

PR #2586, opened 2022-02-24
Fix quote ratings

Bruception:
### Description

- Fixes a bug for new users when they first rate a quote. The DB snapshot would not get updated, resulting in incorrect quote rating stats.
- Refactored backend quote rating logic.
- Related to #2430


backend/api/controllers/quotes.ts
```diff
@@ -1,12 +1,63 @@
...
+    _.setWith(
+      userQuoteRatings,
+      `[${language}][${normalizedQuoteId}]`,
+      normalizedRating,
+      Object
+    );
+
+    await UserDAO.updateQuoteRatings(uid, userQuoteRatings);
+
+    return new MonkeyResponse("Rating updated");
```
Miodec, line 58: Should say `Rating submitted` if `userQuoteRatings[language]?.[normalizedQuoteId]` is `undefined`

Miodec (changes requested):

backend/api/controllers/quotes.ts
Bruception: Yes, but the frontend handles the notification logic it seems.

Miodec: Yeeeah but i think the messages should also be correct on the responses.

Miodec (approved):

PR #2588, opened 2022-02-24
Migrate config client to ts

Bruception:
### Description

- Fixes merge configurations when a value is null
- Moved ConfigurationClient to init from DAO to avoid confusion between configs/configuration
- Related to #2430

### DIFF
--- a/backend/handlers/misc.js
+++ b/backend/handlers/misc.js
@@ -25,3 +25,10 @@ export function kogasa(cov) {
     100 * (1 - Math.tanh(cov + Math.pow(cov, 3) / 3 + Math.pow(cov, 5) / 5))
   );
 }
+
+export function identity(value) {
+  return Object.prototype.toString
+    .call(value)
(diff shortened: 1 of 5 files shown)


backend/handlers/misc.js
```diff
@@ -25,3 +25,10 @@ export function kogasa(cov) {
     100 * (1 - Math.tanh(cov + Math.pow(cov, 3) / 3 + Math.pow(cov, 5) / 5))
   );
 }
+
+export function identity(value) {
```
Bruception, line 29: typeof null = "object"
typeof [] = "object"
typeof {} = "object"

This should fix that.

Miodec: What does this function return in those 3 cases?

Miodec (approved):

backend/handlers/misc.js
Bruception: null
object
array

Miodec: Alright nice.

PR #2590, opened 2022-02-26
Move quote moderation logic to quotes

Bruception:
### Description

- Deleted new quotes controller and moved logic into quotes controller.
- Added user permissions middleware.
- Renamed `cannotReport` to `canReport` for clarity.
- Related to #2430


backend/api/routes/quotes.ts
```diff
@@ -127,6 +136,11 @@ quotesRouter.post(
       captcha: joi.string().required(),
     },
   }),
+  checkUserPermissions({
+    criteria: (user) => {
+      return user.canReport;
```
Bruception, line 141: Would migrating from `cannotReport` to `canReport` be an issue?

Miodec (commented):
Is `checkUserPermissions` tested?

Bruception:
> Is `checkUserPermissions` tested?

I’m going to change the flag back to cannot Report since it we would have to add the  canReport flag to all existing users.

Miodec (approved):

PR #2592, opened 2022-02-26
Add dynamic keymap legend style

Ferotiq:
### Description

Adds a dynamic keymap legend style so that when using CapsLock or Shift, the casing changes accordingly.

The only minor issue with this now should be a very slight lag when changing shift or caps state and typing. Very minimal though and is hard to notice even on high refresh rates.


Miodec (approved):
Thanks

PR #2604, opened 2022-02-28
Partial config save

Miodec:
Added support for partial config save to avoid constantly sending full config all the time


backend/dao/config.ts
```diff
@@ -3,9 +3,16 @@ import db from "../init/db";
 
 class ConfigDAO {
   static async saveConfig(uid, config): Promise<UpdateResult> {
+    //https://stackoverflow.com/questions/10290621/how-do-i-partially-update-an-object-in-mongodb-so-the-new-object-will-overlay
+    const temp = {};
```
Bruception, line 7: Refactor:

```js
_.mapKeys(config, (value, key) => `config.${key}`);
```

Ferotiq:
I’ll probably only be able to review this when I get home, so about another 4 hours.

frontend/src/scripts/controllers/preset-controller.ts
```diff
@@ -9,7 +9,7 @@ export function apply(_id: string): void {
   const snapshot = DB.getSnapshot();
   snapshot.presets?.forEach((preset) => {
     if (preset._id == _id) {
-      Config.apply(JSON.parse(JSON.stringify(preset.config)));
+      UpdateConfig.apply(JSON.parse(JSON.stringify(preset.config)));
```
Ferotiq, line 12: what’s with the parse and stringify?

Miodec: Iiiii... dont remember been like this for a while i think

Ferotiq: Odd, try removing it and see if it breaks something.

Miodec: Works fine

Bruception (approved):
LGTM IIW

PR #2626, opened 2022-03-04
Display Dates in 12 Hour Format Instead of 24

UnrealApex:
### Description
Formatted dates so they are in the 12 hour format.


UnrealApex:
@Miodec I acknowledge that using the 24 hour format makes the dates and site look more minimal. If you think using 12 hour dates will make the UI cluttered, we could have the 12 hour dates in tooltips.

Miodec (changes requested):
Move 12h time to tooltips

Bruception:
Why not make this a configuration?

Miodec:
> Why not make this a configuration?

Because this would be an overly specific setting - i want to avoid things like that, there is already a ton of settings.

Bruception:
The build is failing. You will need to address the issues.

UnrealApex:
> The build is failing. You will need to address the issues.

Yep, I'll inspect it.

strafe:
Might be a good time to drop `moment` since it's not being developed anymore, and use JS's native methods to display dates & times in the user's locale.

Miodec:
> Might be a good time to drop `moment` since it's not being developed anymore, and use JS's native methods to display dates & times in the user's locale.

Good idea. Added #2667

Miodec:
Closing due to inactivity

UnrealApex:
> > The build is failing. You will need to address the issues.
> 
> Yep, I'll inspect it.

Lost access to my dev machine, I'll see if I get it back and fix some of the build errors soon.

PR #2632, opened 2022-03-04
Remove potentially offensive words, capitalization

gimmerain4days:
Commented out some words for easier review and reinstatement. 
Capitalized some words that should be capitalized. 
Removed or commented on some abbreviations.
Added comments to words removed and some not removed. If there is a criteria for inclusion/exclusion, words might be found more quickly. (i.e. remove all demonyms)

### DIFF
--- a/frontend/static/languages/english_25k.json
+++ b/frontend/static/languages/english_25k.json
@@ -10,7 +10,7 @@
     "learner",
     "timekeeper",
     "muddle",
-    //"scrotum",//Potentially Offensive
+    "scrotum",
     "cicatrice",
     "archpriest",
     "heterogeneity",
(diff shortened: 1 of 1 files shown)


gimmerain4days:
looks like .json doesn't like comments. I think they are useful, though. I am open to any suggestions.

Bruception:
> looks like .json doesn't like comments. I think they are useful, though. I am open to any suggestions.

No comments please

frontend/static/languages/english_25k.json
```diff
@@ -10,7 +10,7 @@
     "learner",
     "timekeeper",
     "muddle",
-    "scrotum",
+    //"scrotum",//Potentially Offensive
```
Bruception, line 13: Looks like you are still trying to include comments.

gimmerain4days: Just updated to remove the comments.

Miodec (approved):
Thanks

PR #2603, opened 2022-02-28
Convert from Webpack to ESBuild

Ferotiq:
### Description

Yet another bundler change.

The compile times are a wet dream, though.

Closes #2577

PR #2589, opened 2022-02-25
Multiple Custom Themes

rizwanmustafa:
### This PR will enable users to:
- Create / Use / Edit / Delete multiple custom themes (Max: 10)
- Allow users to randomize their custom themes
- Retain previous custom theme functionality for non-logged in users

### TODOs:
- [x] Allow users to create multiple custom themes
- [x] Allow users to edit their multiple custom themes
- [x] Allow users to use one of their multiple custom themes
- [x] Allow users to delete multiple custom themes
- [x] Restore single custom theme functionality for non-logged in users
- [ ] Test and remove any bugs

PR #2638, opened 2022-03-05
Add spacing between lines

Ferotiq:
### Description

Closes #


Ferotiq:
Oh hold by #2603 and #2589

Bruception:
Adding spacing between every line is not a good idea. Lets keep the spacing as is

PR #2639, opened 2022-03-05
Enable strict null checks in backend

Bruception:
### Description

- Enables strict null checks

### DIFF
--- a/backend/dao/quote-ratings.ts
+++ b/backend/dao/quote-ratings.ts
@@ -28,7 +28,7 @@ class QuoteRatingsDAO {
     const quoteRating = await this.get(quoteId, language);
     const average = parseFloat(
       (
-        Math.round((quoteRating.totalRating / quoteRating.ratings) * 10) / 10
+        Math.round((quoteRating!.totalRating / quoteRating!.ratings) * 10) / 10
       ).toFixed(1)
     );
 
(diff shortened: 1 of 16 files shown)


backend/dao/quote-ratings.ts
```diff
@@ -28,7 +28,7 @@ class QuoteRatingsDAO {
     const quoteRating = await this.get(quoteId, language);
     const average = parseFloat(
       (
-        Math.round((quoteRating.totalRating / quoteRating.ratings) * 10) / 10
+        Math.round((quoteRating!.totalRating / quoteRating!.ratings) * 10) / 10
```
Bruception, line 31: Needed because it is not picking up that a quote rating will exist if we get this far into the function.

backend/utils/pb.ts
```diff
@@ -69,8 +71,8 @@ function updatePersonalBest(
   personalBest.consistency = result.consistency;
   personalBest.difficulty = result.difficulty;
   personalBest.language = result.language;
-  personalBest.punctuation = result.punctuation;
-  personalBest.lazyMode = result.lazyMode;
+  personalBest.punctuation = result.punctuation ?? false;
```
Bruception, line 74: Is false okay here?

Miodec (approved):

backend/utils/pb.ts
Miodec: Yarp

Miodec (approved):

PR #2650, opened 2022-03-06
Added 1k and 5k to Persian

AshkanArabim:
### Description

I have added two word-lists for persian_1k and persian_5k respectively.

That's it.


frontend/static/languages/persian_1k.json
```diff
@@ -0,0 +1,1009 @@
+{
+    "name": "persian_5k",
```
Miodec, line 2: Change 5k into 1k

frontend/static/languages/persian_5k.json
```diff
@@ -0,0 +1,5009 @@
+{
+    "name": "persian",
```
Miodec, line 2: Add 5k

Miodec (changes requested):

Miodec (approved):

PR #2656, opened 2022-03-08
Add public api docs

Bruception:
### Description

- Adds public API docs
Screen Shot 2022-03-07 at 8 31 11 PM


backend/api/controllers/leaderboards.ts
```diff
@@ -7,12 +7,14 @@ class LeaderboardsController {
     const { language, mode, mode2, skip, limit = 50 } = req.query;
     const { uid } = req.ctx.decodedToken;
 
+    const queryLimit = Math.min(parseInt(limit as string), 50);
```
Bruception, line 10: We should reflect this in the request validation middleware

Miodec (approved):

PR #2657, opened 2022-03-08
Remove word "war" from ukrainian/russian dictionaries

rowario:


Bruception:
Despite current global events, I don’t think removing the word “war” makes sense here.

rizwanmustafa:
I agree with Bruce here.

PR #2677, opened 2022-03-10
Added curly rule

Miodec:
- Added curly rule
 - Fixed all errors caused by adding the rule

This was all automatic so needs to be reviewed


Bruception:
Thank you, the world is a little better now

Bruception (approved):
LGTM

PR #2682, opened 2022-03-11
E language

fatcatmaht:
This language will REVOLUTIONISE languages on MonkeyType. All languages before this one are STUPID.
 Vote E language for president!


fatcatmaht:
Ignore mattlanguage

that was dumb

Bruception:
I don’t think we should add this. If you a want custom language, you can use the custom text feature.

fatcatmaht:
i disagreee

PR #2699, opened 2022-03-13
Moved JSON list getters to their own file

Miodec:


frontend/src/scripts/controllers/json-lists-controller.ts
```diff
@@ -0,0 +1,137 @@
+import { getJSON } from "../utils/misc";
+import { hexToHSL } from "../utils/misc";
+
+const layoutsListPromise = getJSON(
```
Bruception, line 4: What if one of these promises rejects? We don't have any retry logic.

Miodec: We didnt have any logic before, and for this to fail the server would either need to be down or the user would need to lose connection. Ill try to come up with something.

frontend/src/scripts/controllers/json-lists-controller.ts
```diff
@@ -0,0 +1,137 @@
...
+    return element.name == funbox;
+  });
+}
+
+type Font = { name: string; display?: string };
+const fontsPromise = getJSON("fonts/_list.json") as Promise<Font[]>;
+
+export async function getFonts(): Promise<Font[]> {
+  return (await fontsPromise).sort(function (a: Font, b: Font) {
+    const nameA = a.name.toLowerCase();
```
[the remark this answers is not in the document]
Miodec: Updated

PR #2700, opened 2022-03-13
Refactored random number getting code, added random from array and object

Miodec:


frontend/src/scripts/utils/misc.ts
```diff
@@ -997,3 +997,13 @@ export function shuffle<T>(elements: T[]): void {
     elements[i] = temp;
   }
 }
+
+export function randomElementFromArray<T>(array: T[]): T {
+  return array[randomIntFromRange(0, array.length - 1)];
```
Bruception, line 1002: How does this behave if array is empty?

Miodec: Currently it will return `undefined`

frontend/src/scripts/utils/misc.ts
```diff
@@ -997,3 +997,13 @@ export function shuffle<T>(elements: T[]): void {
...
+}
+
+export function randomElementFromObject(object: {
```
Bruception, line 1005:
```ts
randomElementFromObject<T extends object>(obj: T): T[keyof T]
```

Miodec: Done

frontend/src/scripts/utils/misc.ts
Bruception: Should be fine as long as we handle this possibility when this function is used.

Bruception (approved):
LGTM IIW

Ferotiq (approved):
LGTM also

PR #2707, opened 2022-03-15
Using date-fns instead of moment

### LINKED ISSUE #2667: Replace moment with date-fns
https://www.npmjs.com/package/date-fns

Miodec:
Still need to update chart.js to 3.x to remove moment completely.

Closes #2667


frontend/package.json
```diff
@@ -57,6 +56,7 @@
     "chartjs-plugin-trendline": "^0.2.2",
     "crypto-browserify": "^3.12.0",
     "damerau-levenshtein": "1.0.8",
+    "date-fns": "2.28.0",
```
Bruception, line 59: Nice :)

Miodec:
> How does this package affect the bundle size?

300kb vs 90kb minified. Date fns uses native javascript Date objects to as opposed to moment creating its own special date. It also supports only importing what is needed, which I'm assuming helps with the 'tree shaking' thing.

Even moment devs themselves don't want you to use it anymore and provide a list of alternatives https://momentjs.com/docs/#/-project-status/

Bruception:
LGTM IIW

PR #2710, opened 2022-03-16
Add initial client version tracking

Bruception:
### Description

- Adds client version tracking
Screen Shot 2022-03-16 at 11 00 30 AM


backend/utils/prometheus.ts
```diff
@@ -144,3 +144,13 @@ export function incrementResult(
 
   resultDuration.observe(res.testDuration);
 }
+
+const clientVersionsCounter = new Counter({
```
Bruception, line 148: Not sure what kind of metrics we want to record here.

frontend/src/scripts/version.ts
```diff
@@ -0,0 +1 @@
+export const CLIENT_VERSION = "1.0.0";
```
Bruception, line 1: Need to look into whether we can automatically bump this when a PR merges. Each PR would need to specify whether it is a `major/minor/patch` change.

Miodec: I was thinking to just make the version a timestamp of when the code was deployed

backend/utils/prometheus.ts
Miodec: This is good

Miodec (approved):

PR #2720, opened 2022-03-17
Fix quotes controller

Bruception:
### Description

- Fixes a minor bug where users who had a variation of a language (english1k, etc) could not see quotes.


Ferotiq (approved):
LGTM

PR #2728, opened 2022-03-18
Compile sass using webpack

Bruception:
### Description

- Compiles/Bundles sass source files using webpack
- Minimizes all json/css files


Miodec:
So, as far as i can tell right now if only scss files are changed in the pr the action will do no compilation?

Miodec (approved):

PR #2738, opened 2022-03-20
Enable live reloading

Bruception:
### Description

- Uses Webpack dev server with live reloading

https://user-images.githubusercontent.com/58147810/159149022-bb224da5-87c7-4e33-adb2-4c1c22852ed4.mov

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -8,7 +8,7 @@
     "dep-graph": "madge -c -i \"dep-graph.png\" ./src/scripts",
     "build": "npx gulp build",
     "build:live": "npx gulp build-production",
-    "start:dev": "concurrently \"webpack serve --config=./webpack/config.dev.js\" \"npx gulp watch\"",
+    "dev": "concurrently \"webpack serve --config=./webpack/config.dev.js\" \"npx gulp watch\"",
     "deploy:live": "npm run build:live && firebase deploy -P live --only hosting",
     "tsc": "tsc"
   },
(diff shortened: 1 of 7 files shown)


frontend/webpack/config.base.js
```diff
@@ -16,8 +16,8 @@ const BASE_CONFIGURATION = {
     extensions: [".ts", ".js"],
   },
   output: {
-    path: path.resolve(__dirname, "../public/js/"),
-    filename: "monkeytype.js",
+    filename: "./js/monkeytype.js",
```
Bruception, line 19: This is relative to the path defined below.

frontend/package.json
```diff
@@ -7,7 +7,7 @@
     "dep-graph": "madge -c -i \"dep-graph.png\" ./src/scripts",
     "build": "npx gulp build",
     "build:live": "npx gulp build-production",
-    "start:dev": "concurrently \"npm run build\" \"firebase serve --only hosting\" \"npx gulp watch\"",
+    "start:dev": "concurrently \"webpack serve --config=./webpack/config.dev.js\" \"npx gulp watch\"",
```
Bruception, line 10: Thinking about just naming this script to `dev`

Miodec: Sure, why not. Be sure to update contributing.md and add an echo to start:dev that the command changed.

Miodec (changes requested):
As mentioned in the comment - please update the script names and contributing.md

Bruception:
> As mentioned in the comment - please update the script names and contributing.md

I don't think it makes sense to keep the `start:dev` script. People can just read the package file.

Miodec (approved):

PR #2748, opened 2022-03-22
adding Everblush colorscheme

Mangeshrex:
I am adding the colorscheme [Everblush](https://github.com/Everblush) for its user's and myself. 

Tue Mar 22 01:09:47 PM IST 2022

### DIFF
--- a/frontend/static/themes/_list.json
+++ b/frontend/static/themes/_list.json
@@ -739,4 +739,9 @@
     "bgColor": "#003950",
     "mainColor": "#ffcefb"
   }
-]
+  ,{
+    "name": "Everblush",
+    "bgColor": "#181f21",
+    "mainColor": "#dadada"
+  }
+}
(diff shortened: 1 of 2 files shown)


Bruception:
Please add more screenshots with the other settings enabled.

Miodec:
I think this looks very simmilar to `terror below`

Mangeshrex:
> I think this looks very similar to `terror below`

so it isn't acceptable ?

Mangeshrex:
Wed Mar 23 08:14:03 AM IST 2022

The vscode theme also you can find this theme on nvchad its totally mine and I have handpicked all the colors

Miodec:
> so it isn't acceptable ? I dont think so that the theme is similar to that its actually very different the bg differs a bit and the foreground colors differ a lot.

Im afraid so. Sorry.

Ferotiq:
It does look similar to terror below imo.

PR #2757, opened 2022-03-23
Update test.scss

sarmadhadii:
The timer was too close to the play button when replaying your test.

### Description

Closes #


sarmadhadii:
without-change
with change


First image is before my changes, second one was with my changes.

Miodec (approved):

PR #2763, opened 2022-03-24
Adding Kurdish Language

taboov:
## Adding Kurdish Language
This file contain 200 words of central Kurdish. 

About [Kurdish language](https://en.wikipedia.org/wiki/Kurdish_languages). More specific about this dialect [Central_Kurdish](https://en.wikipedia.org/wiki/Sorani)

Let me know if anything wasn't clear.


Miodec (changes requested):
Please rename to `kurdish_central`. I think there is some code somewhere that splits the language on `_` and takes the first part which wouldn't work with the current name.

taboov:
Central_Kurdish also called "Sorani" if you think one word is better and solve the problem, I can rename them to `sorani`.

Miodec:
> Central_Kurdish also called "Sorani" if you think one word is better and solve the problem, I can rename them to `sorani`.

Yeah lets go with that

Miodec (approved):

roj1512:
@taboov "Sorani" is not a standard.

PR #2765, opened 2022-03-24
Add Redis client and george task queue

Bruception:
### Description

- Adds a Redis client
- Adds a bull queue for george tasks
- Adds a configuration to control whether or not we add tasks into bull queue

### DIFF
--- a/backend/api/controllers/result.ts
+++ b/backend/api/controllers/result.ts
@@ -253,18 +253,16 @@ class ResultController {
       if (isPb && user.discordId) {
         if (useRedisForBotTasks) {
           George.updateDiscordRole(user.discordId, result.wpm);
-        } else {
-          BotDAO.updateDiscordRole(user.discordId, result.wpm);
         }
+        BotDAO.updateDiscordRole(user.discordId, result.wpm);
       }
     }
 
     if (result.challenge && user.discordId) {
       if (useRedisForBotTasks) {
         George.awardChallenge(user.discordId, result.challenge);
-      } else {
-        BotDAO.awardChallenge(user.discordId, result.challenge);
       }
+      BotDAO.awardChallenge(user.discordId, result.challenge);
     } else {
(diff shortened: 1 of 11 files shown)


backend/constants/base-configuration.ts
```diff
@@ -26,6 +26,9 @@ const BASE_CONFIGURATION: MonkeyTypes.Configuration = {
   enableSavingResults: {
     enabled: false,
   },
+  useRedisForBotTasks: {
```
Bruception, line 29: To be removed once we have migrated

backend/api/controllers/result.ts
```diff
@@ -246,12 +251,20 @@ class ResultController {
     if (result.mode === "time" && String(result.mode2) === "60") {
       UserDAO.incrementBananas(uid, result.wpm);
       if (isPb && user.discordId) {
-        BotDAO.updateDiscordRole(user.discordId, result.wpm);
+        if (useRedisForBotTasks) {
```
Bruception, line 254: Wondering if we should dual write here. Just to be able to test that redis is receiving the tasks.

backend/tasks/george.ts
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+class George {
+  jobQueue: Queue;
+
+  initJobQueue(redisConnection: IORedis.Redis | undefined): void {
+    this.jobQueue = new Queue("george-tasks", {
+      connection: redisConnection,
+      defaultJobOptions: {
```
Bruception, line 22: Are these defaults fine?
Retry attempt delays will be: 4 seconds, 8 seconds, 16 seconds.

backend/api/controllers/result.ts
Miodec: Dual write would cause same commands to be executed twice when the configuration is switched

Miodec (approved):

PR #2766, opened 2022-03-24
Add path aliases

Ferotiq:
### Description

In this pr I replace all the relative imports with aliases that look nice.

Closes #


Bruception:
Save exact

package.json
```diff
@@ -29,6 +29,7 @@
     "@typescript-eslint/parser": "^5.11.0",
     "concurrently": "5.3.0",
     "eslint": "8.8.0",
+    "eslint-import-resolver-typescript": "^2.7.0",
```
Bruception, line 32: Save exacf

Bruception (changes requested):
I think there might be a simpler way to achieve this.

Ferotiq:
I might try this later.

PR #2789, opened 2022-03-28
Added Arabic language without tashkeel (Making Arabic usable)

SuperNova74-K:
Arabic Is a very complicated language, there is something in Arabic called "Tashkeel", which are modifiers to the word, so the word itself has the same characters, but different modifiers change it's meaning and pronunciation.

Most of people using Arabic nowadays don't use Tashkeel unless they are journalists or something, but the original Arabic you had on the website was only limited to Tashkeel which made it literally unusable for most of the Arab users including me.

So I copied the existing Arabic database I found on the repo and put it in python, looped over all words and removed Tashkeel form them so that they are usable for the Majority of Arabs, then added the data back into a proper Json file.

Let me know If you have more questions about this, I told my friend about it and he will modify the 10 K Arabic version also found on the repo, so that the 10 K version will also be usable, we hope that you don't remove the Tashkeel version because some people will find it helpful even tho the majority will use the Vanilla version.

note: I told my friend because we are CS students :) and there is a bonus mark on doing a contribution to an existing app.


Miodec:
Have you tried using lazy mode? I think it would result in the same words.

SuperNova74-K:
@Miodec , Nope, I haven't tried it before and yes It solves the Issue, sorry for the inconvenience, I didn't know that there is such a thing, I love the Idea of that mode and it works just the way you expect it to.

Thanks for making such a great website, I love your work man ^_^, you can now close this pull request, It's useless anyway but try to add like a note to make people notice that lazy mode.

Thanks :)

PR #2793, opened 2022-03-30
Using html templating to split html into smaller files

Miodec:


frontend/static/html/head.html
```diff
@@ -0,0 +1,76 @@
...
+    content="A minimalistic, customisable typing website. Test yourself in various modes, track your progress and improve your typing speed."
+  />
+  <meta name="twitter:title" content="Monkeytype" />
+  <meta
+    name="twitter:image"
+    content="https://monkeytype.com/images/mtsocial.png"
+  />
+  <meta name="twitter:card" content="summary_large_image" />
+  <script>
+    if ("serviceWorker" in navigator) {
```
Bruception, line 67: Can we remove this?

Miodec: I may want to revisit the service workers at some point, especially since now the bundles have a chunk hash which could help with the caching issues.

Bruception:
LGTM IIW

Ferotiq (dismissed):
LGTM IIW

Miodec:
> I'm a little confused on what the RemovePlugin does though?

Webpack moves copies the files to the public dir, creates the HTML file from the index template and then removes the HTML directory from /public

Bruception:
Merge conflicts

rizwanmustafa:
Haven't looked into the code, but the features work so LGTM

rizwanmustafa (approved):

PR #2801, opened 2022-04-03
Added: Special Characters for Portuguese (accents and cedilla) with 2…

psudo-dev:
2K words.

### Description

Added: Accents and Cedilla for Portuguese language (2K words). For people wanting to train special characters - which usually slows down typing in Portuguese.

Closes #


Miodec:
Arent those accents already in the lists?

psudo-dev:
> Arent those accents already in the lists?

Do you mean the normal list? Like the Portuguese 3K? They do have some, but in this case it's to exclusively train accents and cedilla, so all the words either have an accent, cedilla or both, it's like a special training.

Miodec (approved):

PR #2803, opened 2022-04-05
Fix links and styles in some pages

jdbruxelles:
- [x] Fixed email links by adding the `mailto:` prefix.
- [x] Added language name to some blocks of code.
- [x] Added the missing style tag in `head` of some pages (see the following screenshots for examples):

Without the style.css file:
Without the style.css file

With the style.css file (fixed):
With the style.css file

(just, I didn't compile, but I think it's good.)


jdbruxelles:
Okay, alright.

PR #2805, opened 2022-04-05
fixed savedtextpopup overflow

chips-wq:
### Description
If there are a lot of custom texts saved the popup is overflowed , causing some of the custom texts to be inaccessible. 
image
I added a max height to the popup and enabled scroll overflow on it so that you can access all of them.
image


Miodec (approved):
Thanks

PR #2822, opened 2022-04-09
Fix circular dep between logger.ts and db.ts

rizwanmustafa:


backend/utils/db-logger.ts
```diff
@@ -0,0 +1,28 @@
...
+
+interface Log {
+  type?: string;
+  timestamp: number;
+  uid: string;
+  event: string;
+  message: string;
+}
+
+const logToDb = async (
```
Bruception, line 12: Why is this necessary?

rizwanmustafa:
@Ferotiq This is a backend issue.

backend/utils/db-logger.ts
rizwanmustafa: There is a circular dependency between db.ts and logger.ts and I don't think we want that.

Bruception: > There is a circular dependency between db.ts and logger.ts and I don't think we want that.

Is it causing any problems? If not I don’t think this ur necessary.

rizwanmustafa: I guess you are right.

PR #2837, opened 2022-04-14
Changes to add new language - Sinhala

Badhi:
### Description

PR contains changes to add new language - Sinhala


Ferotiq:
Please just do the top 200.

Badhi:
Updated the word count to the top 200. May I ask why there is a limit? I thought the minimum count is 200 not the maximum

Miodec:
> Updated the word count to the top 200. May I ask why there is a limit? I thought the minimum count is 200 not the maximum

200 is minimum but if you want to have more words you can create a separate list with `_1k` at the end of the name (that would include 1000 words). Its just a rule we adopted for all languages and want to stick with it for consistency.

Miodec (approved):

PR #2843, opened 2022-04-19
Converted leaderboards DAO to typescript 

Miodec:
Also renamed to DAL


backend/types/types.d.ts
```diff
@@ -93,6 +93,22 @@ declare namespace MonkeyTypes {
     personalBests?: PersonalBests;
   }
 
+  interface LeaderboardEntry {
```
rizwanmustafa, line 96: You can extend the `PersonalBest` interface

backend/dao/leaderboards.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (leaderboardUpdating[`${language}_${mode}_${mode2}`]) return false;
+  if (limit > 50 || limit <= 0) limit = 50;
+  if (skip < 0) skip = 0;
+  const preset = await db
+    .collection<MonkeyTypes.LeaderboardEntry>(
+      `leaderboards.${language}.${mode}.${mode2}`
+    )
+    .find()
+    .sort({ rank: 1 })
+    .skip(parseInt(skip as unknown as string))
```
rizwanmustafa, line 24: `skip` is already an integer. Why would you need to do that?

backend/dao/leaderboards.ts
```diff
@@ -0,0 +1,167 @@
...
+    .limit(parseInt(limit.toString()))
```
rizwanmustafa, line 25: Again, limit is already a number, why reconvert it into one.

backend/types/types.d.ts
Miodec: Ill keep the explicit definition in case the personal best or leaderboard entry format changes

backend/dao/leaderboards.ts
Miodec: Removed

backend/dao/leaderboards.ts
Miodec: Removed

rizwanmustafa (approved):

PR #2844, opened 2022-04-19
Converted new quotes to typescript

Miodec:
Also renamed from DAO to DAL


backend/api/controllers/quote.ts
```diff
@@ -45,7 +52,11 @@ export async function approveQuote(
 
   const { name } = await getUser(uid);
 
-  const data = await NewQuotesDao.approve(quoteId, editText, editSource, name);
+  if (!name) {
+    throw new MonkeyError(500, "Missing name field");
```
Bruception, line 56: 500?

Miodec: Reasoning is that name should be present at this point. The user would not have access to the approve panel without having a name set, but typescript still complains since name is optional. Not sure what error code would be better here.

rizwanmustafa: If you are sure that name is present you could use a separate variable for name and use `as string` to stop typescript from complaining.

Miodec: Will fix it in the next prs that makes names non optional.

PR #2845, opened 2022-04-19
Fix spacing on about page

douira:
### Description

I've fixed the spacing between the people who helped and what they helped with on the about page. The change just concerns how whitespace is handled in html. If you format it with an automatic tool, there's probably a way to remove spaces from inside the `a` tags so that the spacing is correct.

### DIFF
--- a/frontend/static/html/pages/about.html
+++ b/frontend/static/html/pages/about.html
@@ -187,9 +187,8 @@
         href="https://www.reddit.com/user/montydrei"
         target="_blank"
         rel="noreferrer"
+        >Montydrei</a
       >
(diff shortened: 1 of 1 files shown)


Ferotiq:
HTML doesn't render those lines unless it's a `tag`. It is done this way for formatting

douira:
This PR attempts to fix the missing spaces right after the links you can see on this screenshot. From my testing it does fix the issue. Maybe there is a way to fix this with differently formatted html? image

Miodec:
This will get reverted by prettier so this sadly is not a solution to this.

PR #2849, opened 2022-04-20
Fix spacing in Security Policy

kemonprogrammer:
### Description

- Fix the spacing in Security Policy in the header "How to Disclose a Vulnerability"
- As you can see in the screenshot after the long link, there is a space missing
image

- But I haven't tested it by recompiling it, that's why I'm not sure, if it fixes the spacing

- Note: Maybe "Disclose" should also be lowercase


Miodec:
This will get reverted by prettier. Unfortunately its not a fix for this issue.

PR #2857, opened 2022-04-22
Refactor User DAL

rizwanmustafa:
### Description

Closes #

### DIFF
--- a/backend/dao/user.ts
+++ b/backend/dao/user.ts
@@ -6,13 +6,13 @@ import db from "../init/db";
 import MonkeyError from "../utils/error";
 import { DeleteResult, InsertOneResult, ObjectId, UpdateResult } from "mongodb";
 
+const usersCollection = db.collection<MonkeyTypes.User>("users");
+
 export async function addUser(
   name: string | undefined,
   email: string,
   uid: string
 ): Promise<InsertOneResult<MonkeyTypes.User>> {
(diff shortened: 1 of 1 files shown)


Ferotiq (dismissed):
LGTM IIW

backend/dao/user.ts
```diff
@@ -92,42 +87,32 @@ export async function updateQuoteRatings(
...
-    .collection<MonkeyTypes.User>("users")
-    .updateOne({ uid }, { $set: { email } });
+  await usersCollection.updateOne({ uid }, { $set: { email } });
   return true;
 }
 
 export async function getUser(uid: string): Promise<MonkeyTypes.User> {
-  const user = await db.collection<MonkeyTypes.User>("users").findOne({ uid });
+  const user = await usersCollection.findOne({ uid });
   if (!user) throw new MonkeyError(404, "User not found", "get user");
```
Miodec, line 108: One downside of using this `getUser` everywhere is that we loose the detail from the 3rd `MonkeyError` parameter. I think you should add an optional string parameter to `getUser` which will override the stack parameter.

Miodec (changes requested):

backend/dao/user.ts
rizwanmustafa: Will do.

PR #2897, opened 2022-04-28
add scala

1rre:
### Description

Added a "code scala" languages using commonly used Scala keywords, types, methods & symbols.


Miodec:
thanks

PR #2905, opened 2022-04-29
Create husqy.css

huzqy:
new theme based on what people keep asking me for, named husqy so that people can easily search for it and i think it's a cool name for a theme :D
I am in love with this color scheme and it's a cool variation from the normal matrix for people that want a more chill color as opposed to the hacker color
image


huzqy:
added

Miodec (approved):

PR #2914, opened 2022-04-30
add english_contractions word list

monners:
### Description
Adds a word set of English contractions sourced from https://en.wikipedia.org/wiki/Wikipedia:List_of_English_contractions

I've included all contractions listed on that page for completeness, but recognise that some of them might not be desirable (for example: "y'all'd'n't've").  Happy to drop the less common ones, or break this up into multiple lists of common to uncommon


garyinternet:
Even if we just included won't, don't, can't, couldn't, wouldn't, shouldn't, I'm, I've, I'll, I'd, we're, we've, we'll, we'd, you're, you've, you'll, you'd, they're, they've, they'll, they'd, that's

That would be enough to give people practice at all the contractions that they're probably going to end up typing in real life or run into when typing quotes.

Miodec (approved):
Thanks

PR #2915, opened 2022-05-01
Adds a new Languages

BrookFeleke:
### Description
Amharic is the national language and east African country called Ethiopia. I am from there myself and would like to add the language to your beautiful website.


Miodec (approved):
Thanks

PR #2921, opened 2022-05-02
load challenge bug fix

aNOOBisTheGod:
Now, when user loads challenge, website navigate him to the test mode, so challenge loads properly now. Before that change, user could navigate to settings tab, load the challenge and then it would work weird. It was like typing test in settings. Now this function works good

Also, that will be good if somebody will test this cause I did changes in older website version so that works for me and idk if it will work now


aNOOBisTheGod:
Already tested on a new version. Works fine

Miodec (approved):
thanks

PR #2923, opened 2022-05-03
windows webpack error fix

aNOOBisTheGod:
Solution for the webpack similar files error(with index.html) on windows while running ```npm run dev```.

Suggested by miodec


aNOOBisTheGod:
idk maybe we shouldn't rename this file and just add note for windows users that they should rename index to main but that will be strange. Anyways I guess this solution will work everywhere so...

Ferotiq:
> idk maybe we shouldn't rename this file and just add note for windows users that they should rename index to main but that will be strange. Anyways I guess this solution will work everywhere so...

if there’s no drawbacks to renaming the file, then why not?

Miodec (approved):

PR #2929, opened 2022-05-04
quotes:code_javascript: Add regex-heavy quote.

markstos:
### Description

This quote contains a long regex that will be a workout for typing
symbols and numbers.

Source code can be confirmed to available under an open source license
here: https://github.com/markstos/parse-coordinates/blob/master/index.js


Miodec (dismissed):

Miodec:
Length field is incorrect. Please update.

PR #2934, opened 2022-05-04
Replace spaces with underscores for preset/tag names

aNOOBisTheGod:
Now, when user adds new preset or tag, he can enter name with spacebars and they will be replaced to underlining. And also they will store with underlining in database so weird things shouldn't happen. I think it's kinda useful. Tested. Works fine

### DIFF
--- a/frontend/src/ts/config.ts
+++ b/frontend/src/ts/config.ts
@@ -1541,7 +1541,6 @@ export function setFontSize(
   fontSize: MonkeyTypes.FontSize,
   nosave?: boolean
 ): boolean {
-  fontSize = fontSize.toString() as MonkeyTypes.FontSize; //todo remove after around a week
   if (
     !isConfigValueValid("font size", fontSize, [
       ["1", "125", "15", "2", "3", "4"],


frontend/src/ts/ape/endpoints/users.ts
```diff
@@ -70,6 +70,7 @@ export default function getUsersEndpoints(
   }
 
   async function createTag(tagName: string): Ape.EndpointData {
+    tagName = tagName.replaceAll(" ", "_");
```
Bruception, line 73: Avoid reassignment of arguments

frontend/src/ts/constants/firebase-config-example.ts
```diff
@@ -4,11 +4,11 @@
...
 export const firebaseConfig = {
-  apiKey: "",
-  authDomain: "",
-  databaseURL: "",
-  projectId: "",
-  storageBucket: "",
-  messagingSenderId: "",
-  appId: "",
-};
+    apiKey: "",
```
Bruception, line 7: Whats up with this file change?

frontend/src/ts/popups/edit-preset-popup.ts
```diff
@@ -90,22 +90,24 @@ async function apply(): Promise<void> {
   Loader.show();
 
   if (action === "add") {
-    const response = await Ape.presets.add(presetName, configChanges);
-
+    const response = await Ape.presets.add(
+      presetName.replaceAll(" ", "_"),
```
Bruception, line 94: Why are we doing this if it already happens internally?

frontend/src/ts/popups/edit-preset-popup.ts
```diff
@@ -90,22 +90,24 @@ async function apply(): Promise<void> {
...
+      configChanges
+    );
     if (response.status !== 200) {
       Notifications.add("Failed to add preset: " + response.message, -1);
     } else {
       Notifications.add("Preset added", 1, 2);
       snapshotPresets.push({
-        name: presetName,
+        name: presetName.replaceAll(" ", "_"),
```
Bruception, line 102: Try to keep this code DRY

frontend/src/ts/popups/edit-tags-popup.ts
```diff
@@ -73,7 +73,7 @@ async function apply(): Promise<void> {
   Loader.show();
 
   if (action === "add") {
-    const response = await Ape.users.createTag(tagName);
+    const response = await Ape.users.createTag(tagName.replaceAll(" ", "_"));
```
Bruception, line 76: Why? If this happens internally

frontend/src/ts/ape/endpoints/users.ts
aNOOBisTheGod: should I remake this?

frontend/src/ts/popups/edit-preset-popup.ts
aNOOBisTheGod: > Why are we doing this if it already happens internally?

to be confident that there will be no spacebars in the database

Bruception: Calling the same function twice doesn't increase confidence.

Bruception: Calling the same function twice doesn't increase confidence.

frontend/src/ts/ape/endpoints/users.ts
Bruception: Just store in other variable:
`const normalizedTagName = tagName.replaceAll(" ", "_");`

PR #2938, opened 2022-05-04
Convert Spaces to Underscores for Preset and Tag Names

aNOOBisTheGod:
Now, user can enter tag or preset name with spacebars and they will be changed to underlines(and also in database, so there shouldn't be any cringe spacebar bugs). I think that will be useful


Miodec:
Build is failing - please fix

aNOOBisTheGod:
> Build is failing - please fix

done

frontend/src/ts/db.ts
```diff
@@ -125,6 +125,11 @@ export async function initSnapshot(): Promise<
     // LoadingPage.updateText("Downloading tags...");
     snap.customThemes = userData.customThemes ?? [];
     snap.tags = tagsData;
+
+    snap.tags.forEach((tag) => {
+      tag.display = tag.name.replaceAll("_", " ");
```
Bruception, line 130: I don’t understand, display vs name?

aNOOBisTheGod: display is what user will see on website, name is what will be in the database

Bruception: Just store it with underscores

frontend/src/ts/ape/endpoints/presets.ts
```diff
@@ -12,7 +12,7 @@ export default function getPresetsEndpoints(
     configChanges: MonkeyTypes.ConfigChanges
   ): Ape.EndpointData {
     const payload = {
-      name: presetName,
+      name: presetName.replaceAll(" ", "_"),
```
Miodec, line 15: No need to replace in Ape if you already replace it in the popup code right after the user inputs it. Please remove.

Miodec (changes requested):

Miodec:
Have you done any testing?

aNOOBisTheGod:
> Have you done any testing?

Yep. It works for me

Miodec:
Tags not using `display` on the typing page and results page.
image

aNOOBisTheGod:
Done, guess now displays properly

frontend/src/ts/ape/endpoints/presets.ts
aNOOBisTheGod: removed

Miodec (approved):
Thanks

PR #2939, opened 2022-05-04
Add docker-compose and instructions to README

sondrekje:
### Description

As discussed: Simple docker compose setup for running redis and mongoDb through docker. Running Redis and MongoDB locally is now as simple as running ``docker-compose up`` (assuming Docker is already installed).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,17 +84,19 @@ Alternatively, if you use `nvm` then you can run `nvm install` and `nvm use` (yo
    - Click "Generate New Private Key"
    - Save as `serviceAccountKey.json` inside the `backend/credentials/` directory.
 
-#### Mongo Setup
+#### Backend (optional)
 
 Follow these steps if you want to work on anything involving the database/account system. Otherwise, you can skip this section.
 
-1. Install [MongoDB Community Edition](https://docs.mongodb.com/manual/administration/install-community/) and ensure that it is running
-
 1. Inside the backend folder, copy `example.env` to `.env` in the same directory.
 
-   1. If necessary, uncomment the lines in the `.env` file to use credentials to login to MongoDB.
+2. Setup the database server
 
-1. Optional - Install [MongoDB-compass](https://www.mongodb.com/try/download/compass?tck=docs_compass). This tool can be used to see and manipulate your data visually.
+| Local Server                                                                                                                            | Docker (recommended)                                                                                                                                                    |
+| --------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1. Install [MongoDB Community Edition](https://docs.mongodb.com/manual/administration/install-community/) and ensure that it is running | <ol><li>Install [Docker](http://www.docker.io/gettingstarted/#h_installation) on your machine</li><li>Run `docker-compose up` from the `./backend` directory.</li></ol> |
+
+3. (Optional) Install [MongoDB-compass](https://www.mongodb.com/try/download/compass?tck=docs_compass). This tool can be used to see and manipulate your data visually.
    1. To connect, type `mongodb://localhost:27017` in the connection string box and press connect. The monkeytype database will be created and shown after the server is started.
 
 ## Building and Running Monkeytype
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -1,0 +1,29 @@
+version: "3.8"
+services:
+  monkeytype-redis:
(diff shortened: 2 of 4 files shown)


backend/example.env
```diff
@@ -1,5 +1,7 @@
 DB_NAME=monkeytype
+DB_PORT=27017
```
Bruception, line 2: Would DOCKER_DB_PORT be a better name here?

backend/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+
+  monkeytype-mongodb:
+    container_name: monkeytype-mongodb
+    image: mongo:5.0.8
+    restart: always
+    volumes:
+      - mongo-data:/data/db
+    ports:
+      - "${DB_PORT:-27017}:${DB_PORT:-27017}"
+    #command: mongod --smallfiles --logpath=/dev/null
```
Bruception, line 20: Is it okay to remove this comment?

backend/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.8"
+services:
+  monkeytype-redis:
+    container_name: monkeytype-redis
+    image: redis:6.2.6
+    ports:
+      - "${REDIS_PORT:-6379}:${REDIS_PORT:-6379}"
+    restart: always
```
Bruception, line 8: What does this option do?

backend/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.8"
```
Bruception, line 1: With the current setup? Will data persist if a container is shutdown? It would be a bit annoying to have to create a new user everytime in dev environment.

backend/example.env
```diff
@@ -1,5 +1,7 @@
...
 DB_URI=mongodb://localhost:27017
+REDIS_PORT=6379
```
Bruception, line 4: DOCKER_REDIS_PORT?

backend/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+services:
+  monkeytype-redis:
+    container_name: monkeytype-redis
+    image: redis:6.2.6
+    ports:
+      - "${REDIS_PORT:-6379}:${REDIS_PORT:-6379}"
```
Bruception, line 7: Not sure I understand this port mapping.
What exactly is this doing?

backend/example.env
sondrekje: Perhaps! I would have preferred ``MONGODB_PORT`` and for it to be a part of ``MONGODB_URI`` below, but I can change it to ``DOCKER_DB_PORT``.

backend/docker-compose.yml
sondrekje: Gone!

backend/docker-compose.yml
sondrekje: It ensures that the container always restarts, see [documentation](https://docs.docker.com/compose/compose-file/compose-file-v3/) [More](https://docs.docker.com/config/containers/start-containers-automatically/#restart-policy-details). I didn't realize that it would restart when the daemon restarts even if it was manually stopped though - so I'll change it to ``on-failure``.

backend/docker-compose.yml
sondrekje: Yes, it will persist until the volume is manually removed.

backend/docker-compose.yml
sondrekje: It maps from  container port to host machine port. The environment variables will default to 6379 if the environment variable is either unset or empty, see [docs](https://docs.docker.com/compose/environment-variables/#substitute-environment-variables-in-compose-files)

CONTRIBUTING.md
```diff
@@ -84,7 +84,15 @@ Alternatively, if you use `nvm` then you can run `nvm install` and `nvm use` (yo
...
    - Save as `serviceAccountKey.json` inside the `backend/credentials/` directory.
 
-#### Mongo Setup
+### Running Mongo and Redis with Docker
+
+You can choose to run Mongo and Redis in docker containers if you don't want to install Mongo or Redis locally.
+
+1. If you do not already have Docker installed on your local machine, see the [Docker website](http://www.docker.io/gettingstarted/#h_installation) for installation instructions.
+
+2. Run `docker-compose.up` from the `./backend` directory.
```
Bruception, line 93: I had to do: `docker-compose up`

Bruception (approved):
LGTM IIW. Contingent on @Miodec 's approval

Miodec (approved):
Thanks

PR #2940, opened 2022-05-04
Remove class syntax from db client

Bruception:
### Description

Closes #

### DIFF
--- a/backend/dal/ape-keys.ts
+++ b/backend/dal/ape-keys.ts
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import db from "../init/db";
+import * as db from "../init/db";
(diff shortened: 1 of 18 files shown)


backend/dal/ape-keys.ts
```diff
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import db from "../init/db";
+import * as db from "../init/db";
```
Bruception, line 2: Will change this `import { collection } from "../init/db"` In a separate PR

Miodec (approved):
Thanks

backend/dal/ape-keys.ts
Miodec: Okok

PR #2942, opened 2022-05-05
Add test step in pr-check action

Bruception:
### Description

Closes #

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -44,9 +44,9 @@ jobs:
         working-directory: ./frontend/src/ts/constants
         run: mv ./firebase-config-example.ts ./firebase-config-live.ts
 
-      # - name: Check Anti-cheat
-      #   if: steps.filter.outputs.anti-cheat == 'true'
-      #   run: exit 1
+      - name: Check Anti-cheat
+        if: steps.filter.outputs.anti-cheat == 'true'
+        run: exit 1
 
       - name: Cache node modules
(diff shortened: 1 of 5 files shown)


backend/app.ts
```diff
@@ -8,6 +8,8 @@ import errorHandlingMiddleware from "./middlewares/error";
 function buildApp(): express.Application {
   const app = express();
 
+  console.log("TEST!");
```
sondrekje, line 11: why? 😅

.github/workflows/pr-check.yml
```diff
@@ -89,9 +89,11 @@ jobs:
         if: steps.filter.outputs.any-tsscss == 'true'
         run: npm run lint
 
-      - name: Build backend
+      - name: Build/Test backend
         if: steps.filter.outputs.ts-be == 'true'
-        run: npm run pr-check-build-be
+        run: |
+          npm run pr-check-build-be
+          npm run test-backend
```
sondrekje, line 96: would a coverage report be nice to have?

backend/app.ts
Bruception: Build step won't run if no changes to backend directory

.github/workflows/pr-check.yml
Bruception: We should do coverage in a separate PR. Have to think about how we want to set that up.

Miodec (approved):
Looks good, thanks.

PR #2944, opened 2022-05-05
Testing - mock setup for redis

sondrekje:
### Description

run tests against real redis or mock

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -59,6 +59,7 @@
     "@types/swagger-ui-express": "4.1.3",
     "@types/ua-parser-js": "0.7.36",
     "@types/uuid": "8.3.4",
+    "ioredis-mock": "7.4.0",
     "jest": "27.5.1",
     "jest-environment-node": "27.5.1",
     "supertest": "6.2.3",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -35,7 +35,6 @@
     ],
     "semi": ["error", "always"],
     "no-empty": ["error", { "allowEmptyCatch": true }],
-    "no-unused-vars": ["error", { "argsIgnorePattern": "^(_|e|event)" }],
     "no-var": 2,
     "require-path-exists/notEmpty": 2,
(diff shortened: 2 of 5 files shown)


backend/package.json
```diff
@@ -59,6 +59,7 @@
     "@types/swagger-ui-express": "4.1.3",
     "@types/ua-parser-js": "0.7.36",
     "@types/uuid": "8.3.4",
+    "ioredis-mock": "^7.4.0",
```
Bruception, line 62: save exact

backend/__tests__/tasks/george.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+const IS_USING_REDIS_MOCK = process.env.REDIS_URI?.includes("mock") ?? false;
+
+const describeIf = (condition: boolean): jest.Describe =>
+  condition ? describe : describe.skip;
+
+describeIf(!IS_USING_REDIS_MOCK)("George", () => {
+  it("should be able to initJobQueue", async () => {
+    await RedisClient.connect();
+
+    await initJobQueue(RedisClient.getConnection());
```
Bruception, line 14: initJobQueue is not async

backend/__tests__/tasks/george.spec.ts
```diff
@@ -0,0 +1,17 @@
+process.env.MODE = undefined; // Remove dev mode
```
Bruception, line 1: Does this interfere with other tests?

sondrekje: probably yeah. would have preferred to not run tests in dev mode at all, but if we are to do this we need to restore the old environment

backend/__tests__/tasks/george.spec.ts
sondrekje: didn't notice that after rebasing, removed the entire test now.

Bruception (approved):
LGTM IIW

PR #2949, opened 2022-05-06
Add quote from IT Crowd

DioneJM:
### Description

Adding a quote from one of the best scenes in all of television.
[Link to scene](https://www.youtube.com/watch?v=ab8GtuPdrUQ)

### DIFF
--- a/frontend/static/quotes/english.json
+++ b/frontend/static/quotes/english.json
@@ -34974,6 +34974,12 @@
       "source":"Bradely Whitefod",
       "length": 344,
       "id":5933
+    },
+    {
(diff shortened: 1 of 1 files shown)


Miodec:
Length field is incorrect. Please update.

DioneJM:
> Length field is incorrect. Please update.

My bad! Forgot to update the length after fixing the quote. Length field has been updated. Out of curiosity have we thought about having a script that automates the population of some of these fields?

Miodec:
> Out of curiosity have we thought about having a script that automates the population of some of these fields?

Yeah, I should probably look into that.

Miodec (dismissed):
Also, it needs to be a number

Bruception (approved):
Lgtm

PR #2955, opened 2022-05-07
Add theme "10FastFingers"

MaxACartwright:
### Description

Since I've been using 10FastFingers a bit more than usual, and I like the minimalistic approach of monkeytype, I decided to say the same thing as the girl in that one taco ad, "Why Not Have Both?". So, A couple of color picker screenshots later, I pretty closely replicated the 10FastFingers "aesthetic".

image
My theme with Flip test colors on and colorful mode off

image
Colorful mode off and FTC off.

**BTW: THEME NAME DOESN'T NEED TO BE EXACTLY "10FASTFINGERS" I AM FLEXIBLE WITH WHATEVER IT GETS CALLED.**


Miodec:
Sorry, looks too similar to mizu. Plus im not sure about having a theme named after a 'competitor'.

PR #2956, opened 2022-05-07
install-all fix in contributing

aNOOBisTheGod:
you've changed install-windows, but I guess forgot to change install-all, because on my Mac, install:all doesn't work, and install-all works fine


Miodec (approved):
Thanks

PR #2957, opened 2022-05-07
Create preset addition in presets in command line

aNOOBisTheGod:
Now, when user select "Presets..." in command line, he can also create preset, not only choose


Miodec:
You might as well also add that options to tags ;)

aNOOBisTheGod:
here we go. tags added)

Miodec (approved):
thanks

Issue #2954, opened 2022-05-07
Fix duplicate usernames


monkeytypegeorge:
# Execution Plan
- [x] Setup logic to notify users whose name was changed
- [x] Test run the script with fake data
- [x] Schedule an API server maintenance
- [x] Setup backup for user collection in production
- [x] Run the script on production db

# Concerns
- Approximately ~1 million documents in the users collection, will we run into any process resource limits?
- ~~The `addedAt` field may not correctly determine the first user that claimed a name (Due to the Firestore migration)~~

# Script
- Create a file `script.js` in backend directory
- Run `node script.js`
- ¯\\_(ツ)_/¯ Pray

```js
const { MongoClient } = require("mongodb");
const { randomBytes } = require("crypto");
const { performance } = require("perf_hooks");

function randomCharacters() {
    return randomBytes(12).toString("base64url");
}

async function fixNames() {
    const client = new MongoClient("mongodb://localhost:27017", { useNewUrlParser: true });
    await client.connect();
    
    const db = client.db("monkeytype");
    const usersCollection = db.collection("users");

    const start = performance.now();

    const result = await usersCollection.aggregate(
        [
            {
                $sort: {
                    addedAt: 1
                }
            },
            {
                $group: {
                    _id: {  $toLower: "$name" },
                    duplicates: {
                        $push: {
                            _id: "$_id",
                            name: "$name",
                        }
                    },
                    count: {
                        $sum: 1
                    }
                },
            },
            {
                $match: {
                    count: {
                        "$gt": 1
                    }
                }
            }
        ],
        { allowDiskUse: true },
    ).toArray();

    const updates = result.map(({ duplicates }) => {
        const [_, ...offenders] = duplicates;
        return offenders.map(({ _id, name }) => ({
            updateOne: {
                filter: { _id },
                update: {
                    $set: {
                        name: `${name}${randomCharacters()}`,
                        needsToChangeName: true,
                    }
                },
                upsert: false
            }
        }));
    }).flat();

    if (updates.length > 0) {
        await usersCollection.bulkWrite(updates, { ordered: false });
    }

    const end = performance.now();

    await client.close();

    console.log(`Finished in ${(end - start) / 1000} seconds`);
}

fixNames();
```

Bruception:
```js
const { MongoClient } = require('mongodb');
const { randomBytes } = require("crypto");
const { performance } = require("perf_hooks");

function randomCharacters() {
[...]

PR #2958, opened 2022-05-07
Add backend logic to handle duplicate names

Bruception:
### Description

- Adds logic to remove needsToUpdateName field for users that update their name
- Related to issue #2954


Miodec (approved):

Miodec:
Thanks

PR #2968, opened 2022-05-11
add sillyangel to contributors.json

sillyangel:


Miodec:
Ive got a script that updates the list - no need to update manually

PR #2989, opened 2022-05-17
Make min accuracy more intuitive

JesseCSlater:
### Description

I have changed min accuracy mode to have three choices: off, standard, and strict. 
In strict mode, if your accuracy ever falls below the target accuracy, you instantly fail. (The calculation is now done every time a key is pressed instead of based on the timer.
In standard mode, there is a bit of grace given. Grace is given in two ways:
1. Making a single mistake on a test will never fail you. You must make at least two mistakes to fail. 
2. If you drop below your target accuracy, but getting the next word (next 5 characters) correct would bring you back above, you do not fail.


frontend/.firebase-example
```diff
@@ -0,0 +1,5 @@
+{
+	"projects": {
+		"default": "monkeytype-f6501"
```
Bruception, line 3: Remove this file.

frontend/package.json
```diff
@@ -62,7 +62,7 @@
     "crypto-browserify": "3.12.0",
     "damerau-levenshtein": "1.0.8",
     "date-fns": "2.28.0",
-    "firebase": "9.6.0",
+    "firebase": "^9.6.0",
```
Bruception, line 65: What’s up with this change. Please revert.

frontend/src/ts/constants/firebase-config.t
```diff
@@ -0,0 +1,10 @@
+// For Firebase JS SDK v7.20.0 and later, measurementId is optional
+const firebaseConfig = {
```
Bruception, line 2: What’s up with this file change?

JesseCSlater:
I think I have fixed all the accidental changes.

frontend/.firebase-example
JesseCSlater: Done

frontend/package.json
JesseCSlater: Don't know how I even did that

frontend/src/ts/constants/firebase-config.t
JesseCSlater: Deleted

frontend/package-lock.json
```diff
@@ -18,7 +18,7 @@
         "crypto-browserify": "3.12.0",
         "damerau-levenshtein": "1.0.8",
         "date-fns": "2.28.0",
-        "firebase": "9.6.0",
```
Bruception, line 21: You need to change this file too.

JesseCSlater: I think it should be good now

frontend/src/ts/controllers/input-controller.ts
```diff
@@ -507,6 +510,13 @@ function handleChar(char: string, charIndex: number): void {
     return;
   }
 
+  if (!thisCharCorrect && Config.minAcc != "off" && TestLogic.failByMinAcc()) {
```
Bruception, line 513: Use strict equality `!==`

frontend/src/ts/controllers/input-controller.ts
```diff
@@ -181,6 +181,9 @@ function handleSpace(): void {
         TestLogic.fail("difficulty");
         return;
       }
+      if (Config.minAcc != "off" && TestLogic.failByMinAcc()) {
```
Bruception, line 184: Use strict equality `!==`

frontend/src/ts/test/test-logic.ts
```diff
@@ -56,6 +56,19 @@ import { Auth } from "../firebase";
 
 let failReason = "";
 
+export function failByMinAcc() : boolean {
+  if (Config.minAcc == "strict") {
```
Bruception, line 60: Use strict equality `!==`

frontend/src/ts/test/test-logic.ts
```diff
@@ -56,6 +56,19 @@ import { Auth } from "../firebase";
...
+    return TestStats.calculateAccuracy() < Config.minAccCustom;
+  }
+  else if (Config.minAcc == "standard") {
```
Bruception, line 63: Use strict equality `!==`

Bruception:
I think there might be some stylistic issues (spacing, indentation, etc)

PR #2990, opened 2022-05-17
Add pretty check

Bruception:
### Description

Closes #


.github/workflows/pr-check.yml
```diff
@@ -83,6 +83,10 @@ jobs:
         if: steps.filter.outputs.changes != '[]'
         run: npm run install-all
 
+      - name: Pretty Check
+        if: steps.filter.outputs.changes != '[]'
+        run: npm run pretty-code
+
```
Miodec, line 89: Why not use filters like the other pr check steps instead of adding a pretty command just to prettify code?

.github/workflows/pr-check.yml
```diff
@@ -39,6 +39,10 @@ jobs:
               - 'backend/**/*.ts'
             anti-cheat:
               - 'backend/**/anticheat/**'
+            any-src:
+              - 'backend/**/*.{ts,js,json,lua}'
+              - 'frontend/**/*.{ts,js}'
+              - 'frontend/src/**/*.scss'
```
Miodec, line 45: Arent these two lines the same?

Miodec: Wait im stupid

PR #2995, opened 2022-05-19
adds language validation check

tmuplayer33:
### Description
I created this script to check the 'quality' of the language files by counting all duplicates and calculating a duplicate percentage. It might be useful in enforcing properly populated word lists in the future. I did not add it to the `compile-production` task yet as the validation will fail due to russian_10k consisting of ten copies of russian_1k (90% duplicate rate).


frontend/scripts/language-validation.js
```diff
@@ -0,0 +1,69 @@
...
+
+function findDuplicates(arr) {
+  return arr.filter((item, index) => {
+    let indexOf = arr.indexOf(item.toLowerCase());
+    return indexOf !== index && indexOf !== -1;
+  });
+}
+
+function validateLanguageContents(languageNameFilter) {
+  return new Promise((resolve, reject) => {
```
Bruception, line 14: Why are we using promises here?

frontend/scripts/language-validation.js
```diff
@@ -0,0 +1,69 @@
+const fs = require("fs");
+
+const MAX_WORDS_LENGTH = 10000;
+const DUPLICATE_PERCENTAGE_QUALITY_GATE = 10;
+
+function findDuplicates(arr) {
+  return arr.filter((item, index) => {
+    let indexOf = arr.indexOf(item.toLowerCase());
```
Bruception, line 8: const

frontend/scripts/language-validation.js
```diff
@@ -0,0 +1,69 @@
+const fs = require("fs");
+
+const MAX_WORDS_LENGTH = 10000;
+const DUPLICATE_PERCENTAGE_QUALITY_GATE = 10;
+
+function findDuplicates(arr) {
```
Bruception, line 6: This function is slow since it checks all possible pairs of words. Consider a linear time algorithm.

frontend/gulpfile.js
```diff
@@ -25,6 +26,10 @@ task("validate-json-schema", function () {
   return JSONValidation.validateAll();
 });
 
+task("validate-language-contents", function () {
```
Bruception, line 29: Need to setup this check in actions.

frontend/scripts/language-validation.js
```diff
@@ -0,0 +1,69 @@
...
+    ).filter(
+      (language) =>
+        languageNameFilter === undefined ||
+        language.startsWith(languageNameFilter)
+    );
+
+    const results = [];
+    let allLanguageFilesOk = true;
+
+    languagesData.forEach((language) => {
```
Bruception, line 29: const allLanguageFilesOk = languagesData.every((language) => ...);

frontend/scripts/language-validation.js
```diff
@@ -0,0 +1,69 @@
...
+          flag: "r",
+        })
+      );
+
+      // Skip the validation of large language lists
+      if (languageFileData.words.length > MAX_WORDS_LENGTH) return;
+
+      let duplicates = findDuplicates(languageFileData.words);
+
+      let duplicatePercentage = Math.round(
```
Bruception, line 42: const

frontend/scripts/language-validation.js
```diff
@@ -0,0 +1,69 @@
+const fs = require("fs");
+
+const MAX_WORDS_LENGTH = 10000;
```
Bruception, line 3: Why this constraint?

frontend/scripts/json-validation.js
```diff
@@ -523,6 +523,24 @@ function validateLanguages() {
       if (!languageFileValidator.valid) {
         languageFilesAllGood = false;
         languageFilesErrors = languageFileValidator.errors;
+        return;
+      }
+
+      // skip validation of large language lists
+      if (languageFileData.words.length > 10000) return;
+
+      const languageWordsDuplicates = languageFileData.words.filter(
```
Bruception, line 532: Still slow

tmuplayer33: i changed the detection to something faster. guess it would be helpful to actually get a list of found duplicates instead of just counting them

frontend/scripts/json-validation.js
```diff
@@ -523,6 +523,27 @@ function validateLanguages() {
       if (!languageFileValidator.valid) {
         languageFilesAllGood = false;
         languageFilesErrors = languageFileValidator.errors;
+        return;
+      }
+
+      const findDuplicates = (arr) => {
```
Bruception, line 529: Wouldn't this double count some duplicates? If there are three duplicates, the final array will have two of the same element.

```js
const findDuplicates = (words) => {
  const wordFrequencies = {};
  const duplicates = [];
  
  words.forEach((word) => {
    wordFrequencies[word] = (word in wordFrequencies) ? wordFrequencies[word] + 1 : 1;

    if (wordFrequencies[word] === 2) {
      duplicates.push(word);
    }
  });

  return duplicates;
}
```

Bruception:
I think we should include a frequency threshold like you had in the previous implementation. Its not reasonable to go and fix all duplicate words for every language lol

frontend/scripts/json-validation.js
tmuplayer33: yes, you are right! double checked it and found duplicates in the duplicate detection 😄 using your function now

frontend/scripts/json-validation.js
```diff
@@ -523,6 +523,31 @@ function validateLanguages() {
       if (!languageFileValidator.valid) {
         languageFilesAllGood = false;
         languageFilesErrors = languageFileValidator.errors;
+        return;
+      }
+
+      const findDuplicates = (words) => {
```
Bruception, line 529: One last thing 😬

Can we move this function outside of this loop? To the outermost scope.

tmuplayer33:
I re-implemented the treshold.

> We should create an issue and let people that know the languages fix them.
After the languages are fixed, we can lower the treshold and enforce zero duplicates in newly added languages.

Bruception (approved):
LGTM IIW

tmuplayer33:
snippet to make word lists unique:
`echo "$(jq '.words= (.words |unique)' code_html.json)" > code_html.json`

Bruception (dismissed):
RE-LGTM IIW

Miodec (approved):
Thank you

PR #2996, opened 2022-05-19
Remove lua

Bruception:
### Description

- Prettier does not work with lua files.


Miodec (approved):

Ferotiq (approved):
yea, prettier really only works on web development files

PR #2998, opened 2022-05-19
Remove uid from last result

Bruception:
### Description

- Fixes discrepancy between public API documentation data and actual data (unexpected uid field)
Screen Shot 2022-05-19 at 1 00 49 PM


Miodec (approved):
Good catch

PR #3001, opened 2022-05-22
Fixed some words

ItzSphinx:
Deleted and fixed some arabic words that had no meaning, made it easier to read

### Description

Closes #


Miodec:
As discussed on Discord - this achieves the same effect as enabling lazy mode.

PR #3002, opened 2022-05-22
use interfaces instead of types

Ferotiq:
### Description

Closes #

### DIFF
--- a/frontend/src/ts/test/custom-text.ts
+++ b/frontend/src/ts/test/custom-text.ts
@@ -39,7 +39,9 @@ export function setDelimiter(val: string): void {
   delimiter = val;
 }
 
-type CustomTextObject = { [key: string]: string };
+interface CustomTextObject {
+  [key: string]: string;
+}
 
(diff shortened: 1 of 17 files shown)


Bruception:
I don’t know if this is necessary. Should perhaps be done on a case by case basis where we need to extend types.

Ferotiq:
def not necessary but it's good for consistency and it's because it's what interfaces are made for ¯\_(ツ)_/¯

frontend/src/ts/test/custom-text.ts
```diff
@@ -39,7 +39,9 @@ export function setDelimiter(val: string): void {
   delimiter = val;
 }
 
-type CustomTextObject = { [key: string]: string };
+interface CustomTextObject {
```
Bruception, line 42: Might as well change this to:

type CustomTextObject = Record<string, string>;

Miodec: @Ferotiq don't forget to comment and resolve conversations if you addressed them.

PR #3006, opened 2022-05-23
Daily leaderboards pagination

Bruception:
### Description

Closes #


backend/redis-scripts/add-result.lua
```diff
@@ -1,3 +1,4 @@
+local redis_call = redis.call
```
Bruception, line 1: Screen Shot 2022-05-23 at 2 31 52 PM

PR #3015, opened 2022-05-23
Add more long english quotes

yingzhehe:
### Description

Adding new English quotes that are long / thicc in length


Miodec:
Thanks

PR #3049, opened 2022-05-30
Russian language 375k addition

aNOOBisTheGod:
found it in here https://en.wiktionary.org/wiki/Wiktionary:Frequency_lists so some words may not exist, but that will be really small part as I think. Tried to remove bad words but they can still be in there.


aNOOBisTheGod:
ok will fix it now

aNOOBisTheGod:
bruh
wanted to create another PR for that

aNOOBisTheGod:
ea226212322db148a9d7923ca1717cb9818161a4 - in this commit I've renamed some files so now auto-reloading when you change files while debugging on windows works fine. Before this fix  it didn't work

Miodec:
Why the are the pages renamed?

PR #3050, opened 2022-05-30
Changed dark magic girl caret color for visibility issues

wizpizz:
### Description

Since the caret color and the sub color of the theme "dark magic girl" is the same `#93e8d3`, the block caret made it impossible for the user to see the which character they are typing, so I have changed the caret color to the text color `#a288d9` so that the current character is visible again:

### BEFORE
chrome_6763Fygx7v

### AFTER
chrome_foS7l9JCFh


Miodec:
Thanks

PR #3060, opened 2022-06-01
Add get avatar endpoint

Bruception:
### Description

- Adds an endpoint to get and update a user's associated discord avatar.
- Depends on the implicit OAuth2 grant flow


Miodec (approved):
Thanks

PR #3066, opened 2022-06-03
Added an english quote

Jitesh117:
### Description

Added an English quote from Halo 3.


Miodec:
I think this reads quite weirdly without any quotation marks to mark different speakers

Miodec:
Please sync with master, there are conficts.

PR #3068, opened 2022-06-03
Update discord avatar

Bruception:
### Description

- Updating avatar now goes through the same code path as linking


Miodec (approved):
Thanks

PR #3077, opened 2022-06-07
Use maximum tf normalization

Bruception:
### Description

- Quote search now uses a variant of tf-idf
- This should improve search results and make documents that have a search term infrequently more relevant

Example:

Before:
Screen Shot 2022-06-07 at 1 30 07 AM

After:
Screen Shot 2022-06-07 at 1 29 42 AM


Miodec (approved):
Thanks

PR #3084, opened 2022-06-08
Add Custom Filters

nocommentcode:
### Description

"As a user i would like to be able to save filter configurations to be able to generate mutiple reports"

This PR allows users to create custom filters, which are just `ResultFilter` instances.
Previously the front end would store the `ResultFilter` in the local storage on the browser, however this means only 1 filter can be saved, and it is constantly overriden.
This PR now stores these custom filters in the backend database.

# Creating a filter:
- The user can click on "new filter" at any time and this will create a new custom filter based on the current filters applied.
- The user will be prompted to name the filter

https://user-images.githubusercontent.com/26475724/172541220-47027be5-58a5-49a1-b5f8-36dd72f85b72.mp4

# Applying a custom filter:
- The front end gets all the user's triggers, and if the user clicks on one of them, the filter is applied.

https://user-images.githubusercontent.com/26475724/172541380-bf89229f-f025-4f6b-bec4-34294cc6c2f3.mp4

# Deleting a custom filter:
- When a user is selecting a custom filter, the delete button is dsiplayed
- The user can click this button to permanently delete a custom filter

https://user-images.githubusercontent.com/26475724/172541486-3a2cb819-e32c-4b3e-9f62-6519921123d8.mp4

### DIFF
--- a/backend/src/api/controllers/user.ts
+++ b/backend/src/api/controllers/user.ts
@@ -186,6 +186,34 @@ export async function unlinkDiscord(
   return new MonkeyResponse("Discord account unlinked");
 }
 
+export async function getResultFilters(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.ctx.decodedToken;
+  const res = await UserDAL.getResultFilters(uid);
+  return new MonkeyResponse("Found Result filters", res);
+}
+
+export async function addResultFilter(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.ctx.decodedToken;
+  const filter = req.body;
+
+  const createdId = await UserDAL.addResultFilter(uid, filter);
+  return new MonkeyResponse("Created Result Filter", createdId);
+}
+
+export async function removeResultFilter(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.ctx.decodedToken;
+  const { filterId } = req.params;
+
+  const result = await UserDAL.removeResultFilter(uid, filterId);
+  return new MonkeyResponse("Result filter deleted", result);
+}
+
 export async function addTag(
   req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
--- a/backend/src/api/routes/users.ts
+++ b/backend/src/api/routes/users.ts
@@ -10,6 +10,7 @@ import {
 import * as RateLimit from "../../middlewares/rate-limit";
 import apeRateLimit from "../../middlewares/ape-rate-limit";
 import { isUsernameValid } from "../../utils/validation";
+import filterSchema from "../schemas/filter-schema";
 
 const router = Router();
 
@@ -166,6 +167,35 @@ router.delete(
   RateLimit.userClearPB,
   authenticateRequest(),
   asyncHandler(UserController.clearPb)
+);
+
+router.get(
+  "/resultFilters",
+  RateLimit.userCustomFilterGet,
+  authenticateRequest(),
+  asyncHandler(UserController.getResultFilters)
+);
+
+router.post(
+  "/resultFilters",
+  RateLimit.userCustomFilterAdd,
+  authenticateRequest(),
+  validateRequest({
+    body: filterSchema,
+  }),
+  asyncHandler(UserController.addResultFilter)
+);
+
+router.delete(
+  "/resultFilters/:filterId",
+  RateLimit.userCustomFilterRemove,
+  authenticateRequest(),
+  validateRequest({
+    params: {
+      filterId: joi.string().required(),
+    },
+  }),
+  asyncHandler(UserController.removeResultFilter)
 );
 
 router.get(
--- a/backend/src/api/schemas/filter-schema.ts
+++ b/backend/src/api/schemas/filter-schema.ts
@@ -1,0 +1,74 @@
+import joi from "joi";
+
+const FILTER_SCHEMA = {
+  _id: joi.string().required(),
+  name: joi.string().required(),
+  difficulty: joi
+    .object({
+      normal: joi.bool().required(),
+      expert: joi.bool().required(),
+      master: joi.bool().required(),
+    })
+    .required(),
+  mode: joi
+    .object({
+      words: joi.bool().required(),
+      time: joi.bool().required(),
+      quote: joi.bool().required(),
+      zen: joi.bool().required(),
+      custom: joi.bool().required(),
+    })
+    .required(),
+  words: joi
+    .object({
+      10: joi.bool().required(),
+      25: joi.bool().required(),
+      50: joi.bool().required(),
+      100: joi.bool().required(),
+      custom: joi.bool().required(),
+    })
+    .required(),
+  time: joi
+    .object({
+      15: joi.bool().required(),
+      30: joi.bool().required(),
+      60: joi.bool().required(),
+      120: joi.bool().required(),
+      custom: joi.bool().required(),
+    })
+    .required(),
+  quoteLength: joi
+    .object({
+      short: joi.bool().required(),
+      medium: joi.bool().required(),
+      long: joi.bool().required(),
+      thicc: joi.bool().required(),
+    })
+    .required(),
+  punctuation: joi
+    .object({
+      on: joi.bool().required(),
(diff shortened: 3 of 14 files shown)


backend/src/api/controllers/user.ts
```diff
@@ -186,6 +185,34 @@ export async function unlinkDiscord(
...
+  return new MonkeyResponse("Found Result filters", res);
+}
+
+export async function addResultFilter(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.ctx.decodedToken;
+  const filter = req.body;
+
+  const created_id = await UserDAL.addResultFilter(uid, filter);
```
Bruception, line 202: Please use camel case

backend/src/api/routes/users.ts
```diff
@@ -168,6 +169,36 @@ router.delete(
...
+    body: filterSchema,
+  }),
+  asyncHandler(UserController.addResultFilter)
+);
+
+router.delete(
+  "/result-filters/:filterId",
+  RateLimit.userCustomFilterRemove,
+  authenticateRequest(),
+
```
Bruception, line 193: Remove spacs

backend/src/api/schemas/filter-schema.ts
```diff
@@ -0,0 +1,57 @@
+import joi from "joi";
+
+const FILTER_SCHEMA = {
+  _id: joi.string(),
```
Bruception, line 4: Shouldn’t this be required?

backend/src/dal/user.ts
```diff
@@ -150,6 +150,45 @@ export async function isDiscordIdAvailable(
...
+): Promise<ObjectId> {
+  const _id = new ObjectId();
+  await getUsersCollection().updateOne(
+    { uid },
+    { $push: { customFilters: { ...filter, _id } } }
+  );
+  return _id;
+}
+
+export async function getResultFilters(
```
Bruception, line 165: Since result filters are already part of the user document. I think it’s kind of wasteful to request the same resource again. I think it would be better to read result filters from the user document in the frontend and update the snapshot.

Bruception:
Overall nice 😊 

If you can. Could you please also include some unit tests in the backend. Thank you :)

backend/src/dal/user.ts
```diff
@@ -150,6 +150,45 @@ export async function isDiscordIdAvailable(
...
+  if (
+    user.customFilters === undefined ||
+    user.customFilters.filter((t) => t._id.toHexString() == _id).length === 0
+  ) {
+    throw new MonkeyError(404, "Custom filter not found");
+  }
+  return await getUsersCollection().updateOne(
+    {
+      uid: uid,
+      "customFilters._id": new ObjectId(_id),
```
Bruception, line 186: Create object id object once and reuse.

backend/src/api/controllers/user.ts
```diff
@@ -98,7 +98,6 @@ export async function getUser(
   req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
   const { uid } = req.ctx.decodedToken;
-
```
Bruception, line 101: Why this change?

frontend/src/ts/account/result-filters.ts
```diff
@@ -105,6 +109,84 @@ export function load(): void {
...
+    );
+  });
+}
+
+// sets the current filter to be a user custom filter
+export async function setCustomFilter(id: string): Promise<void> {
+  const filter = customFilters.find((filter) => filter._id === id);
+  if (filter) {
+    // deep copy filter
+    filters = JSON.parse(JSON.stringify(filter));
```
Bruception, line 133: I don't think we need to do this

backend/src/api/controllers/user.ts
nocommentcode: sorry i write in python all day ;)

backend/src/dal/user.ts
nocommentcode: But there are times when we need to refresh the custom filters, but not the user. We can optimise to not call this endpoint the first time the app loads, but when we create a new filter or delete a filter, we then need to call getResultFilters() from the front end to update?

backend/src/api/controllers/user.ts
nocommentcode: accident most likely

backend/src/dal/user.ts
Bruception: > But there are times when we need to refresh the custom filters, but not the user. We can optimise to not call this endpoint the first time the app loads, but when we create a new filter or delete a filter, we then need to call getResultFilters() from the front end to update?

Not quite. We keep track of state with the snapshot singleton. We can just add/delete the new custom filter to/from the snapshot. Anytime a user refreshes the page, we will just load the snapshot with the custom filters in the user document.

frontend/src/ts/account/result-filters.ts
nocommentcode: We do need a deep copy of the filter, as if we set the filter to be the custom filter, when the user changes the filter, it edits the front end copy of the custom filter.
This is not the target behaviour, when the user selects a custom filter, and then changes the settings, the custom filter should be de selected and the changes should not apply to the custom filter.
So I need a copy of the object, is there any other way to do this then what ive done im not too familiar with js?
i could do:
`filters = { language : {...filter.language}, funbox: {...filter.funbox} }` and do that for every field but there are so many so i decided to just do `filters = JSON.parse(JSON.stringify(filter));` instead.
What do you think

nocommentcode:
> Please also add a backend configuration to control enabling/disabling this feature. As well as allowing us to impose resource limits.

What do you mean exactly by impose resource limits?

Also for the unit tests, im only performing CRUD operations, is there any point testing that?

frontend/src/ts/account/result-filters.ts
```diff
@@ -2,8 +2,11 @@ import * as Misc from "../utils/misc";
 import * as DB from "../db";
 import Config from "../config";
 import * as Notifications from "../elements/notifications";
-
+import Ape from "../ape/index";
+import { showNewCustomFilterePopup } from "../popups/new-custom-filter-popup";
```
Bruception, line 6: NIT: Add a new line here

backend/src/api/controllers/user.ts
```diff
@@ -186,6 +186,34 @@ export async function unlinkDiscord(
...
+}
+
+export async function removeResultFilter(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.ctx.decodedToken;
+  const { filterId } = req.params;
+
+  const result = await UserDAL.removeResultFilter(uid, filterId);
+  return new MonkeyResponse("Result filter deleted", result);
```
Bruception, line 214: Don't return the update result back to the client

frontend/src/ts/account/result-filters.ts
```diff
@@ -105,6 +109,95 @@ export function load(): void {
...
+// shows popup for user to select name
+export async function startCreateCustomFilter(): Promise<void> {
+  showNewCustomFilterePopup((name: string) => createCustomFilterCallback(name));
+}
+
+// deletes the current selected custom filter
+export async function deleteCustomFilter(): Promise<void> {
+  const result = await Ape.users.removeCustomFilter(filters._id);
+  if (result.status === 200) {
+    getCustomFilters();
```
Bruception, line 176: If successful, just delete the custom filter in the snapshot.

Bruception:
> > Please also add a backend configuration to control enabling/disabling this feature. As well as allowing us to impose resource limits.
> 
> What do you mean exactly by impose resource limits? Also what do you mean by enable disable? If its disabled what should the back end do? Also for the unit tests, im only performing CRUD operations, is there any point testing that?

We want to control how many filters people can create dynamically, and whether or not they can create filters or not.
Checkout the server configuration panel to get a sense of what I mean.

Yes we should still test CRUD operations, because we need to verify that the code we write actually does what we want it to do.

backend/src/dal/user.ts
nocommentcode: ok i understand what you mean, the new PR is using the DB snapshot, and also i removed the getCustomFilters endpoint since it wont be needed

nocommentcode:
This PR is now Split into Front end (#3106 ) and Back end (#3105 )

PR #3101, opened 2022-06-10
Add an option for Arabic without diacritics

T-O-R-U-S:
### Description

In the modern day, most people (and by people I mean Arabic natives) do not type Arabic with its diacritics because they can simply be inferred. If you look at most Arabic text on the internet, you will notice that it is lacking these features present in both Arabic and Arabic 10k. This version of Arabic uses Arabic 10k, minus the diacritics; the only exceptions to these are ً ، ٍ ، and ٌ  since they are still written in text (they're all slightly different, but they add an 'n' sound at the end on top of modifying the pronunciation). 

There are a few odd pieces of text here and there that still make use of it, and every instance of the Qu'ran that you can find will be using them (pronunciation is important there!), but in general, people do not use them.


Miodec:
Already available. Enable "lazy mode" in the settings to remove diacritics from the words.

T-O-R-U-S:
Ah. My apologies...

PR #3104, opened 2022-06-10
extended: language 'code rust'

codybloemhard:
### Description

Extended the rust lang word list.
Added all derives since some were in and some were not.
Added very common traits that are being used all the time.
Added two miscellaneous items: `pub(crate)` and `#[should_panic]`


Bruception (approved):
lgtm

PR #3105, opened 2022-06-11
Custom Filters [Backend]

nocommentcode:
As recommended by @Bruception , This PR is for the Backend and will ovveride #3084 

### Description

"As a user i would like to be able to save filter configurations to be able to generate mutiple reports"

This PR allows users to create custom filters, which are just `ResultFilter` instances.
Previously the front end would store the `ResultFilter` in the local storage on the browser, however this means only 1 filter can be saved, and it is constantly overriden.
This PR now stores these custom filters in the backend database.

# Creating a filter:
- The user can click on "new filter" at any time and this will create a new custom filter based on the current filters applied.
- The user will be prompted to name the filter

https://user-images.githubusercontent.com/26475724/172541220-47027be5-58a5-49a1-b5f8-36dd72f85b72.mp4

# Applying a custom filter:
- The front end gets all the user's triggers, and if the user clicks on one of them, the filter is applied.

https://user-images.githubusercontent.com/26475724/172541380-bf89229f-f025-4f6b-bec4-34294cc6c2f3.mp4

# Deleting a custom filter:
- When a user is selecting a custom filter, the delete button is dsiplayed
- The user can click this button to permanently delete a custom filter

https://user-images.githubusercontent.com/26475724/172541486-3a2cb819-e32c-4b3e-9f62-6519921123d8.mp4

### DIFF
--- a/backend/src/constants/base-configuration.ts
+++ b/backend/src/constants/base-configuration.ts
@@ -42,6 +42,10 @@ export const BASE_CONFIGURATION: MonkeyTypes.Configuration = {
   discordIntegration: {
     enabled: false,
   },
+  customFilters: {
+    enabled: true,
+    maxFiltersPerUser: 0,
+  },
 };
 
 export const CONFIGURATION_FORM_SCHEMA = {
@@ -206,6 +210,21 @@ export const CONFIGURATION_FORM_SCHEMA = {
           type: "number",
           label: "Top Results To Announce",
           min: 1,
+        },
+      },
+    },
+    customFilters: {
+      type: "object",
+      label: "Custom Filters",
+      fields: {
+        enabled: {
+          type: "boolean",
+          label: "Enabled",
+        },
+        maxFiltersPerUser: {
+          type: "number",
+          label: "Max Filters per user",
+          min: 0,
         },
       },
     },
--- a/backend/src/dal/user.ts
+++ b/backend/src/dal/user.ts
@@ -150,6 +150,53 @@ export async function isDiscordIdAvailable(
   return _.isNil(user);
 }
 
+export async function addResultFilter(
+  uid: string,
+  filter: MonkeyTypes.ResultFilters,
+  maxFiltersPerUser: number
+): Promise<ObjectId> {
+  // ensure limit not reached
+  const filtersCount = (
+    (await getUser(uid, "Add Result filter")).customFilters ?? []
+  ).length;
+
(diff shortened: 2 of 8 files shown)


backend/src/constants/base-configuration.ts
```diff
@@ -42,6 +42,10 @@ export const BASE_CONFIGURATION: MonkeyTypes.Configuration = {
   discordIntegration: {
     enabled: false,
   },
+  customFilters: {
+    enabled: true,
+    maxFiltersPerUser: 5,
```
Bruception, line 47: Let’s set the default to 0

backend/src/dal/user.ts
```diff
@@ -150,6 +150,53 @@ export async function isDiscordIdAvailable(
...
+  maxFiltersPerUser: number
+): Promise<ObjectId> {
+  // ensure limit not reached
+  const filtersCount = (
+    (await getUser(uid, "Add Result filter")).customFilters ?? []
+  ).length;
+
+  if (filtersCount >= maxFiltersPerUser) {
+    throw new MonkeyError(
+      503,
```
Bruception, line 165: Let’s do: 409 status code here

Bruception:
Besides my last few comments. Everything else looks solid. :)

backend/src/dal/user.ts
```diff
@@ -150,6 +150,53 @@ export async function isDiscordIdAvailable(
...
+    { uid },
+    { $push: { customFilters: { ...filter, _id } } }
+  );
+  return _id;
+}
+
+export async function removeResultFilter(
+  uid: string,
+  _id: string
+): Promise<UpdateResult> {
```
Bruception, line 181: Make the return type of this function void.

backend/src/constants/base-configuration.ts
```diff
@@ -42,6 +42,10 @@ export const BASE_CONFIGURATION: MonkeyTypes.Configuration = {
   discordIntegration: {
     enabled: false,
   },
+  customFilters: {
```
Bruception, line 45: I think we should move customFilters into the [results sub-configuration](https://github.com/monkeytypegame/monkeytype/blob/16411ca78ecf84979ddf075b0a51c7858754800c/backend/src/constants/base-configuration.ts#L8)

Bruception (approved):
Lgtm iiw

PR #3109, opened 2022-06-12
Added one english quote

Jitesh117:
Added one english quote from "By the river Piedra I sat down and wept" by Paulo Coelho


Bruception (approved):
lgtm

PR #3106, opened 2022-06-11
Custom Filters [FrontEnd]

nocommentcode:
As recommended by @Bruception , This PR is for the FrontEnd and will ovveride #3084 
NOTE: I will rebase this pr onto main once #3105 is merged, so ignore the backend changes in the pr, they are from the previous pr

### Description

"As a user i would like to be able to save filter configurations to be able to generate mutiple reports"

This PR allows users to create custom filters, which are just `ResultFilter` instances.
Previously the front end would store the `ResultFilter` in the local storage on the browser, however this means only 1 filter can be saved, and it is constantly overriden.
This PR now stores these custom filters in the backend database.

#### Updated flow:
- When the user clicks on the preset button, the filter preset section opens up
- The user can click on “save filter as preset” which will prompt them to select a name
- When the user clicks on an existing preset, the preset button is highlighted and the filters are set to the preset
- When the user changes the filters in any way (time frame, language, mode etc) the custom filter button is deselected, as the applied filters are no longer matching the preset 
- When a user selects a preset and its button is highlighted, the “delete preset” button appears
- When the user presses on the delete button, the preset is deleted

https://user-images.githubusercontent.com/26475724/173274961-a31f2989-cb9a-452c-a75f-949705ad7d85.mp4

PR #3110, opened 2022-06-12
Custom Filters Name Validation

nocommentcode:
As raised by @Bruception in #3106 , This PR adds validation to the custom filter names, same as the tage name validation


Bruception:
Thank you! Another thing I missed on the backend PR review:

https://github.com/monkeytypegame/monkeytype/blob/9e0017bd43f6aa37ef23b72c73e9a1a196cd6663/backend/src/constants/base-configuration.ts#L46

this should be false by default

Bruception (approved):
Lgtm

PR #3111, opened 2022-06-12
wordle, wordle_1k languages. 

devkennyy:
@Bruception I hope I've done this right :)


frontend/static/languages/wordle_1k.json
```diff
@@ -0,0 +1 @@
+
```
Bruception, line 1: ?

frontend/static/languages/wordle.json
```diff
@@ -0,0 +1,208 @@
+{
```
Bruception, line 1: I'm not sure if @Miodec would approve of this. Considering these are basically english words.

devkennyy: But they're all 5 letters so it fits a theme.

Miodec (approved):

PR #3123, opened 2022-06-13
Add user profile API

Bruception:
### Description

- Adds the user profile API.

### DIFF
--- a/backend/src/api/controllers/user.ts
+++ b/backend/src/api/controllers/user.ts
@@ -371,3 +371,59 @@ export async function removeFavoriteQuote(
 
   return new MonkeyResponse("Quote removed from favorites");
 }
+
+export async function getProfile(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.params;
+
+  const {
+    name,
+    banned,
+    badgeIds,
+    profile,
+    personalBests,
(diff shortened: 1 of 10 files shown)


backend/src/api/controllers/user.ts
```diff
@@ -371,3 +371,59 @@ export async function removeFavoriteQuote(
...
+    typingStats: {
+      completedTests,
+      startedTests,
+      timeTyping,
+    },
+    details: {
+      bio: "",
+      keyboard: "",
+      socialProfiles: {},
+      ...profile,
```
Miodec, line 405: What is `...profile` ?

backend/src/api/controllers/user.ts
```diff
@@ -371,3 +372,63 @@ export async function removeFavoriteQuote(
...
+    },
+    details: {
+      bio: "",
+      keyboard: "",
+      socialProfiles: {},
+      ...profileDetails,
+    },
+  };
+
+  return new MonkeyResponse("Profile retrieved", profileData);
```
Miodec, line 416: If used is banned, only return name, added at and banned. Also addedAt field is missing in general.

Miodec (changes requested):

Miodec (approved):
Thanks

PR #3129, opened 2022-06-14
Add rate limits to forgot password

### LINKED ISSUE #3013: Add captcha/rate limit to forgot password

mehdihdev:
### Description

I moved all of the forgot password code into a `forgotPassword()` function and then set a rate limit of one call every 5 seconds.

Closes #3013


frontend/src/ts/controllers/account-controller.ts
```diff
@@ -350,6 +351,31 @@ export function signIn(): void {
   });
 }
 
+export function forgotPassword(email: any): void {
+  if (canCall) {
+    console.log(canCall);
```
Bruception, line 356: remove console logs

frontend/src/ts/controllers/account-controller.ts
```diff
@@ -350,6 +351,31 @@ export function signIn(): void {
   });
 }
 
+export function forgotPassword(email: any): void {
```
Bruception, line 354: Make this function async

frontend/src/ts/controllers/account-controller.ts
```diff
@@ -350,6 +351,31 @@ export function signIn(): void {
...
+  if (canCall) {
+    console.log(canCall);
+    if (email) {
+      console.log(email);
+      sendPasswordResetEmail(Auth, email)
+        .then(function () {
```
Bruception, line 360: use async/await

frontend/src/ts/controllers/account-controller.ts
```diff
@@ -52,6 +52,7 @@ import {
 } from "../test/test-config";
 
 export const gmailProvider = new GoogleAuthProvider();
+export let canCall = true;
```
Bruception, line 55: why are we exporting this

mehdihdev:
I made the neccesary changes, let me know if anything else needs changing.

frontend/src/ts/controllers/account-controller.ts
```diff
@@ -350,6 +351,28 @@ export function signIn(): void {
   });
 }
 
+export async function forgotPassword(email: any): Promise<void> {
+  if (canCall) {
+    if (email) {
+      await sendPasswordResetEmail(Auth, email)
+        .then(function () {
```
Miodec, line 358: Pretty sure things will break if you use `await` and `.then` at once - gotta pick one. Also the limit is still applied even if email is empty. 

Todo:
 - Only use `await`
 - Dont set `canCall` to `false` if email is empty
 - instead of using `if (email)`, use `if (!email) return` to avoid nesting

Miodec (changes requested):

mehdihdev:
- [x ]Only use await - replaced .then with a try/catch method
- [x] Dont set canCall to false if email is empty
- [x] instead of using if (email), use if (!email) return to avoid nesting
- [x] Increase limit to 10 seconds

PR #3131, opened 2022-06-15
Added 64K+ Kurdish words

SarkarKurdish:
I've added more than 46K new kurdish words since the were only a few kurdish words


Miodec:
If you are adding a lot of words could you create a new file (kurdish_central_50k) instead of editing this one? This will give the user more options when it comes to the complexity of the word set.

SarkarKurdish:
I've updated the pull request now.

kurdish_central.json            -  This is now same as the original so no change was made.
kurdish_central_45k.json     - This is the new file with the new words

roj1512:
@SarkarKurdish These words include characters that aren’t available on Kurdish keyboard layouts.

SarkarKurdish:
@roj1512 The data were given to me so I was not awere of that. I'm giving it a revew once I had time.

roj1512:
@SarkarKurdish You need to replace ك with ک, and ھ with ه, remove all U+0200C, and make sure each word matches `/^[قوەرڕتیئحعۆپاسشدفگغهژکلڵزخجچڤبنم]{2,}$/`.

PR #3135, opened 2022-06-16
Add profile editing

Bruception:
### Description

Closes #


frontend/src/ts/popups/edit-profile-popup.ts
```diff
@@ -0,0 +1,115 @@
...
+    keyboard,
+    socialProfiles: {
+      twitter,
+      github,
+      website,
+    },
+  };
+
+  return updates;
+}
```
Miodec, line 75: Should probably add here some validation like removing excessing new lines, protecting against zalgo, use trim() and others

Miodec (changes requested):

frontend/src/ts/popups/edit-profile-popup.ts
Bruception: This validation should be done in backend too.

PR #3159, opened 2022-06-19
added missing challenge 

devkennyy:
also added messages for accuracy tests and autorole for accuracy test

### DIFF
--- a/frontend/static/challenges/_list.json
+++ b/frontend/static/challenges/_list.json
@@ -692,4 +692,11 @@
     "type": "customTime",
     "parameters": [3600]
   }
+  ,{
+    "name": "lookAtMeIAmTheDeveloperNow",
+    "display": "Look at me. I am the developer now.",
+    "autoRole": true,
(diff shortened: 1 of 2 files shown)


frontend/static/challenges/_list.json
```diff
@@ -358,6 +370,13 @@
     "type": "script",
     "parameters": ["shrek.txt",null,null]
   }
+  ,{
```
Bruception, line 373: What is this?

devkennyy: The same thing you see on line 175 and line 191

Bruception: I mean the challenge? Was adding this new challenge discussed? Also the related file: `sourcecode.txt` is empty

devkennyy: This challenge was already made, it was just missing from the site. Fixing the empty file now!

Miodec (approved):
Thanks

PR #3165, opened 2022-06-19
Fix escaped HTML for banner notifications

Bruception:
### Description

- Only escapes html for non-banner notifications

Before:
Screen Shot 2022-06-19 at 7 24 52 PM

After:
Screen Shot 2022-06-19 at 7 25 09 PM

### DIFF
--- a/frontend/src/ts/elements/notifications.ts
+++ b/frontend/src/ts/elements/notifications.ts
@@ -34,7 +34,7 @@ class Notification {
   ) {
     this.type = type;
     this.message =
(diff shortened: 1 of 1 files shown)


devkennyy (approved):
Looks great!

rizwanmustafa:
@Bruception I think a better solution would be to have an optional boolean parameter: `allowHTML`. If it is true, we won't escape html, else we will.

rizwanmustafa:
I have created a better version of this pr. Check it out: #3168

PR #3167, opened 2022-06-20
Fix webpack hot reload not working on html file changes.

rizwanmustafa:
Credits to @devkennyy for leading me to mess around with webpack and fix this 

The `monkeytype.js` bundle has to be included in html files for the webpack live reload to work. I inferred this from the tests. Hence I have began including `monkeytype.js` bundle in html files in `development` mode but not in production mode.

This will provide us with the best of the both worlds as we will be able to enjoy live reload in html files in development mode and not have to bear unnecessary load on the server in production for serving javascript bundles.


Miodec (approved):
Thanks

PR #3171, opened 2022-06-20
Fix logic to not delete user if DB goes down

Bruception:
### Description

Closes #


backend/src/api/controllers/user.ts
```diff
@@ -103,13 +103,17 @@ export async function getUser(
...
+      await admin.auth().deleteUser(uid);
+      throw new MonkeyError(
+        404,
+        "User not found. Please try to sign up again.",
+        "get user",
+        uid
+      );
+    }
+
+    throw e;
```
Bruception, line 116: Let the error propagate to our error handling middleware

Miodec (approved):

PR #3174, opened 2022-06-20
Profile

Bruception:
### Description

Closes #


Bruception:
There is a race condition with leaderboard avatars bcs img load callback is asynchronous.

Miodec (approved):

PR #3177, opened 2022-06-21
add languages: arabic_no_diacritics, arabic_no_diacritics_10k

yamenw:
### Description

Added languages `arabic_no_diacritics`  and `arabic_no_diacritics_10k`, they are exactly the same as `arabic`  and `arabic_10k`, but with the diacritics stripped as they are rarely used in modern settings outside certain formal scenarios, (said diacritics seem to 
 have been contributed here https://github.com/monkeytypegame/monkeytype/commit/68bdb6f083c157883e46be5e7a783013572dcf9e ), I tried to stick to the implied naming scheme as much as possible.


Miodec:
"No diacritics" version is already available by enabling "lazy mode" in the settings.

PR #3192, opened 2022-06-22
removed non-existent challenge

devkennyy:
image

this challenge is _not_ a challenge. i think i may have added it in one of my pr (silly me)

the challenges also check for accuracy but that's not necessary since it's master mode. 

spotted by Kelsier


Miodec:
This is a generic "accuracy" challenge rules and should be left in. Its used by some challenges.

PR #3194, opened 2022-06-22
replaced show yesterday toggle with view today/yesterday button

devkennyy:
I may have missed some things because my leaderboard entries were not working so feel free to test :)

### DIFF
--- a/frontend/src/styles/z_media-queries.scss
+++ b/frontend/src/styles/z_media-queries.scss
@@ -293,7 +293,7 @@
         .buttonGroup {
           grid-auto-flow: column;
         }
-        .showYesterdayButton {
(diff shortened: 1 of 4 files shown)


frontend/src/styles/z_media-queries.scss
```diff
@@ -293,7 +293,10 @@
         .buttonGroup {
           grid-auto-flow: column;
         }
-        .showYesterdayButton {
+        .viewYesterdayButton {
```
Bruception, line 296: I think we can just have one button and just update the text when toggling.

devkennyy: That's what I was going to do but I feared there might be some errors. Okay, I will restart.

PR #3195, opened 2022-06-22
replaced theme generation method

### LINKED ISSUE #3128: Make sure random theme doesnt repeat until all themes were randomised

devkennyy:
i am having a hard time understanding where i should be declaring the `randomThemeIndex` variable. 

Closes #3128


Bruception:
Given that  themes are shuffled we need to maintain this as state. This state needs to change when a config changes.

devkennyy:
Right...that sounds complicated. Can you do it and then I'll understand??? 😉

frontend/src/ts/controllers/theme-controller.ts
```diff
@@ -212,16 +213,19 @@ export function randomizeTheme(): void {
...
 
-    const previousTheme = randomTheme;
-    randomTheme = Misc.randomElementFromArray(randomList);
+    //! setting randomThemeIndex to 0 everytime randomizeTheme is called
+
+    const randomTheme = randomList[randomThemeIndex];
+    randomThemeIndex++;
+
+    if (randomThemeIndex >= randomList.length) {
+      Misc.shuffle(randomList);
```
Bruception, line 222: This doesn't quite work because of this:
https://github.com/monkeytypegame/monkeytype/blob/ba000381e62587cb45c919172feebebfdbb6ef03/frontend/src/ts/controllers/theme-controller.ts#L197

The random list is always reset when `randomizeTheme` is called.

Bruception (approved):
Lgtm

PR #3206, opened 2022-06-23
Adds a reset account button to the settings page

### LINKED ISSUE #3173: Add button to reset account

YummyOreo:
### Description

Adds reset button

Closes #3173


backend/src/dal/user.ts
```diff
@@ -47,6 +47,33 @@ export async function deleteUser(uid: string): Promise<DeleteResult> {
   return await getUsersCollection().deleteOne({ uid });
 }
 
+export async function resetUser(uid: string): Promise<UpdateResult> {
```
Bruception, line 50: Don't you also have to delete all result data?

YummyOreo: that is not done there

YummyOreo: It is done after resetting the stats

YummyOreo:
ok everything works

backend/src/dal/user.ts
Bruception: I don’t see any code that deletes all result data for the user.

YummyOreo: should i move it to that place

Miodec (changes requested):

backend/src/dal/user.ts
```diff
@@ -47,6 +47,36 @@ export async function deleteUser(uid: string): Promise<DeleteResult> {
...
+        lbMemory: {},
+        bananas: 0,
+        profileDetails: {
+          bio: "",
+          keyboard: "",
+          socialProfiles: {},
+        },
+      },
+    }
+  );
```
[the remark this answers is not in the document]
YummyOreo: Ok

backend/src/api/routes/ape-keys.ts
```diff
@@ -60,6 +60,14 @@ router.post(
   asyncHandler(ApeKeyController.generateApeKey)
 );
 
+router.delete(
+  "/",
```
Bruception, line 64: I think this will conflict with the existing delete endpoint.

backend/src/api/controllers/ape-key.ts
```diff
@@ -82,3 +82,13 @@ export async function deleteApeKey(
 
   return new MonkeyResponse("ApeKey deleted");
 }
+
+export async function resetApeKeys(
```
Bruception, line 86: reset -> deleteAll

backend/src/api/controllers/ape-key.ts
```diff
@@ -82,3 +82,13 @@ export async function deleteApeKey(
...
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { uid } = req.ctx.decodedToken;
+
+  await ApeKeysDAL.resetApeKeys(uid);
```
Bruception, line 91: reset -> deleteAll

backend/src/api/controllers/preset.ts
```diff
@@ -42,3 +42,11 @@ export async function removePreset(
 
   return new MonkeyResponse("Preset deleted");
 }
+
+export async function reset(req: MonkeyTypes.Request): Promise<MonkeyResponse> {
```
Bruception, line 46: reset -> deleteAll

backend/src/api/routes/ape-keys.ts
YummyOreo: ok

backend/src/api/routes/ape-keys.ts
```diff
@@ -60,6 +60,14 @@ router.post(
   asyncHandler(ApeKeyController.generateApeKey)
 );
 
+router.delete(
+  "/deleteAll",
```
Bruception, line 64: Sorry for the confusion, let’s remove the deleteAll path and just leave it as /

frontend/src/ts/popups/simple-popups.ts
```diff
@@ -688,6 +688,102 @@ list["deleteAccount"] = new SimplePopup(
...
+          user.email as string,
+          password
+        );
+        await reauthenticateWithCredential(user, credential);
+      } else {
+        await reauthenticateWithPopup(user, AccountController.gmailProvider);
+      }
+      Loader.show();
+      Notifications.add("Reseting account and stats...", 0);
+      const responses = [await Ape.users.reset()];
```
Bruception, line 720: If you use the await keyword, this will not do it parallel.

Just one await:

await Promise.all(responses);

backend/src/api/routes/ape-keys.ts
YummyOreo: Ok

backend/__tests__/dal/ape-keys.spec.ts
```diff
@@ -18,5 +18,38 @@ describe("ApeKeysDal", () => {
...
+        enabled: true,
+      },
+    ];
+
+    const apeKeyId0 = await addApeKey(apeKeys[0]);
+    const apeKeyId1 = await addApeKey(apeKeys[1]);
+
+    await deleteAllApeKeys("123");
+
+    expect(apeKeyId0).toBe(undefined);
```
Bruception, line 52: This isn’t really testing what we want. You need to make a query to the DB and verify that their is no ape keys

YummyOreo: Ok it may be a couple if days till I can fix it.

YummyOreo: Would it be undefined if it was deleted tho

frontend/src/ts/popups/simple-popups.ts
```diff
@@ -688,6 +688,102 @@ list["deleteAccount"] = new SimplePopup(
...
+        snap.discordId = undefined;
+        AccountButton.update();
+        DB.setSnapshot(snap);
+        Settings.updateDiscordSection();
+      }
+      responses.push(Ape.results.deleteAll());
+
+      const results = await Promise.all(responses);
+
+      const failedResponse = (await results).find(
```
Bruception, line 741: Do we need an extra await here? Wouldn't the promises be resolved?

backend/__tests__/dal/ape-keys.spec.ts
Bruception: > Would it be undefined if it was deleted tho

I'm not sure (I think the library we are using stores everything in memory). But we should use the database API to verify that they are actually deleted.

backend/src/dal/user.ts
```diff
@@ -47,6 +47,39 @@ export async function deleteUser(uid: string): Promise<DeleteResult> {
   return await getUsersCollection().deleteOne({ uid });
 }
 
+export async function resetUser(uid: string): Promise<UpdateResult> {
+  return await getUsersCollection().updateOne(
```
Bruception, line 51: Let's not return anything from this function.

Return type should be `Promise<void>`

frontend/src/ts/popups/simple-popups.ts
```diff
@@ -688,6 +688,102 @@ list["deleteAccount"] = new SimplePopup(
...
+        const credential = EmailAuthProvider.credential(
+          user.email as string,
+          password
+        );
+        await reauthenticateWithCredential(user, credential);
+      } else {
+        await reauthenticateWithPopup(user, AccountController.gmailProvider);
+      }
+      Loader.show();
+      Notifications.add("Reseting account and stats...", 0);
```
Bruception, line 719: Resting -> Resetting

frontend/src/ts/popups/simple-popups.ts
```diff
@@ -688,6 +688,102 @@ list["deleteAccount"] = new SimplePopup(
...
+          user.email as string,
+          password
+        );
+        await reauthenticateWithCredential(user, credential);
+      } else {
+        await reauthenticateWithPopup(user, AccountController.gmailProvider);
+      }
+      Loader.show();
+      Notifications.add("Resetting account and stats...", 0);
+      const responses = [Ape.users.reset()];
```
Bruception, line 720: can this just be:

```ts
const responses = [
    Ape.users.reset(),
    Ape.presets.reset(),
    Ape.apeKeys.reset(),
    Ape.results.deleteAll(),
];
```

backend/__tests__/dal/ape-keys.spec.ts
YummyOreo: So like try and get them? Using getApeKeys?

Bruception (changes requested):

frontend/src/ts/popups/simple-popups.ts
YummyOreo: Thanks never would have thought of that

frontend/src/ts/popups/simple-popups.ts
```diff
@@ -688,6 +688,102 @@ list["deleteAccount"] = new SimplePopup(
...
+        await reauthenticateWithCredential(user, credential);
+      } else {
+        await reauthenticateWithPopup(user, AccountController.gmailProvider);
+      }
+      Loader.show();
+      Notifications.add("Resetting account and stats...", 0);
+      const responses = [
+        Ape.users.reset(),
+        Ape.presets.reset(),
+        Ape.apeKeys.reset(),
```
Bruception, line 723: Can we change
reset -> deleteAll

YummyOreo: Yeah

YummyOreo: Should i do it for users because its not deleting all its just deleting some stuff

Bruception: Not for users

backend/src/api/routes/ape-keys.ts
```diff
@@ -60,6 +60,14 @@ router.post(
   asyncHandler(ApeKeyController.generateApeKey)
 );
 
+router.delete(
+  "/",
+  RateLimit.apeKeysDelete,
```
Bruception, line 65: Rate limiter middleware should come after authentication.

backend/src/api/routes/presets.ts
```diff
@@ -58,6 +58,13 @@ router.patch(
   asyncHandler(PresetController.editPreset)
 );
 
+router.delete(
+  "/",
+  RateLimit.presetsRemove,
```
Bruception, line 63: All rate limiter middlewares need to come after

frontend/src/ts/popups/simple-popups.ts
```diff
@@ -688,6 +688,102 @@ list["deleteAccount"] = new SimplePopup(
...
+
+      const results = await Promise.all(responses);
+
+      const failedResponse = results.find(
+        (response) => response.status !== 200
+      );
+      if (failedResponse) {
+        Loader.hide();
+        return Notifications.add(
+          "There was a error Resetting your account. Try retrying.",
```
Bruception, line 747: "There was an error resetting your account. Please try again."

backend/__tests__/dal/ape-keys.spec.ts
```diff
@@ -1,5 +1,8 @@
 import { ObjectId } from "mongodb";
-import { addApeKey } from "../../src/dal/ape-keys";
+import { addApeKey, deleteAll } from "../../src/dal/ape-keys";
+import * as db from "../../src/init/db";
+
+const COLLECTION_NAME = "ape-keys";
```
Bruception, line 5: You should do something similar for ape keys like we did for users:
https://github.com/monkeytypegame/monkeytype/blob/master/backend/src/dal/user.ts#L14

Bruception:
You need to merge the latest changes. There was a change to our rate limit flow:
https://github.com/monkeytypegame/monkeytype/blob/master/backend/src/middlewares/rate-limit.ts

backend/__tests__/dal/ape-keys.spec.ts
```diff
@@ -19,4 +23,41 @@ describe("ApeKeysDal", () => {
...
+        hash: "11113",
+        createdOn: Date.now(),
+        modifiedOn: Date.now(),
+        lastUsedOn: Date.now(),
+        useCount: 0,
+        enabled: true,
+      },
+    ];
+
+    const results = [addApeKey(apeKeys[0]), addApeKey(apeKeys[1])];
```
Bruception, line 52:
```ts
const results = apeKeys.map(addApeKey);
```

backend/src/dal/ape-keys.ts
```diff
@@ -94,3 +104,7 @@ export async function deleteApeKey(uid: string, keyId: string): Promise<void> {
     throw new MonkeyError(404, "ApeKey not found");
   }
 }
+
+export async function deleteAll(uid: string): Promise<void> {
+  await db.collection<MonkeyTypes.ApeKey>(COLLECTION_NAME).deleteMany({ uid });
```
Bruception, line 109: Could you refactor this to use `getApeKeysCollection`

Bruception:
Code looks good to me. Needs @Miodec’s approval for client side experience.

On a separate note, do we have any other mechanisms to protect compromised accounts from account data deletion?

Right now it seems like there’s only client side checks but a compromised account’s authentication token can still be used in another client. Maybe we can look into a more robust server side validation check. @Miodec do you know of any firebase APIs that we can use for this? Perhaps 2FA with email pass code?

YummyOreo:
> On a separate note, do we have any other mechanisms to protect compromised accounts from account data deletion?

you could send a email to the user asking for them to verify that its them

YummyOreo:
> If we do go this route, perhaps we can consolidate all the account reset logic into a single API endpoint.

consolidating the reset to a single API endpoint would be easy. But we would have find a way to just test the ape-key resetting by its self.

YummyOreo:
also if we did add 2fa for account deletion, why not add a option to add it on logging in with text message or google Authenticator app

Miodec:
> Code looks good to me. Needs @Miodec’s approval for client side experience.
> 
> On a separate note, do we have any other mechanisms to protect compromised accounts from account data deletion?
> 
> Right now it seems like there’s only client side checks but a compromised account’s authentication token can still be used in another client (therefore bypassing authentication). Maybe we can look into a more robust server side validation check. @Miodec do you know of any firebase APIs that we can use for this? Perhaps 2FA with email pass code?
> 
> If we do go this route, perhaps we can consolidate all the account reset logic into a single API endpoint.

I mean, do we have a account token leak?

YummyOreo:
> I mean, do we have a account token leak? We reauthenticate before doing the dangerous stuff, maybe there is a way to check how old the token is

I think even if we reauthenticate it would be safe to do 2fa for that because they may get access to just their account and not their email or phone (however we would do it).

Miodec:
> > I mean, do we have a account token leak? We reauthenticate before doing the dangerous stuff, maybe there is a way to check how old the token is
> 
> 
> 
> I think even if we reauthenticate it would be safe to do 2fa for that because they may get access to just their account and not their email or phone (however we would do it).

The issue is Firebase doesn't have any easy way of doing 2fa

Bruception:
> What about google Authenticator

No longer needed. We added a small protection check in the backend that requires the issued token to be recent. This should help a bit.

YummyOreo:
> > What about google Authenticator
> 
> No longer needed. We added a small protection check in the backend that requires the issued token to be recent. This should help a bit.

ok

Miodec (changes requested):
After looking into this more, i think it would be better to just hit one endpoint `/user/reset` and do all the changes from there

backend/src/dal/user.ts
```diff
@@ -41,6 +44,43 @@ export async function deleteUser(uid: string): Promise<void> {
...
+          socialProfiles: {},
+        },
+        favoriteQuotes: {},
+        customThemes: [],
+        tags: [],
+      },
+    }
+  );
+
+  await getApeKeysCollection().deleteMany({ uid });
```
Bruception, line 79: Given that these collections are separate from each other, I think we can parallelize these requests.

backend/src/dal/user.ts
```diff
@@ -41,6 +44,43 @@ export async function deleteUser(uid: string): Promise<void> {
...
+  await getPresetsCollection().deleteMany({ uid });
+  await db.collection<MonkeyTypesResult>("results").deleteMany({ uid });
```
Bruception, line 81: On a separate note, I think it is weird that we delete presets and ape keys. Those are deletable on their own. @Miodec I think we should only delete results and user data.

YummyOreo: If it is a full account reset, then it would make sense to reset everything expect login stuff.

backend/src/dal/user.ts
```diff
@@ -41,6 +44,43 @@ export async function deleteUser(uid: string): Promise<void> {
...
+        },
+        lbPersonalBests: {
+          time: {},
+        },
+        completedTests: 0,
+        startedTests: 0,
+        timeTyping: 0,
+        lbMemory: {},
+        bananas: 0,
+        profileDetails: {
```
Bruception, line 67: profileDetails can just be an empty object.

backend/src/dal/user.ts
YummyOreo: isnt it already doing that?

backend/src/dal/user.ts
YummyOreo: I cant be an empty object

YummyOreo: It cant be an empty object

backend/src/dal/user.ts
Miodec: @Bruception its meant to be a full reset

backend/src/dal/user.ts
Miodec: > isnt it already doing that?

If you have await on each call that means they will be ran one after each other

YummyOreo: ok

Miodec: Also, lets move these calls. Instead of DALs calling other DALs, lets have the user controller call all the DAL files.

Miodec (changes requested):

backend/src/dal/user.ts
YummyOreo: > Also, lets move these calls. Instead of DALs calling other DALs, lets have the user controller call all the DAL files.

so do you mean have it call the `getXCollection` or a function that does what it is doing

Bruception: A function, ideally mongodb logic does not exist in controller layer.

YummyOreo: ok

backend/src/dal/user.ts
Bruception: How come?

YummyOreo: It can only be undefined or a specific type of object somewhere in MonkeyTypes.

Miodec:
Thanks

PR #3209, opened 2022-06-23
Converted badges to inventory

Miodec:
Converted the whole badges system into inventory which is more flexible and allows for different typed of "items" to be stored in the future (for example, achievements)

### DIFF
--- a/backend/src/api/controllers/result.ts
+++ b/backend/src/api/controllers/result.ts
@@ -330,6 +330,12 @@ export async function addResult(
     (process.env.MODE === "dev" || (user.timeTyping ?? 0) > 7200);
 
   if (dailyLeaderboard && validResultCriteria) {
+    //get the selected badge id
+    let badgeId: number | undefined;
+    if (user.inventory?.badges) {
+      badgeId = user.inventory.badges.find((b) => b.selected === true)?.id;
+    }
+
     incrementDailyLeaderboard(result.mode, result.mode2, result.language);
     dailyLeaderboardRank = await dailyLeaderboard.addResult(
       {
@@ -342,7 +348,7 @@ export async function addResult(
         uid,
         discordAvatar: user.discordAvatar,
         discordId: user.discordId,
-        badgeIds: user.badgeIds,
+        badgeId,
       },
       dailyLeaderboardsConfig
     );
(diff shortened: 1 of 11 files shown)


backend/src/dal/leaderboards.ts
```diff
@@ -90,7 +90,7 @@ export async function update(
             [str + ".name"]: "$name",
             [str + ".discordId"]: "$discordId",
             [str + ".discordAvatar"]: "$discordAvatar",
-            [str + ".badgeIds"]: "$badgeIds",
+            [str + ".badges"]: "$inventory.badges",
```
Bruception, line 93: We want to make sure the schemas are the same for leaderboards.

backend/src/types/types.d.ts
```diff
@@ -209,6 +218,8 @@ declare namespace MonkeyTypes {
     uid: string;
     name: string;
     rank: number;
+    badges?: Badge[];
```
Bruception, line 221: Why both of these fields?

frontend/src/ts/elements/profile.ts
```diff
@@ -38,15 +38,20 @@ export async function update(
...
-
-    details.find(".badges").empty().append(getHTMLById(firstBadgeId));
-    details.find(".allBadges").empty().append(miniBadges);
+  if (profile.inventory?.badges && !banned) {
+    let mainHtml = "";
+    let restHtml = "";
+
+    for (const badge of profile.inventory.badges) {
+      if (badge.selected === true) {
+        mainHtml = await getHTMLById(badge.id);
```
Bruception, line 47: Await?

backend/src/api/controllers/result.ts
```diff
@@ -330,6 +330,12 @@ export async function addResult(
     (process.env.MODE === "dev" || (user.timeTyping ?? 0) > 7200);
 
   if (dailyLeaderboard && validResultCriteria) {
+    //get the selected badge id
+    let badgeId: number | undefined;
+    if (user.inventory?.badges) {
```
Bruception, line 335: I think this can be refactored to just:

```ts
const badgeId = user.inventory?.badges?.find((b) => b.selected)?.id;
```

Miodec: done

Bruception (approved):
LGTM IIW

PR #3210, opened 2022-06-23
Inventory badge selection

Miodec:
Still need to update inventory correctly on the backend


backend/__tests__/utils/misc.spec.ts
```diff
@@ -71,6 +71,35 @@ describe("Misc Utils", () => {
     });
   });
 
+  it("identity", () => {
```
Bruception, line 74: Needed for coverage

PR #3216, opened 2022-06-23
Add Workspaces

Ferotiq:
### Description

Uses NPM Workspaces so that types can be shared among the frontend and backend.
This removes some scripts like `install-all` because running `npm install` installs for every workspace.
Also Updates some types and other files.

Overall a huge PR but most of the changes are just moving files

Closes #


Bruception:
I don't think we should do this. There really isnt a lot of overlap between the types (and some vary). Also changing the project structure again for minimal benefit is not worth it imo

Miodec:
Yeah, not sure if its worth it

PR #3221, opened 2022-06-24
Added Tokyonight theme

sanchay9:


Miodec:
Sorry, looks similar to other quotes available, including (but not only) onedark

sanchay9:
Hey, I know this looks very similar to onedark(and others?) but tokyonight is a rather popular colorscheme and I think monkeytype should have this already than the other similar looking ones.

PR #3234, opened 2022-06-26
View account page fix

aNOOBisTheGod:
Before this fix, this thing was not working for me(tested in 2 browsers). And now everything works fine!


Miodec (approved):
Thanks, looks like i missed that one

PR #3236, opened 2022-06-26
Custom theme edit fix

aNOOBisTheGod:
Before this fix you couldn't add themes with " ", and now you can. And also when you make invalid name for your theme, now you don't see popup that theme was updated successfully


Ferotiq (approved):
LGTM

PR #3251, opened 2022-06-27
Added Korean language

jeeiid:
### Description

Korean language(200 words) is added. 
  > file 'korean.json' added
  > '_groups.json' and '_list.json' is modified

and this is my first PR so I don't know how to write this.

commit message 's' and 'ㄴ' is just added words but I was lazy to write 'word added'.

Closes #


Ferotiq:
This has already been proposed numerous times (see #811) and we are unable to implement this due to how our input method is setup.

PR #3264, opened 2022-06-30
Add GitHub funbox mode to pull words from source code on GitHub

drewgingerich:
### Description

This PR adds a "GitHub" funbox mode that pulls source code from GitHub to provide words.

The code language of the files pulled from GitHub can be selected using the existing language selection, and falls back to JavaScript when the current language is not a code language. It searches for files in the top 25 most popular repositories that make use of the selected language.

Some code files have front-matter that the user must type through before getting to code. This could potentially be filtered out in the future for a more consistent experience.

The GitHub search API has has a rate limit of 10 requests a minute. This funbox makes two requests to the search API for each file, so it is low enough to be hit if someone refreshes the test a few times. Hitting the API limit hangs the website and requires a refresh, so I'd love some help figuring out a more graceful failure mode.

### DIFF
--- a/frontend/src/ts/test/github.ts
+++ b/frontend/src/ts/test/github.ts
@@ -1,0 +1,131 @@
+import * as Loader from "../elements/loader";
+import * as Misc from "../utils/misc";
+
+export class Section {
+  public title: string;
+  public author: string;
+  public words: string[];
+  constructor(title: string, author: string, words: string[]) {
+    this.title = title;
+    this.author = author;
+    this.words = words;
+  }
+}
+
+export async function getTLD(
+  languageGroup: MonkeyTypes.LanguageGroup
+): Promise<"en" | "es" | "fr" | "de" | "pt" | "it" | "nl"> {
+  // language group to tld
+  switch (languageGroup.name) {
+    case "english":
+      return "en";
+
+    case "spanish":
+      return "es";
+
+    case "french":
+      return "fr";
+
+    case "german":
+      return "de";
+
+    case "portuguese":
+      return "pt";
+
+    case "italian":
+      return "it";
+
+    case "dutch":
+      return "nl";
+
+    default:
+      return "en";
+  }
+}
+
+interface Post {
+  title: string;
+  author: string;
+  pageid: number;
+}
+
+interface SectionObject {
+  title: string;
+  author: string;
+}
+
+export async function getSection(language: string): Promise<Section> {
+  // console.log("Getting section");
+  Loader.show();
+
+  // get TLD for wikipedia according to language group
+  let urlTLD = "en";
+  const currentLanguageGroup = await Misc.findCurrentGroup(language);
+  if (currentLanguageGroup !== undefined) {
+    urlTLD = await getTLD(currentLanguageGroup);
+  }
+
+  const randomPostURL = `https://${urlTLD}.wikipedia.org/api/rest_v1/page/random/summary`;
+  const sectionObj: SectionObject = { title: "", author: "" };
+  const randomPostReq = await fetch(randomPostURL);
+  let pageid = 0;
+
+  if (randomPostReq.status == 200) {
+    const postObj: Post = await randomPostReq.json();
+    sectionObj.title = postObj.title;
+    sectionObj.author = postObj.author;
+    pageid = postObj.pageid;
+  }
+
+  return new Promise((res, rej) => {
+    if (randomPostReq.status != 200) {
+      Loader.hide();
+      rej(randomPostReq.status);
+    }
+
+    const sectionURL = `https://${urlTLD}.wikipedia.org/w/api.php?action=query&format=json&pageids=${pageid}&prop=extracts&exintro=true&origin=*`;
+
+    const sectionReq = new XMLHttpRequest();
+    sectionReq.onload = (): void => {
+      if (sectionReq.readyState == 4) {
+        if (sectionReq.status == 200) {
+          let sectionText: string = JSON.parse(sectionReq.responseText).query
+            .pages[pageid.toString()].extract;
+
+          // Converting to one paragraph
+          sectionText = sectionText.replace(/<\/p><p>+/g, " ");
+
+          // Convert HTML to text
+          sectionText = $("<div/>").html(sectionText).text();
+
+          // Remove reference links
+          sectionText = sectionText.replace(/\[\d+\]/gi, "");
+
+          // Remove invisible characters
+          sectionText = sectionText.replace(/[\u200B-\u200D\uFEFF]/g, "");
+
+          // Convert all whitespace to space
+          sectionText = sectionText.replace(/\s+/g, " ");
+
+          // Removing whitespace before and after text
+          sectionText = sectionText.trim();
+
+          const words = sectionText.split(" ");
+
+          const section = new Section(
+            sectionObj.title,
+            sectionObj.author,
+            words
+          );
+          Loader.hide();
+          res(section);
+        } else {
+          Loader.hide();
+          rej(sectionReq.status);
+        }
+      }
+    };
+    sectionReq.open("GET", sectionURL);
+    sectionReq.send();
+  });
+}
--- a/frontend/src/ts/test/github.ts
+++ b/frontend/src/ts/test/github.ts
@@ -123,6 +123,7 @@ function extractWords(fileContent: string): string[] {
 }
 
 async function sendGithubApiRequest(endpoint: string): Promise<unknown> {
+  const origin = "https://api.github.com";
   const url = origin + endpoint;
   const fileRequest = await fetch(url);
   if (!fileRequest.ok) {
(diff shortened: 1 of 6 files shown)


drewgingerich:
> Im pretty sure this is not worth merging with the amount of users the website gets and the potential of just one user triggering the rate limit

The rate limit is per user since the API calls are made from the browser!

Miodec:
I did one test and hit the rate limiter. Im not sure if this is good UX, regardless if we show a notification or not

drewgingerich:
Huh, I'm curious as to what happened. The code makes two requests per fetched file, so it should be able to fetch 5 code files per minute. Did you type through 5 code files in under a minute? It's all random, so there is a low change it pulls 5 files that are very small.

If we cache the repo, we'd be able to use only a single request to fetch a code file, so that'd be 9 code files per minute.

We could also implement a fallback to use existing code word generation while it waits for the api limit to refresh.

Thoughts?

Miodec:
> Is there a way to tell when a new test has started? Fetching the file content uses a different GitHub API than searching, and I believe it has a much higher limit. If we cache the repo for each test, we could potentially load a lot of files without reaching an api limit.

You can use `TestActive.get()`

Miodec:
> Huh, I'm curious as to what happened. What does the network tab say? The code makes two requests per fetched file, so it should be able to fetch 5 code files per minute. Did you type through 5 code files in under a minute? It's all random, so there is a chance it pulls 5 files that are very small.

I just loaded up the funbox, changed mode to words 10 and did a test. Loading up the funbox took some requests, changing mode was another few requests (I might have changed the mode twice). After finishing the test the next one didn't load.

Miodec:
Im gonna convert this to draft since its not ready

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,112 @@
...
+
+async function apiRequest(url: string): Promise<unknown> {
+  const fileRequest = await fetch(url);
+  if (!fileRequest.ok) {
+    throw Error(fileRequest.statusText);
+  }
+  return await fileRequest.json();
+}
+
+function getRandomItem<T>(list: T[], options?: { maxIndex?: number }): T {
```
Bruception, line 106: I think this function should be used to misc/utils, since this might be used in the future.

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,112 @@
...
+    .split(" ");
+
+  const section = { words };
+
+  Loader.hide();
+
+  return section;
+}
+
+async function apiRequest(url: string): Promise<unknown> {
```
Bruception, line 98: Same with this one.

drewgingerich: I understand the desire to generalize these things, but the way these function are written they're very specific to the needs of this code and I think they'd have to change a lot to be useful as general utils (especially the `apiRequest` function). I'm a big fan of writing things two or three times before trying to share functionality, because it always seems like things that look similar at first blush end up being different later. I suggest keeping them here and scheduling an audit for shared functionality at some point, so that there's a better understanding about what might be shared and what the requirements are.

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,177 @@
...
+
+async function sendGithubApiRequest(url: string): Promise<unknown> {
+  const fileRequest = await fetch(url);
+  if (!fileRequest.ok) {
+    throw Error(fileRequest.statusText);
+  }
+  return await fileRequest.json();
+}
+
+function getRandomItem<T>(list: T[]): T {
```
Bruception, line 131: I think this function is generic enough to move to utils/misc

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,177 @@
...
+  const randomIndex = Math.floor(Math.random() * list.length);
+  return list[randomIndex];
+}
+
+function cache<T extends unknown[], U>(
```
Bruception, line 136: Same with this one.

frontend/src/ts/test/github.ts
Bruception: Some of the functions seem generic enough, such that they could easily be re-used someplace else.
I don't think getting a random element from an array is specific to the needs of your code.

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,177 @@
...
+
+function extractWords(fileContent: string): string[] {
+  return fileContent
+    .replace(/[\u200B-\u200D\uFEFF]/g, "") // Remove invisible characters
+    .replace(/\s+/g, " ") // Convert all whitespace to space
+    .trim() // Removing whitespace before and after text
+    .split(" ");
+}
+
+async function sendGithubApiRequest(url: string): Promise<unknown> {
```
Bruception, line 123: I think if you are going to name the function this. The url should be built within the function.
Considering that you are reusing the base url: `https://api.github.com/`

drewgingerich:
Good news, I've added caching for the Search API requests. Meh news, the core GitHub API has a separate limit of 60 per day. This means that while a user can now use this funbox without interruptions due to the GitHub Search API limit, they will currently only be use this funbox until they hit this daily limit (it's like a solid 10-20 min of typing, I think).

We can stall reaching this daily limit by caching file content requests, as files occasionally get re-selected in subsequent tests. If we want a more robust experience there could be a fallback mode that uses only cached files or the existing code language word generation.

With good communication about what's going on, I think people can enjoy this as is even with the daily limit. I'll add error handling so it doesn't hang the site when the rate limit is reached.

Is there a standard way of communicating application state with the user? Something like the loader, but with text? I've looked through the elements briefly but figure someone might know off the top.

Bruception:
> Good news, I've added caching for the Search API requests. Meh news, the core GitHub API has a separate limit of 60 per day. This means that while a user can now use this funbox without interruptions due to the GitHub Search API limit, they will currently only be use this funbox until they hit this daily limit (it's like a solid 10-20 min of typing, I think).
> 
> We can stall reaching this daily limit by caching file content requests, as files occasionally get re-selected in subsequent tests. If we want a more robust experience there could be a fallback mode that uses only cached files or the existing code language word generation.
> 
> With good communication about what's going on, I think people can enjoy this as is even with the daily limit. I'm hoping other people feel the same.
> 
> I'll add error handling so it doesn't hang the site when the rate limit is reached.
> 
> Is there a standard way of communicating application state with the user? Something like the loader, but with text? I've looked through the elements briefly but figure someone might know off the top.

We usually use the Notifications API.

frontend/src/ts/test/github.ts
drewgingerich: I've moved the origin into the function like you suggest. I don't want to start making a whole client because the GitHub API surface is large and I'm only using a fraction of a sliver. If interfacing with GitHub becomes common I'd suggest searching for a library that does the whole API and has good support, though I don't how folks here feel about adding libraries.

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,132 @@
...
+    url: string;
+  }[];
+}
+
+interface FileResponse {
+  content: string;
+}
+
+export async function getSection(language: string): Promise<Section> {
+  Loader.show();
```
Bruception, line 59: I think we should move this UI logic outside of the GitHub related action and instead use it where this function is called.

frontend/src/ts/test/github.ts
```diff
@@ -0,0 +1,132 @@
...
+function extractWords(fileContent: string): string[] {
+  return fileContent
+    .replace(/[\u200B-\u200D\uFEFF]/g, "") // Remove invisible characters
+    .replace(/\s+/g, " ") // Convert all whitespace to space
+    .trim() // Removing whitespace before and after text
+    .split(" ");
+}
+
+async function sendGithubApiRequest(endpoint: string): Promise<unknown> {
+  const url = origin + endpoint;
```
Bruception, line 126: Where is origin defined?

drewgingerich: Great point, this wasn't caught by linting and I'm not sure why. It's properly defined now.

frontend/src/ts/test/github.ts
drewgingerich: I agree, but I'd like to push this into a later refactor if that's all right because the Wikipedia funbox does the same thing and I don't want to touch that in this PR if possible.

Bruception: Sounds good

drewgingerich:
I briefly looked into making this funbox fall back to using the regular word generation when it reaches the API limit or errors for any other reason, but it seems difficult to do this as it appears the logic for regular word generation is peppered throughout the [`test-logic.ts`](https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/ts/test/test-logic.ts) file and not easily accessible or reusable. I don't want to take on what might be a large refactor so I'm going to let this PR go.

PR #3268, opened 2022-06-30
Add 200 words under tamil.json and created tamil_1k.json

bala418:
### Description

Added 200 words for tamil.json, it earlier ahd only letters, and hence removed them. 
Created a new file tamil_1k.json which has 1000 words.
Updated the list and group
The meanings for these words could be found here: https://1000mostcommonwords.com/1000-most-common-tamil-words/


bala418:
Yup, just tested it. Typing works fine but the way the letters are displayed is not in the proper tamil format ie: a single letter is often represented as a combination of 2 letters, and here it is displayed separately.

Miodec:
Try adding `  "ligatures": true,` to the word list files

bala418:
That fixed the issue, and is rendered properly now.

PR #3286, opened 2022-07-01
Add version commit hash

Bruception:
### Description

Closes #

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -7,7 +7,7 @@
     "build": "tsc --build",
     "watch": "tsc --build --watch",
     "clean": "tsc --build --clean",
-    "start": "npm run build && node ./build/server.js",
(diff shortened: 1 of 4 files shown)


backend/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc --build",
     "watch": "tsc --build --watch",
     "clean": "tsc --build --clean",
-    "start": "npm run build && node ./build/server.js",
+    "start": "npm run build && export COMMIT_HASH=`git rev-parse --short HEAD` && node ./build/server.js",
```
Miodec, line 10: Can we move this to the build command? Will it still work?

Miodec (changes requested):

backend/package.json
Bruception: I don't think it makes sense for this to be in the build command. Since build command should just build (not set environment variable related to application run time).

PR #3289, opened 2022-07-02
added the canary matrix variant

j-fosstvedt:
### Description
As defined here:
https://github.com/Apsu/Canary#ortho


Miodec:
Number of keys in some of the rows is incorrect. Please fix.

PR #3297, opened 2022-07-03
remove time result when in time mode.

MysteriousLog6:
Fixes issue #3282

### Description

Fixes issue #3282 and hides the timer when in timed gamemode.

Closes #
 #3282


Miodec:
This is not the fix for this issue

PR #3306, opened 2022-07-05
Added two new quotes in English.

connorsdylan18:
### Description

Closes #


Miodec:
Sorry but I'm avoiding any quotes that refer to god or gods in any way. People get way too riled up about this topic (the people who believe in god and the people who don't) so I just avoid it entirely. Sorry again.

PR #3313, opened 2022-07-06
Changed two typos in norwegian 10k

j-fosstvedt:
Changed "levanger" to "Levanger", since it's the name of a place,vwhich means it should be capitalized. Also changed "insiden" to "innsiden", since insiden is the wrong spelling. It means inside, and "in" is translated to "inn" in norwegian. 
Might need to change 5K 1k for these too, they're quite common. But I'm on my phone now lol


j-fosstvedt:
Ok, then the second "innsiden" needs to get replaced with something else?

Miodec:
> Ok, then the second "innsiden" needs to get replaced with something else?

When its a duplicate? Either replace or just remove

j-fosstvedt:
It says that spanish 10k contains a dupe, weird

Ferotiq:
> It says that spanish 10k contains a dupe, weird

lol, a mistake on my pr from a little bit ago, opened a new one to fix it

j-fosstvedt:
oh you;re the one from the discord that inspired me to typo hunt lol

Miodec (approved):

PR #3318, opened 2022-07-07
Rewrite input system to use composition events

### LINKED ISSUE #2536: If the input contains ^^ it only shows the first ^, the second is still in the input buffer but input only progresses on next letter

- [X] I have cleared my cache

Yes

Yes

Yes

Yes

### Account information

_No response_

### Current Behavior

I find that in all my test setups entering ^^ in a row only progresses the cursor one place until the next character is typed.  Then all the "missing" ^ are used to progress the cursor, but not to color the letters progressed over.

### Expected Behavior

On every letter typed I would expect the cursor to progress one letter.

### Steps To Reproduce

1. start any test
2. type multiple ^
3. now you see the cursor not progress as expected
4. type any other letter
5. now you see the cursor progress over the right number of letters, but not color them

### Environment

- OS: OS X 12.2.1
- Browser: firefox 97.0.1, safari 15.3

### Anything else?

_No response_

omackenzie:
### Description
Changed from using trailingComposeChars to using compositionstart and compositionend events to check when a composing character was entered.
This doesn't fix any of the accuracy issues for languages like Japanese, but this should be helpful for any future attempts to fix this.
Closes #2536

I've tested this on pretty much every major browser and OS, but this could possibly break some input systems I'm either unaware of or don't have access to.
These changes should ideally be tested by a few people who use input systems with accents or any type of composing characters, in case anything breaks.

### DIFF
--- a/frontend/src/ts/controllers/input-controller.ts
+++ b/frontend/src/ts/controllers/input-controller.ts
@@ -24,6 +24,7 @@ import * as MonkeyPower from "../elements/monkey-power";
 import * as WeakSpot from "../test/weak-spot";
 import * as ActivePage from "../states/active-page";
 import * as TestActive from "../states/test-active";
+import * as CompositionState from "../states/composition";
 import * as TestInput from "../test/test-input";
 import * as TestWords from "../test/test-words";
 import { navigate } from "./route-controller";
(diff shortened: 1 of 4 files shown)


Miodec:
> This doesn't fix any of the accuracy issues for languages like Japanese, but this should be helpful for any future attempts to fix this.

Why is it not fixing accuracy calculation there? Cant we just only affect the accuracy on `compositionEnd` events?

PR #3321, opened 2022-07-08
add more Filipino quotes; changed Rizal quote to falsely attributed

janreggie:
### Description

- Add several more Filipino quotes
- Modify first source to be "Sa Aking mga Kabatà" by Rizal (_falsely attributed_ instead of _disputed_)
- Set dashes to be only one hyphen long (`-` instead of `--`)

PR #3322, opened 2022-07-08
spanish quotes: replaced en dash with `-`

janreggie:
### Description

Replaced en dash U+2013 at quote 103 to one hyphen `-` as the former is difficult to type.

### DIFF
--- a/frontend/static/quotes/spanish.json
+++ b/frontend/static/quotes/spanish.json
@@ -620,9 +620,9 @@
       "id": 102
     },
     {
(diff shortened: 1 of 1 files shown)


Miodec:
I think it would make more sense to just replace with a single `-`

janreggie:
That could work. Should I replace the `--` to `-` in #3321 as well?

Miodec:
> That could work. Should I replace the `--` to `-` in #3321 as well?

Yes

PR #3352, opened 2022-07-23
Changes in the profile

guijun13:
### Description

* In the 'Edit profile' popup:
  - Change the Github and Twitter input positions, to match the same position in the profile page.
  - Add static Github and Twitter URL before their inputs ("prevents" user insert the whole URL, causing nesting URL errors such as `https://github.com/https://github.com/<username>`)

| Before | After |
| --------------- | --------------- |
| image | image |

  - Idea: Add a regex similar to the website input to extract the username from Github and Twitter input URL. What do you guys think?

### DIFF
--- a/frontend/src/styles/popups.scss
+++ b/frontend/src/styles/popups.scss
@@ -1287,6 +1287,11 @@
     display: flex;
   }
 
+  .socialURL > p {
+    margin-block: 0.5rem;
+    margin-inline-end: 0.5rem;
+  }
+
   .badgeSelectionContainer {
     display: flex;
     flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


Miodec:
Looks like the github and twitter input fields are taller than usual. Please change. Also please add a bit of spacing between the static URL and the input field.

Bruception:
I think a simpler solution is to rename the input label to: "github username, twitter username"

Miodec:
> I think a simpler solution is to rename the input label to: "github username, twitter username"

True, but I guess with this PR there is absolute no second guessing

Miodec (approved):
Thanks

PR #3353, opened 2022-07-25
Added config for displaying the keymaps top row

bowski23:
### Description

Added a setting for displaying the top key row in the keymap. 
Three options are available: always show the top row, show it depending on the layout setting or never show the top row. 

I thought this change was necessary as the top row normally pollutes the keymap but for the funbox simon says mode it is necessary to see some other characters( like the ß in qwertz).

### DIFF
--- a/backend/src/api/schemas/config-schema.ts
+++ b/backend/src/api/schemas/config-schema.ts
@@ -64,6 +64,9 @@ const CONFIG_SCHEMA = joi.object({
     .string()
     .valid("lowercase", "uppercase", "blank", "dynamic"),
   keymapLayout: joi.string().valid(),
+  keymapShowTopRow: joi
+    .string()
+    .valid("always", "layout", "never"),
   fontFamily: joi.string(),
   smoothLineScroll: joi.boolean(),
   alwaysShowDecimalPlaces: joi.boolean(),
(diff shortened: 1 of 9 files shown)


bowski23:
just realized that the top row is split wrong in split styles

Miodec:
If we have an option for this then I would remove the "layout dependent" option and just have on or off.

bowski23:
I could imagine someone switching between keymap layouts and wanting to have a minimal look on one. But it's your call.
Can i fix the top row split in this PR or should i open a new one?

Miodec:
> I could imagine someone switching between keymap layouts and wanting to have a minimal look on one. But it's your call. Can I fix the top row split in this PR or should I open a new one?
> 
> Edit: I also wanted it to be minimally intrusive. So that if people have set up their layout, the new default config doesn't override their setting for the top row.

Alright, lets keep it then

Miodec:
Todo: 

 - [ ] Remove first key from top row when using staggered (to shift everything left)
image
- [ ] Hide the first key when using alice 
image
- [ ] Remove first key in matrix to shift everything to the left
image
- [ ] Hide first key in split 
image
- [ ] Remove first key in split matrix 
image

bowski23:
image
image
image
image
image

Miodec (approved):
Thanks

PR #3357, opened 2022-07-30
Added new german quotes

schuebel:
### Description

Added new quotes in german

### DIFF
--- a/frontend/static/quotes/german.json
+++ b/frontend/static/quotes/german.json
@@ -3126,6 +3126,48 @@
       "source": "Frank Herbert - Der Wüstenplanet",
       "length": 224,
       "id": 526
+    },
+    {
+      "text": "Selbst wenn ich wüsste, dass die Welt morgen in Stücke zerfällt, würde ich immer noch meinen Apfelbaum einpflanzen.",
+      "source": "Martin Luther",
+      "length": 115,
+      "id": 527
+    },
+    {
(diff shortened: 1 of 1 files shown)


Miodec:
Could you please translate the quotes for me

schuebel:
> 

Sure, there you go: 

Even if I knew that the world would fall to pieces tomorrow, I would still plant my apple tree.
- Martin Luther

Wonderfully sheltered by good powers, we confidently await what may come. God is with us in the evening and in the morning, and certainly every day.
- Dietrich Bonhoeffer (written while imprisoned by the nazis)

Who comes to the light? He who does the truth. What does that mean? Truth is to be done, it is not only to be thought or willed, but to be done.
- Dietrich Bonhoeffer (written while imprisoned by the nazis)

Our days may come to seventy years, or eighty, if our strength endures; yet the best of them are but trouble and sorrow, for they quickly pass, and we fly away. If only we knew the power of your anger! Your wrath is as great as the fear that is your due. Teach us to number our days, that we may gain a heart of wisdom.
- Psalm 90

The terrible thing is that beauty is both mysterious and terrible. God and the devil are fighting there, and the battlefield is the heart of man.
- Dostojewski

The mystery of human life is not in just living, but in the meaning of life.
- Dostojewski

We need routine and tradition. That is order. Order can become excessive, and that is not good, but chaos can swamp us so that we drown - and that is not good either. We need to stay on the straight and narrow path.
- Peterson

PR #3373, opened 2022-08-05
Fix/refactor xp

Bruception:
### Description

- Added better configuration for xp.
- Fixed daily bonus logic
- Refactored code

### DIFF
--- a/backend/src/api/controllers/result.ts
+++ b/backend/src/api/controllers/result.ts
@@ -8,7 +8,12 @@ import {
   recordAutoBanEvent,
 } from "../../dal/user";
 import * as PublicStatsDAL from "../../dal/public-stats";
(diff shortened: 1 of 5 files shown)


backend/src/api/controllers/result.ts
```diff
@@ -404,80 +404,89 @@ export async function addResult(
...
+      `Could not fetch last result: ${err}`
+    );
+  }
+
   if (lastResultTimestamp) {
-    const lastResultDay = new Date(lastResultTimestamp).getDay();
-    const today = new Date().getDay();
-    if (lastResultDay != today) {
-      dailyBonus = 1000;
+    const lastResultDay = getStartOfDayTimestamp(lastResultTimestamp);
```
Bruception, line 478: Old method is unreliable. What if someone takes a test on a Tuesday, and then takes a test a week later? No bonus :(

Miodec (approved):
Thanks

PR #3375, opened 2022-08-05
Added 5 new Albanian Quotes

deniscerri:
title

### DIFF
--- a/frontend/static/quotes/albanian.json
+++ b/frontend/static/quotes/albanian.json
@@ -138,6 +138,36 @@
       "source": "Amik Kasoruho",
       "length": 606,
       "id": 20
+    },
+    {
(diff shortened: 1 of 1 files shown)


Miodec:
Please translate them so we can verify content

deniscerri:
1. Perhaps the biggest mistake he made in the past was believing that love was about finding the right person. Actually, love is about becoming the right person. Do not search for the person with whom you will spend your life. Be the person you spend your life with

2. I wish you to find love, but more than that I wish you to be strong enough to walk away from where love is missing

3. If you are going to live a happy life, associate it with a goal, not with people or objects

4. You can talk about someone for years, day after day, and still not feel the same as standing in front of someone, without saying a word, and still feeling for that person with all your heart. You feel like you've known that person for a long time... Connections are made with the heart, not the tongue

5. Kindness is a choice. Happiness is a choice. Optimism is a choice. Goodness is a choice. Forgiveness is a choice. Respect is a choice. Whatever you choose defines you, so choose wisely

PR #3381, opened 2022-08-07
Add version indicator option 1

lukew3:
### Description
This pr removes the banner that shows when a new version is released and instead shows a 5px diameter circle at the upper right corner of the version number to alert users of a new version. Don't think I can test properly on localhost but it's pretty straightforward. Remove the hidden class on line 83 of bottom.html to see the indicator

Other edits:
* Removed a block of code that wouldn't be executed since it was checking if localhost after the code executing it ensures that it is not localhost
* Fixes a flaw in `CONTRIBUTING_ADVANCED.md` relating to the code copied to  `firebase-config.ts` containing more than is expected. 

What the indicator looks like:
Screenshot from 2022-08-07 01-53-32

### DIFF
--- a/CONTRIBUTING_ADVANCED.md
+++ b/CONTRIBUTING_ADVANCED.md
@@ -52,14 +52,13 @@ Alternatively, if you use `nvm` then you can run `nvm install` and `nvm use` (yo
 
    - Run `firebase projects:list` to find your firebase project id.
 
-1. Within the `frontend/src/ts/constants` directory, duplicate `firebase-config-example.ts`, rename it to `firebase-config.ts` and paste in your firebase config
+1. Within the `frontend/src/ts/constants` directory, duplicate `firebase-config-example.ts`, rename it to `firebase-config.ts` and paste in your firebase config object
 
    - To find it, go to the Firebase console
    - Navigate to `Project Settings > General > Your apps`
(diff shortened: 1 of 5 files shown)


Ferotiq:
IMO, I think you should make the indication more visible, maybe like a little tag colored with the main color that says "New Version" placed beside the version number, or change the background color of the version number to the main color, something like that.

CONTRIBUTING_ADVANCED.md
```diff
@@ -59,6 +59,7 @@ Alternatively, if you use `nvm` then you can run `nvm install` and `nvm use` (yo
      - If there are no apps in your project, create a new web app
    - In the `SDK setup and configuration` section, select `npm`
    - The Firebase config will be visible below
+   - After pasting, remove everything above and below the object `firebaseConfig` and add export to the front of the line containing `const firebaseConfig`
```
Ferotiq, line 62: This isn't necessary because we have `firebase-config-example.ts` that has the correct format, all they need to do is paste the object overtop.

Also unrelated.

lukew3: True. I just missed it when I was speeding through the setup. I removed that line I added and specified to copy the object in the main instruction.

PR #3385, opened 2022-08-07
Implement streaks

typermonkeyuser:
Hi,

I've added the concept of streaks.
I updated the profile page to show the user's current streak.
image

The streak will also affect the XP bonus, the larger the streak, the higher bonus.

I have updated the XP modifier to add the log10 calculation of the streak.
That means that a streak below 10 days will increase the modifier with 0.x, a streak below 100 will increase the modifier with 1.x, a streak below 1000 will increase the modifier with 2.x etc.


Ferotiq:
I think the implementation could be better. Installing that package and calculating the streak on both the frontend and backend seems a bit wasteful.

Instead I propose I new user endpoint on the backend that just gets the streak of the user.

backend/package.json
```diff
@@ -21,6 +21,7 @@
     "chalk": "4.1.2",
     "cors": "2.8.5",
     "cron": "1.8.2",
+    "date-streaks": "^1.2.1",
```
Ferotiq, line 24: save exact

backend/src/api/controllers/result.ts
```diff
@@ -34,6 +34,7 @@ import * as George from "../../tasks/george";
 import { getDailyLeaderboard } from "../../utils/daily-leaderboards";
 import AutoRoleList from "../../constants/auto-roles";
 import * as UserDAL from "../../dal/user";
+import { summary } from 'date-streaks';
```
Ferotiq, line 37: you need to run the pretty-fix script to fix all prettier related issues

backend/src/api/controllers/result.ts
```diff
@@ -468,6 +469,20 @@ async function calculateXp(
       modifier += 0.1;
     }
   }
+  
+  // Increment modifier according to the user's current streak.
+  // Modifier is incremented with the log10 calculation of the current streak.
+  // That means that streaks greater than 10 days will increase modifier with 1.x,
+  // streaks greater than 100 will increase modifier with 2.x, greater than 1000 with 3.x, etc.
+  const allResults = await ResultDAL.getResults(uid);
```
Ferotiq, line 477: Getting all results might be a bit expensive and we could probably find a way to make this better.

Ferotiq: Also `getResults` returns only the last 1000 results and so the streak will be limited to some degree depending on how many tests a day the user does.

fasttyperdog:
@Ferotiq I have moved the logic to the backend.
Maybe someone with more MongoDB knowledge than me could come up with a smart query to calculate the streak?

fasttyperdog:
@Ferotiq I now store the streak on the user.
The expensive calculation is only necessary once, after that the streak is calculated based on the last result.

fasttyperdog:
@Ferotiq Would you like to take an additional look, after my latest changes?

Miodec:
I think it's fine to just avoid the init calculation and ignore any streaks before this feature is introduced

fasttyperdog:
@Miodec Thanks, I have removed the code performing the initial calculation.

fasttyperdog:
@Miodec Would you like to re-activate the tests?

fasttyperdog:
@Miodec @Ferotiq Will one of you take an additional look when you have time? :)

Miodec:
Why close?

typermonkeyuser:
Mistake

Miodec:
I don't like how you're getting the whole user doc everytime the profile is viewed. Plus it looks like if you view someone else's profile, their streak data will be put into the local db snapshot. 

You should get the streak data from the profile endpoint, and update local db snapshot with the data returned from adding result.

backend/__tests__/api/controllers/user.spec.ts
```diff
@@ -52,5 +54,12 @@ describe("user controller test", () => {
         })
         .expect(409);
     });
+
```
Bruception, line 57: This is the wrong file for this test. There is a DAL test directory.

backend/src/api/controllers/user.ts
```diff
@@ -487,3 +488,40 @@ export async function updateProfile(
 
   return new MonkeyResponse("Profile updated");
 }
+
+function isToday(someDate): boolean {
+  const today = new Date();
+  return (
+    someDate.getDate() == today.getDate() &&
```
Bruception, line 495: Use strict equality, also move these functions to utils/misc

Miodec:
Waiting for you to address the issues raised by me and Bruce

rizwanmustafa:
@typermonkeyuser any updates?

backend/__tests__/api/controllers/user.spec.ts
```diff
@@ -52,5 +54,12 @@ describe("user controller test", () => {
...
+    it("updateStreak should update streak", async () => {
```
Bruception, line 58: This test should be moved to the dal spec

backend/src/api/controllers/user.ts
```diff
@@ -4,13 +4,19 @@ import MonkeyError from "../../utils/error";
...
+  isYesterday,
+  sanitizeString,
+} from "../../utils/misc";
 import * as George from "../../tasks/george";
 import admin from "firebase-admin";
 import { deleteAllApeKeys } from "../../dal/ape-keys";
 import { deleteAllPresets } from "../../dal/preset";
 import { deleteAll as deleteAllResults } from "../../dal/result";
 import { deleteConfig } from "../../dal/config";
+import UserStreak = MonkeyTypes.UserStreak;
```
Bruception, line 19: I don't think we have to import this

backend/__tests__/api/controllers/user.spec.ts
```diff
@@ -1,5 +1,7 @@
 import request from "supertest";
 import app from "../../../src/app";
+import * as UserDAL from "../../../src/dal/user";
```
Bruception, line 3: Unused importa

backend/__tests__/dal/user.spec.ts
```diff
@@ -600,4 +605,25 @@ describe("UserDal", () => {
       },
     ]);
   });
+
+  it("updateStreak should update streak", async () => {
+    await UserDAL.addUser("testStack", "test email", "TestID");
+
+    const mockDate = new Date(1659749879000);
+    jest
```
Bruception, line 613: Let’s restore this Mock

Bruception (approved):
Lgtm

PR #3388, opened 2022-08-07
Add 15 new Dutch quotes

Max-Verbinnen:
### Description

This PR adds 15 new Dutch quotes :)

### DIFF
--- a/frontend/static/quotes/dutch.json
+++ b/frontend/static/quotes/dutch.json
@@ -1,22 +1,10 @@
 {
   "language": "dutch",
   "groups": [
-    [
-      0,
-      100
-    ],
-    [
-      101,
-      300
-    ],
-    [
-      301,
-      600
-    ],
-    [
-      601,
-      9999
-    ]
+    [0, 100],
+    [101, 300],
+    [301, 600],
+    [601, 9999]
   ],
   "quotes": [
     {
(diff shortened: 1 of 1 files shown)


Max-Verbinnen:
Here are the translations:

- Everything beautiful hurts a little. Art hurts too. Why do people cry at music or in front of a painting? Because the feelings of lust are only at their peak when it also hurts. (Rik Torfs)
- We remember from some people, years after their death, a few moments that seemed unimportant at the time. A casual statement that was not meant for eternity, an unexpected and sometimes one-time moment of connection. Laterally, the essence intrudes on life. (Rik Torfs)
- I often read that, following Karl Popper, we should not be tolerant of intolerant people if we want to keep society tolerant. Nice. But if we apply this principle, the first question must be: are we absolutely sure that we ourselves are tolerant people? (Rik Torfs)
- It doesn't matter how beautiful your theory is, it doesn't matter how smart you are. If it doesn't agree with experiment, it's wrong. (Richard Feynman)
- Poets say science takes away from the beauty of the stars - mere globs of gas atoms. I, too, can see the stars on a desert night, and feel them. But do I see less or more? (Richard Feynman)
- To be ignorant of what occurred before you were born is to remain always a child. For what is the worth of human life, unless it is woven into the life of our ancestors by the records of history? (Cicero)
- It is more important to know where you are going than to get there quickly. Do not mistake activity for achievement. (Isocrates)
- What is morally wrong can never be advantageous, even when it enables you to make some gain that you believe to be to your advantage. The mere act of believing that some wrongful course of action constitutes an advantage is pernicious. (Cicero)
- A rationalist is simply someone for whom it is more important to learn than to be proved right; someone who is willing to learn from others - not by simply taking over another's opinions, but by gladly allowing others to criticize his ideas and by gladly criticizing the ideas of others. (Popper)
- The war of ideas is a Greek invention. It is one of the most important inventions ever made. Indeed, the possibility of fighting with with words and ideas instead of fighting with swords is the very basis of our civilization, and especially of all its legal and parliamentary institutions. (Popper)
- Without training, they lacked knowledge. Without knowledge, they lacked confidence. Without confidence, they lacked victory. (Julius Caesar)
- When I was 5 years old, my mom always told me that hap­pi­ness was the key to life. When I went to school, they asked me what I wanted to be when I grew up. I wrote down “happy”. They told me I didn’t under­stand the assign­ment and I told them they didn’t under­stand life. (Julius Caesar)
- Freedom is the freedom to say that two plus two make four. If that is granted, all else follows. (George Orwell)
- Reflect upon your present blessings – of which every man has many – not on your past misfortunes, of which all men have some. (Charles Dickens)
- The caress of tenderness is the movement of the hand, in which the distinction between rite and ordinary act has been removed. (Ton Lemaire)

erum-hassan:
*Intrested*

PR #3392, opened 2022-08-07
Added version indicator option 2

lukew3:
### Description
Same thing as #3381 but replaced 5px circle indicator with a rounded text box.

Screenshot from 2022-08-07 15-58-54


lukew3:
@Ferotiq What do you think about this one?

Ferotiq:
Looks good, I'll ask for @Miodec's opinion though.

PR #3393, opened 2022-08-07
Added version indicator option 3

lukew3:
### Description
Version of #3381 
Clicking on the indicator will close it without opening the list of versions. If you `git reset HEAD^` you can see what it's like when an X shows on hover, but I think it's better without showing X on hover
Screenshot from 2022-08-07 17-38-13
.


Miodec:
Thanks

PR #3406, opened 2022-08-12
adding korean to languages

neezacoto:
### Description

Adding Korean: this includes korean (500 words), korean_1k, korean_2k, korean_3k, korean_4k, korean_5k, korean_6k (5890 words)


neezacoto:
realized I had a lot of duplicates, I've now condensed it down to korean (500), korean_1k, and korean_4k (4201)

Miodec:
Have you tested these lists? Korean is having some problems with the accuracy calculation because of the complicated input method.

neezacoto:
> Have you tested these lists? Korean is having some problems with the accuracy calculation because of the complicated input method.

Thanks for pointing that out, I think I'm going to have to come up with a different way to measure for accuracy. It might have to be measured by blocks of words instead, because if I type if the word is 하다 and I type 하 and then type ㄷ it creates 핟. I'm also thinking of adding some sort of visual to show what the user is typing, since that is sort of long with this input style.

neezacoto:
going to work on it some more before reopening

PR #3433, opened 2022-08-24
Change the error colour in the blue dolphin theme to make it more readable

saieeshb:
### Description

For the blue dolphin theme, when the colourful mode was used, errors were really hard to see.

image

Colorful mode is on in the above image.

In the above image, the "c" is wrongly typed but it barely looks like it. The reason being the difference between the error color and the main color is too little.  Therefore I suggest a change in the colourful error colour.

image

Main colour vs **Previous** colourful error colour

image

Main colour vs **New** colourful error colour

----------------------------------------------------------------------------------
Here's how it looks:

image

**Before**

image

**After**

Original colourful error : #ffbde6
New colourful error : #d1a5fd

I tried to keep it with the pastel-like colours that @vimiomori used.

### DIFF
--- a/frontend/static/themes/blue_dolphin.css
+++ b/frontend/static/themes/blue_dolphin.css
@@ -7,6 +7,6 @@
   --text-color: #82eaff;
   --error-color: #ffbde6;
   --error-extra-color: #ff8188;
-  --colorful-error-color: #ffbde6;
+  --colorful-error-color: #d1a5fd;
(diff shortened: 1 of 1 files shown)


rizwanmustafa:
I really appreciate the effort you put into the documentation of PR.

Miodec (commented):
Nice, thanks. Also, very good docs

saieeshb:
Thank you!

PR #3435, opened 2022-08-24
refactoring input controller (strict equality & TS typing)

neezacoto:
## small refactor

- Adding TS typing
- Changed some equality from loose to strict. I didn't change equalities that had to do with `Config`,
  like `Config.mode == "zen"`, because I wasn't sure if it was intentional or not.
- Other small changes like the use of `startsWith` and a spelling error.


Miodec (approved):
Thanks

PR #3461, opened 2022-08-28
Commandline to previous page on option selection

neezacoto:
### Changes
- Commandline now goes back to previous page on option selection

I thought it would be a better user experience for use cases like testing out your keyboard layout or wanting to add multiple commands at once. Personally when trying to do commands I gave up, because every time I selected an option it hid the menu. I feel like this change would encourage the use of the commandline more.

back


Miodec:
I dont think i agree with this change. All the command lines that i was basing this feature on (vscode command pallette, chrome devtools, others) close the menu on command execution.

PR #3456, opened 2022-08-27
Korean Keymap

neezacoto:
### Description
- added Korean keymap
- fixed some logic that wasn't accounting for korean character length
- fixed Korean input visual so that it appears above the keymap
- Korean next support
- Korean react support

demo

#### react demo 

react

PR #3474, opened 2022-08-30
fixing wpm and correctedHistory for Korean

neezacoto:
other bug fixes and keymaping is in this draft pr: [Korean Keymap #3456](https://github.com/monkeytypegame/monkeytype/pull/3456)
### Fixing
- fixed wpm calculation in ``test-inputs.ts`` and ``test-stats``
- fixed Korean word compiling counting as an input in ``input-controller``.
- fixed corrected history in ``input-controller`` 
- fixed burst calculation ``test-stats``
- now handles Korean characters regardless of language.

demokofix

Handle Korean in any language
both

### DIFF
--- a/frontend/src/ts/controllers/input-controller.ts
+++ b/frontend/src/ts/controllers/input-controller.ts
@@ -32,6 +32,8 @@ import { navigate } from "./route-controller";
 
 let dontInsertSpace = false;
 let correctShiftUsed = true;
+let isKoCompiling = false;
+let isBackspace: boolean;
 
 const wordsInput = document.getElementById("wordsInput") as HTMLInputElement;
 const koInputVisual = document.getElementById("koInputVisual") as HTMLElement;
@@ -291,7 +293,8 @@ function isCharCorrect(char: string, charIndex: number): boolean {
     return true;
   }
 
-  if (Config.language.startsWith("korean")) {
+  //Checking for Korean char
+  if (TestInput.input.getKoreanStatus()) {
     //disassembles Korean current Test word to check against char Input
     const koWordArray: string[] = Hangul.disassemble(
       TestWords.words.getCurrent()
@@ -366,7 +369,7 @@ function handleChar(
   if (TestUI.resultCalculating || TestUI.resultVisible) {
     return;
   }
-
+  const isCharKorean: boolean = TestInput.input.getKoreanStatus();
   if (char === "…") {
     for (let i = 0; i < 3; i++) {
       handleChar(".", charIndex + i);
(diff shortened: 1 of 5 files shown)


frontend/src/ts/test/test-ui.ts
```diff
@@ -1041,9 +1048,3 @@ $(document.body).on("click", "#showWordHistoryButton", () => {
 $("#wordsWrapper").on("click", () => {
   focusWords();
 });
-
-$(document).on("keypress", () => {
-  if (resultVisible) {
-    skipXpBreakdown();
-  }
-});
```
Miodec, line 1049: Bring this back

frontend/src/ts/test/test-ui.ts
```diff
@@ -15,7 +15,6 @@ import * as ConfigEvent from "../observables/config-event";
 import * as Hangul from "hangul-js";
 import format from "date-fns/format";
 import { Auth } from "../firebase";
-import { skipXpBreakdown } from "../elements/account-button";
```
Miodec, line 18: Bring this back

Miodec (changes requested):

frontend/src/ts/controllers/input-controller.ts
```diff
@@ -911,6 +923,12 @@ $("#wordsInput").keyup((event) => {
   Monkey.stop();
 });
 
+$("#wordsInput").on("keydown", (event) => {
+  if (Config.language.startsWith("korean")) {
+    isBackspace = event.key === "Backspace" || event.key === "delete";
+  }
+});
+
 $("#wordsInput").on("beforeinput", (event) => {
```
Miodec, line 932: Instead of adding a new event listener you could integrate into the one in `input-controller.ts` line`792` or `817`

Miodec (changes requested):

frontend/src/ts/controllers/input-controller.ts
```diff
@@ -291,7 +293,12 @@ function isCharCorrect(char: string, charIndex: number): boolean {
     return true;
   }
 
-  if (Config.language.startsWith("korean")) {
+  //Checking for Korean char
+  if (
+    char.match(
+      /[\uac00-\ud7af]|[\u1100-\u11ff]|[\u3130-\u318f]|[\ua960-\ua97f]|[\ud7b0-\ud7ff]/g
+    )
+  ) {
```
Miodec, line 301: Why switch to regex? Whats the performance looking like when doing this on every keypress?

Miodec:
Overall it looks good, all though it touches the wpm calculation in a lot of places. I think it would be a good idea to get some sort of macro program and verify that the wpm before and after this change is unchanged based on the same exact input

frontend/src/ts/controllers/input-controller.ts
neezacoto: I did this because the isHangul function wasn't working. I could try and see why/play around with it. Is there a definitive way to check performance with mk?

neezacoto:
performance console print was added, [heres the control](https://github.com/monkeytypegame/monkeytype/pull/3496)

neezacoto:
[this pr verifies that wpm for Latin isn't being altered ](https://github.com/monkeytypegame/monkeytype/pull/3498)

PR #3499, opened 2022-09-03
Added hinglish under hindi (not as separate language)

KParthSingh:
### Description

Closes #


rizwanmustafa:
The `bcp47` field's value seems incorrect. This is not Indian english but Hindi written in Latin script.

rizwanmustafa:
@Miodec there is no proper bcp47 value for Roman Hindi (Hindi written in Latin script). So I am thinking of removing the attribute from the json file. What do you say?

KParthSingh:
> @Miodec there is no proper bcp47 value for Roman Hindi (Hindi written in Latin script). So I am thinking of removing the attribute from the json file. What do you say?

It's your choice 🤷🏻‍♂️

Miodec:
> @Miodec there is no proper bcp47 value for Roman Hindi (Hindi written in Latin script). So I am thinking of removing the attribute from the json file. What do you say?

If there is no proper value then remove it

Miodec (approved):

Issue #3493, opened 2022-09-02
Improve language checking


monkeytypegeorge:
Currently we either use `Config.language.replace(/_\d*k$/g, "")` or `Config.language.split("_")[0]` to get the langauge from a file name (english from english_1k and so on)

All of these instances should be replaced with `Config.language.startsWith(language)`

neezacoto:
I can do this

PR #3502, opened 2022-09-04
language checks use startsWith()

neezacoto:
### Description

Closes # [Improve language checking #3493](https://github.com/monkeytypegame/monkeytype/issues/3493)


Miodec:
I am 100% sure there are more cases than just three

neezacoto:
startsWith() returns a boolean. There are other cases where it's used, but it's for display.

Miodec:
Ahhh, youre right - my bad

PR #3503, opened 2022-09-04
Added some russian quotes, standardized the sequence of fields in json

Faurazeko:
I added 11 russian quotes and standardized the sequence of fields in json file (the first 10 quotes had a different field sequence, but now they have the same sequence as the other quotes)


Miodec:
Please translate all quotes so i can verify the content of them

Faurazeko:
> Please translate all quotes so i can verify the content of them

id 1026: Joanne Rowling and the Half-Blood Prince. Greatness inspires envy. Envy engenders spite. Spite spawns lies. You must know this, Dumbledore.
id 1027 - Joanne Rowling and the Order of the Phoenix. The things we lose have a way of coming back to us in the end. If not always in the way we expect.
id 1028 - Harry Potter and the Deathly Hallows. If you have to ask, you’ll never know. If you know, you need only ask.
id 1029 - "April" Movie (2001 year). A samurai has no goal, only a path.
id 1030 - Harry Potter and the Goblet of Fire. Curiosity is not a sin, Harry, but you have to be a little more patient.
id 1031 -  Joanne Rowling and the Philosopher's stone. The fear of the name increases the fear of the one who wears it.
id 1032 - Joanne Rowling and the Order of the Phoenix. Do not forget that all the great wizards in history at one time were just like us - schoolchildren. If they can do it, then we can do it too.
id 1033 - Joanne Rowling and the Philosopher's stone. It takes great courage to stand up to your enemies... but it's even harder to stand up to your friends.
id 1034 - Harry Potter and the prisoner of Azkaban. It turns out that more than anything in the world you are afraid of fear. It is commendable!
id 1035 - Joanne Rowling and the Order of the Phoenix. Nothing is impossible in the world - it's just a matter of whether you have the courage.
id 1036 - Harry Potter and the prisoner of Azkaban. Happiness can be found even in dark times, if you remember to turn to the light.

PR #3520, opened 2022-09-06
Add Georgian quotes

avetisovi:
Added quotes for Georgian language, which hadn't quotes before.


Miodec:
Could you translate them so we can verify the content?

avetisovi:
> Could you translate them so we can verify the content?

How can I send translated version?

Miodec:
You can just comment here

avetisovi:
1. Quite often good things have hurtful consequences. There are instances of men who have been ruined by their money or killed by their courage. (Aristotle)

2. Ancient laws remain in force long after the people have the power to change them. (Aristotle)

3. In general, people force what they are dealing with. (Aristotle)

4. Sometimes it's better to say than not to say, Sometimes speach can get you hurt. (The Knight in the Panther's Skin)

5. To the future or to the past, to a time when thought is free, when men are different from one another and do not live alone— to a time when truth exists and what is done cannot be undone. (1984)

6. Power means power over people, of course over their bodies too, but above all, over their minds. (1984)

7. You are right, my Luarsab, you have not killel a man, you have not stolen from a man, in a word - you have not done what you shouldn't have done - that's a good thing: it's better than worse. But now I have to ask you this: Have you done what you should have done? (Is he a human?!)

8. True courage is about knowing not when to take a life, but when to spare one (The Hobbit)

9. One day it started raining, and it didn't quit for four months. We been through every kind of rain there is. Little bitty stingin' rain, and big ol' fat rain, rain that flew in sideways, and sometimes rain even seemed to come straight up from underneath. (Forrest Gump)

10. Most people say that it is the intellect which makes a great scientist. They are wrong: it is character. (Albert Einstein)

11. The day that man allows true love to appear, those things which are well made will fall into cofusion and will overturn everything we believe to be right and true. And the world will become real only when a person learns to love. Until then, we can live with only one faith - faith, as if we understood something about love, but we still haven't learned to appreciate it. (Dante Alighieri)

12. In this game that we're playing, we can't win. Some kinds of failure are better than other kinds. (1984)

13. He who controls the past controls the future. He who controls the present controls the past. (1984)

14. Belief is a matter of conscience, of conscience, and here everyone is free and untouchable (Ilia Chavchavadze)

15. All happy families resemble one another, each unhappy family is unhappy in its own way. (Leo Tolstoy)

Miodec:
Alright, thanks

PR #3525, opened 2022-09-08
Add xp reward overrides

Bruception:
### Description

Adds xp reward overrides for specific ranks
Screen Shot 2022-09-07 at 8 59 49 PM

### DIFF
--- a/backend/__tests__/utils/daily-leaderboards.spec.ts
+++ b/backend/__tests__/utils/daily-leaderboards.spec.ts
@@ -23,7 +23,7 @@ const dailyLeaderboardsConfig = {
   topResultsToAnnounce: 3,
   maxXpReward: 0,
   minXpReward: 0,
-  xpRewardOverrides: [],
+  xpRewardBrackets: [],
 };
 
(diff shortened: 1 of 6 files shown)


backend/src/types/types.d.ts
```diff
@@ -79,6 +79,10 @@ declare namespace MonkeyTypes {
       topResultsToAnnounce: number;
       maxXpReward: number;
       minXpReward: number;
+      xpRewardOverrides: {
+        rank: number;
```
Bruception, line 83: If we see ourselves needing to do ranges of ranks. We will use a regex pattern instead.

backend/__tests__/utils/daily-leaderboards.spec.ts
```diff
@@ -23,6 +23,7 @@ const dailyLeaderboardsConfig = {
   topResultsToAnnounce: 3,
   maxXpReward: 0,
   minXpReward: 0,
+  xpRewardOverrides: [],
```
Bruception, line 26: This needs to be updated

PR #3553, opened 2022-09-15
Change displayed modifier key based on platform

cubedhuang:
### Description

This PR changes the bottom key tip to just be `ctrl + shift + p` or `cmd + shift + p` based on the detected platform, rather than `ctrl/cmd + shift + p`.

Before:
image

After (Windows):
image


rizwanmustafa (changes requested):
This function is deprecated. Please use an alternative.

image

rizwanmustafa (approved):

PR #3567, opened 2022-09-20
Update indonesian.json

RFFadhilR:
### Description

Closes #


Miodec:
These seem not "quotable" at all. Please use correct punctuation and content from books, movies, famous people or popular sayings.

PR #3585, opened 2022-09-22
Get profile by name

Miodec:


backend/src/api/controllers/user.ts
```diff
@@ -438,7 +438,14 @@ export async function removeFavoriteQuote(
 export async function getProfile(
   req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
-  const { uid } = req.params;
+  const { uidOrName } = req.params;
+
+  const hasUidInQuery =
```
Bruception, line 443: Just:

const { uid } = req.query;

it will be undefined if it’s not there

backend/src/dal/user.ts
```diff
@@ -175,6 +165,29 @@ export async function getUser(
   return user;
 }
 
+async function findByName(name: string): Promise<MonkeyTypes.User | undefined> {
```
Bruception, line 168: i think it would be better if this function returns the full list of users rather than [0]

backend/src/api/routes/users.ts
```diff
@@ -430,7 +430,10 @@ router.get(
   withApeRateLimiter(RateLimit.userProfileGet),
   validateRequest({
     params: {
-      uid: joi.string().required(),
+      uidOrName: joi.string().required(),
+    },
+    query: {
+      uid: joi.string().allow(""),
```
Bruception, line 436: Nit:

isUid

backend/src/dal/user.ts
Miodec: As mentioned before, i wanna keep this as a lookup initially - after that we can add proper search with results and stuff

backend/__tests__/api/controllers/user.spec.ts
```diff
@@ -17,8 +18,48 @@ describe("user controller test", () => {
...
 
+      // getCachedConfiguration = jest.fn().mockResolvedValue({
+      //   users: {
+      //     signUp: true,
+      //   },
+      // });
+
+      jest.spyOn(Configuration, "getCachedConfiguration").mockResolvedValue({
+        //if stuff breaks this might be the reason
+        users: {
```
Bruception, line 32: We had to specify this whole config? :O

backend/__tests__/api/controllers/user.spec.ts
```diff
@@ -17,8 +18,48 @@ describe("user controller test", () => {
         name: "NewUser",
         uid: "123456789",
         email: "newuser@mail.com",
+        captcha: "captcha",
       };
 
+      // getCachedConfiguration = jest.fn().mockResolvedValue({
```
Bruception, line 24: Hmm 🤨

backend/src/api/controllers/user.ts
```diff
@@ -438,7 +438,14 @@ export async function removeFavoriteQuote(
 export async function getProfile(
   req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
-  const { uid } = req.params;
+  const { uidOrName } = req.params;
+
+  const { isUid } = req.query;
```
Bruception, line 443: Yay nice

backend/__tests__/api/controllers/user.spec.ts
Miodec: Just user

Bruception (commented):
Overall looks good to me just some comments

Bruception (approved):
LGTM IIW

PR #3626, opened 2022-10-04
De-emphasize unreached words in burst heatmap

sartak:
### Description

I noticed the "unreached" words at the end of the input burst heatmap use the same styling as the median speed, which is misleading. In this test I never reached "under first":

スクリーンショット 2022-10-04 13 11 50

These words don't have a `.heatmap#` class so they get the default `.word` styling.

This PR de-emphasizes those words:

スクリーンショット 2022-10-04 13 11 38

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -540,6 +540,9 @@
         letter.incorrect.extra {
           color: var(--error-extra-color);
         }
+        &.unreached letter {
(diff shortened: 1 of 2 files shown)


Miodec:
Nice, thanks

Miodec:
> Sweet, thanks for the quick merge. Any chance I could get that contributor badge :D my monkeytype username is sartak

Sure thing

PR #3647, opened 2022-10-13
Fixed db snapshot type

Miodec:
Dbsnapshot for some reason didnt include `undefined` in its type definition (it will be undefined if user is not signed into an account). This pr fixes that.

It should be fine to just return / optional chain everything since in *most* cases the code in question would not even execute (it would probably be stopped by `Auth.currentUser` check) but an extra guard statement is always good. Posting as pr in case anyone thinks things might break somewhere.

### DIFF
--- a/frontend/src/ts/account/all-time-stats.ts
+++ b/frontend/src/ts/account/all-time-stats.ts
@@ -9,7 +9,7 @@ export function clear(): void {
 
 export function update(): void {
   const snapshot = DB.getSnapshot();
-
+  if (!snapshot) return;
   if (snapshot.typingStats !== undefined) {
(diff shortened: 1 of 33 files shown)


Bruception:
I think we might as well define a more rigid interface for the snapshot.

ie:

```ts
const filterPresets = db.getFilterPresets();
```

Having this scattered logic across all these files is hard to maintain.

Ferotiq:
> 

Agreed.
I also may have been the culprit of this lol

Ferotiq:
Also I wanna know what issues this caused on the site, unless you just noticed that the type was wrong.

PR #3652, opened 2022-10-14
Save speed stats in leaderboard update

### LINKED ISSUE #3650: Histogram of speed for time 15 and 60 for all users

mrbrianevans:
Saves a histogram data structure of speeds for buckets rounded to the nearest 10.

Signed-off-by: Brian Evans <ebrian101@gmail.com>

### Description
I have added to the leaderboard update function to save some global stats for all users.
It saves in `public` collection with type `speedStats`.
The data structure saved is:
```json
{
"type":"speedStats",
"english_time_15":{"110":1},
"english_time_60":{"90":1}
}
```

Closes #3650

### DIFF
--- a/backend/src/api/controllers/public-stats.ts
+++ b/backend/src/api/controllers/public-stats.ts
@@ -1,0 +1,10 @@
+import * as PublicStatsDAL from "../../dal/public-stats";
+import { MonkeyResponse } from "../../utils/monkey-response";
+
+export async function getPublicSpeedStats(
+  _req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { language, mode, mode2 } = _req.query;
+  const data = await PublicStatsDAL.getSpeedStats(language, mode, mode2);
+  return new MonkeyResponse("Public speed stats retrieved", data);
+}
--- a/backend/src/api/controllers/public-stats.ts
+++ b/backend/src/api/controllers/public-stats.ts
@@ -2,9 +2,9 @@ import * as PublicStatsDAL from "../../dal/public-stats";
 import { MonkeyResponse } from "../../utils/monkey-response";
 
 export async function getPublicSpeedStats(
-  _req: MonkeyTypes.Request
+  req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
-  const { language, mode, mode2 } = _req.query;
+  const { language, mode, mode2 } = req.query;
   const data = await PublicStatsDAL.getSpeedStats(language, mode, mode2);
   return new MonkeyResponse("Public speed stats retrieved", data);
 }
--- a/backend/src/api/controllers/public-stats.ts
+++ b/backend/src/api/controllers/public-stats.ts
@@ -1,10 +1,17 @@
 import * as PublicStatsDAL from "../../dal/public-stats";
 import { MonkeyResponse } from "../../utils/monkey-response";
 
-export async function getPublicSpeedStats(
+export async function getPublicSpeedHistogram(
   req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
   const { language, mode, mode2 } = req.query;
-  const data = await PublicStatsDAL.getSpeedStats(language, mode, mode2);
-  return new MonkeyResponse("Public speed stats retrieved", data);
+  const data = await PublicStatsDAL.getSpeedHistogram(language, mode, mode2);
+  return new MonkeyResponse("Public speed histogram retrieved", data);
+}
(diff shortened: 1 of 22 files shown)


mrbrianevans:
## Status update
Finished: 
 - save public stats when leaderboard is calculated
 - serve public stats on endpoint
 - get data on frontend with Ape client
 
Still todo:
 - implement histogram display logic (following https://github.com/monkeytypegame/monkeytype/commit/7dff50dc3032cd0c7a897da70825892745878a5b)
Should look something like this (but with users on the y axis):
image

mrbrianevans:
Made some test data and this is what the graph looks like:
image


## Perf
The performance of calculating the buckets when the leaderboard is updated:
(using my test data set of ~50,000 users)
```
system_lb_update_english_time_60   "Aggregate 1.078039699971676s, loop 0.002202499985694885s, insert 2.367039000034332s, index 0.24052349996566771s, histogram 0.005574600040912628"
```
Only for comparison purposes, ignore absolute values. 
Takes 20x less time to calculate histogram buckets than to aggregate, so performance shouldn't be a problem.

I think this feature is ready to be merged, unless there are any requests for changes?

backend/src/api/controllers/public-stats.ts
```diff
@@ -0,0 +1,10 @@
+import * as PublicStatsDAL from "../../dal/public-stats";
+import { MonkeyResponse } from "../../utils/monkey-response";
+
+export async function getPublicSpeedStats(
+  _req: MonkeyTypes.Request
```
Bruception, line 5: Nit:

req instead of _req

backend/src/api/routes/public-stats.ts
```diff
@@ -0,0 +1,24 @@
...
+const GET_MODE_STATS_VALIDATION_SCHEMA = {
+  language: joi.string().required(),
+  mode: joi.string().required(),
+  mode2: joi.string().required(),
+};
+
+const router = Router();
+
+router.get(
+  "/speed",
```
Bruception, line 16: Should be more descriptive.

/all-time-leaderboards

frontend/src/ts/ape/endpoints/public-stats.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class PublicStats {
+  constructor(private httpClient: Ape.HttpClient) {
+    this.httpClient = httpClient;
+  }
+
+  async getSpeedStats(searchQuery: SpeedStatsQuery): Ape.EndpointData {
+    return await this.httpClient.get(`${BASE_PATH}/speed`, {
+      searchQuery,
+      headers: {
+        "Client-Version": CLIENT_VERSION,
```
Bruception, line 20: No need to specify this header.

backend/src/api/routes/public-stats.ts
Miodec: As discussed on Discord - personally disagree on the leaderboards API thing.

mrbrianevans: Even though the data source is from the leaderboard, the data itself being served on this endpoint is not related to the leaderboard. Its general site statistics. If in future we wanted an endpoint which serves total time spent typing on monkeytype, it would be similar to these speed stats, but unrelated to leaderboard. So personally, I think the `public-stats` route makes sense for this and could have future use as well, but I'm happy to change it to leaderboards if that is decided and agreed.

mrbrianevans: As for the specific endpoint being `/all-time-leaderboards`, I'm not sure what to go with. 

I would describe the action as "get public speed stats data", and those components are described by `GET /public-stats/speed`. 

Reluctant to make it all time leaderboards, as it is only peoples most recent pb and isn't ranking users (which in my mind is what a leaderboard is). Any ideas?

mrbrianevans:
Thanks @Bruception for the feedback, I've implemented those points (except for the route change discussed on discord).

backend/src/api/routes/public-stats.ts
Miodec: I think rename `public-stats` to `public` and name this route `speedHistogram`

Miodec:
Things left to do:

 - [ ] Add an endpoint which gets the typing stats (started, completed, time typing)
 - [ ] Add a section above the about section, call it stats
 - [ ] Display tests started, completed, time typing in hours (3 columns)
 - [ ] Display the time *60* histogram below the stats
 - [ ] Make sure the about page is responsive (currently the histogram is causing an overflow)

PR #3664, opened 2022-10-17
New language: english doubleletter 

QuackerDeezles:
english doubleletter is a new language under the english set that exclusively uses words that have the same letter used consecutively. Examples include mullet, zoom, and mayonnaise.

### Description

Closes #


frontend/static/languages/english_doubleletter.json
```diff
@@ -0,0 +1,212 @@
+{
+    "name": "english doubleletter",
```
Miodec, line 2: Missing `_`

Miodec (changes requested):

PR #3681, opened 2022-10-25
New Language: "lorem ipsum" by me

QuackerDeezles:
New typing language labeled "lorem ipsum" that houses roughly 200 lorem-ipsum type words from artificial intelligence and the original script. Of course, I'm using both lorem and ipsum as examples of the words.


Miodec:
The filename, "name" field inside and all the name references need to be the same. 

Please update lipsum.json to lorem_ipsum.json and the name field inside.

PR #3687, opened 2022-10-26
French 630k

EMRD95:
Got the full list of french words with all variations and grammatical forms.
I'd like to also propose two others lists with all the french words containing accents/diacritics (280k), and all the french words not containing accents/diacritics (350k) if that's okay.
If no objections, which name to give to the lists?
1: french no accents / french accents
2: french no diacritics / french diacritics


Miodec (changes requested):
Rename to 600k

EMRD95:
> Not sure about the proposed lists, cant you achieve the same thing with lazy mode?

Yeah I was just rethinking about it, and a "commonly misspelled" or "hard" mode would be better.

PR #3695, opened 2022-10-29
Fixed #2361

### LINKED ISSUE #2361: Commandline does not work on Android

- [X] I have cleared my cache

### Current Behavior

If I type something in the textbox, it should filter the commands but it does not do that on Android Chrome.

https://user-images.githubusercontent.com/69350358/151539967-eb3041dd-0e9a-4548-8960-61f66c06b171.mp4

### Expected Behavior

It should filter the commands.

### Steps To Reproduce

_No response_

### Environment

- OS: Android 11
- Browser: Google Chrome
- Browser Version: 97.0.4692.87 

### Anything else?

_No response_

### Config

_No response_

aungKhantPaing:
### Description

Closes #2361. Fixed command line input bug on android. 

1. Replaced the `keyup` listener with `input` listener on the command line component.
2. Removed meta key checks that won't be needed with `input` listener.

### DIFF
--- a/frontend/src/ts/commandline/index.ts
+++ b/frontend/src/ts/commandline/index.ts
@@ -382,7 +382,7 @@ function restoreOldCommandLine(sshow = true): void {
   if (sshow) show();
 }
 
-$("#commandLine input").on("input", () => {
+$("#commandLine input").on("input", (e) => {
   commandLineMouseMode = false;
   $("#commandLineWrapper #commandLine .suggestions .entry").removeClass(
     "activeMouse"
--- a/frontend/src/ts/commandline/index.ts
+++ b/frontend/src/ts/commandline/index.ts
@@ -382,21 +382,11 @@ function restoreOldCommandLine(sshow = true): void {
   if (sshow) show();
 }
 
-$("#commandLine input").keyup((e) => {
+$("#commandLine input").on("input", () => {
   commandLineMouseMode = false;
   $("#commandLineWrapper #commandLine .suggestions .entry").removeClass(
     "activeMouse"
   );
-  if (
-    e.key === "ArrowUp" ||
-    e.key === "ArrowDown" ||
-    e.key === "Enter" ||
-    e.key === "Tab" ||
-    e.code == "AltLeft" ||
-    (e.key.length > 1 && e.key !== "Backspace" && e.key !== "Delete")
-  ) {
-    return;
-  }
   updateSuggested();
 });
 
--- a/frontend/src/ts/commandline/index.ts
+++ b/frontend/src/ts/commandline/index.ts
@@ -382,7 +382,7 @@ function restoreOldCommandLine(sshow = true): void {
   if (sshow) show();
 }
 
-$("#commandLine input").on("input", (e) => {
+$("#commandLine input").on("input", () => {
   commandLineMouseMode = false;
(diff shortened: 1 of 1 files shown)


frontend/src/ts/commandline/index.ts
```diff
@@ -382,21 +382,11 @@ function restoreOldCommandLine(sshow = true): void {
   if (sshow) show();
 }
 
-$("#commandLine input").keyup((e) => {
+$("#commandLine input").on("input", () => {
   commandLineMouseMode = false;
   $("#commandLineWrapper #commandLine .suggestions .entry").removeClass(
     "activeMouse"
   );
-  if (
```
rizwanmustafa, line 390: Why did you remove these lines?

frontend/src/ts/commandline/index.ts
```diff
@@ -382,21 +382,11 @@ function restoreOldCommandLine(sshow = true): void {
   if (sshow) show();
 }
 
-$("#commandLine input").keyup((e) => {
+$("#commandLine input").on("input", () => {
```
rizwanmustafa, line 385: Replace this with:
```suggestion
$("#commandLine input").on("input", (e) => {
```
 and keep the other lines in.

rizwanmustafa (changes requested):

frontend/src/ts/commandline/index.ts
aungKhantPaing: Those lines are for filtering meta keys (Alt, Shift) apart from other keys. `keyup` listener captures both types of event. This won't be the case with `input` listener since it only captures the events from non-meta keys.

rizwanmustafa: My bad. I guess then we need to listen to both of these events. 'input' for filtering and 'keyup' for listening to these meta events.

rizwanmustafa: Like so:
```
$("#commandLine input").on("keyup", (e) => {
  commandLineMouseMode = false;
  $("#commandLineWrapper #commandLine .suggestions .entry").removeClass(
    "activeMouse"
  );
  if (
    e.key === "ArrowUp" ||
    e.key === "ArrowDown" ||
    e.key === "Enter" ||
    e.key === "Tab" ||
    e.code == "AltLeft" ||
    (e.key.length > 1 && e.key !== "Backspace" && e.key !== "Delete")
  ) {
    return;
  }
  updateSuggested();
});

$("#commandLine input").on("input", () => { updateSuggested(); });
```

frontend/src/ts/commandline/index.ts
```diff
@@ -382,21 +382,11 @@ function restoreOldCommandLine(sshow = true): void {
   if (sshow) show();
 }
 
-$("#commandLine input").keyup((e) => {
+$("#commandLine input").on("input", (e) => {
```
rizwanmustafa, line 385: My bad again. Please remove the `e` that I added as a parameter. Everything else seems fine.

```suggestion
$("#commandLine input").on("input", () => {
```

rizwanmustafa (changes requested):

rizwanmustafa (approved):

Miodec:
Well, that was much easier than i thought. Thanks.

PR #3709, opened 2022-10-31
Hide favourite quote button for unauthenticated user

### LINKED ISSUE #3707: Favourite quote option should not be visible if the user is not logged in

- [X] I have cleared my cache

No

Yes

No

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

The user is able to click on the favorite quote button even when logged out, because of which an error toast is displayed which isn't relevant.

### Expected Behavior

User shouldn't see the favourite quote button when logged out.

### Steps To Reproduce

_No response_

### Environment

- OS: Linux
- Browser: Firefox
- Browser Version: 103.0 (64 bit)

image

### Anything else?

_No response_

Prajwalrajbasnet:
### Description
Hide favourite quote option button in top bar if user is not authenticated/logged out.

Closes #3707 

### Logged Out
image

### Logged In
image


Miodec:
No need to import auth. We can let the account controller handle this on sign in and sign out.

The default state was just incorrect. Setting the button to hidden by default fixes this.

PR #3710, opened 2022-10-31
Created a theme, first time contributing on github

Harshnatuskar:
### Description

Closes #


Miodec:
Thanks. Unfortunately, the color combination is not my favorite - plus its similar to other themes already available.

We really have too many themes at the moment so I will only accept truly good looking and unique themes.

PR #3725, opened 2022-11-07
prevent redirect to /profile/undefined when going to the page of a non-existing profile

FurriousFox:
Fix it automatically changing the url in the url bar (which gets requested if someone reloads the page) to https://monkeytype.com/profile/undefined (which apparently is already the username of an existing user) when going to the profile url of a non-existent user (for ex. https://monkeytype.com/profile/dfjksahjkadfsjhkasfhjkadhjksf ) , by not replacing url state unless the response status is 200


Miodec (approved):
Thanks

PR #3733, opened 2022-11-08
Fix Typo

Shakenbeet:
Removed duplicate "to" on line 14993.


Miodec:
Make sure to update the length field

Shakenbeet:
Sorry, still new to contributing, but I got the character count right this time 😁

Miodec:
After reading through the quote and looking it up online it looks like the double `to` is correct.

PR #3736, opened 2022-11-09
Fixed two typos in Georgian quotes

avetisovi:


rizwanmustafa:
Please update the length field so that the build can complete. Thanks

rizwanmustafa:
If you could translate the quotes, it would be great. By the way thanks for fixing the length field.

avetisovi:
> If you could translate the quotes, it would be great. By the way thanks for fixing the length field.
I changed just two letters in two quotes, so the meaning remains the same.

17. The people, who were in a state of despair a moment ago, were now cheering and cheering. The mass of people broke up, dispersed and spread across the field. Everyone went home fully satisfied that they had seen this disgraceful and heart-wrenching game. Peter looked around and saw that no one was staying here. The army also left, only two guards remained near gallows. That red-shirted man also changed into some other clothes and left. The unfortunate child was left hanging motionless.

20. When Teimuraz's life bent his neck to the ground and bent him to the ground, the bent scribe also took his hand to his books - those books, which for Teimuraz were comparable to the children and were more necessary and precious than furniture, family treasures, Margot's precious stones, crockery and bed linen.

avetisovi:
> > If you could translate the quotes, it would be great. By the way thanks for fixing the length field.
>  I changed just two letters in two quotes, so the meaning remains the same.
> 
>     17. The people, who were in a state of despair a moment ago, were now cheering and cheering. The mass of people broke up, dispersed and spread across the field. Everyone went home fully satisfied that they had seen this disgraceful and heart-wrenching game. Peter looked around and saw that no one was staying here. The army also left, only two guards remained near gallows. That red-shirted man also changed into some other clothes and left. The unfortunate child was left hanging motionless.
> 
>     18. When Teimuraz's life bent his neck to the ground and bent him to the ground, the bent scribe also took his hand to his books - those books, which for Teimuraz were comparable to the children and were more necessary and precious than furniture, family treasures, Margot's precious stones, crockery and bed linen.

Miodec (approved):

PR #3779, opened 2022-11-28
add spacing between sentences

ntzm:
### Description

before:
image

after:
image


rizwanmustafa (approved):

Miodec:
And this won't get removed by prettier?

rizwanmustafa:
I don't think so

Miodec:
And i've been manually adding mergins with CSS like a caveman..

PR #3780, opened 2022-11-29
Fix weekly leaderboard timestamp logic

Bruception:
### Description

Closes #


backend/src/utils/misc.ts
```diff
@@ -206,7 +206,7 @@ export function mapRange(
 }
 
 export function getStartOfWeekTimestamp(timestamp: number): number {
-  const date = new Date(timestamp);
+  const date = new Date(getStartOfDayTimestamp(timestamp));
```
Bruception, line 209: argh

Miodec (approved):

Issue #3207, opened 2022-06-23
Implement a better profanity detection solution


monkeytypegeorge:
Our current naive solution marks a lot of false positives. We need to consider existing technologies that might handle this better.

angusjoshi1:
Hey! How about using this package for this? https://www.npmjs.com/package/bad-words . It works in much the same way as the current profanity detection, but contains a few extra useful functions and perhaps a more comprehensive list?

rustom:
What are examples of the false positives? Could implement a library solution quickly @Miodec

Miodec:
The main problem is with short words in those lists (either ours or the one that @angusjoshi1 suggested) - if your nickname happens to contain those combinations of letters it will be considered profane. I think best solution here for now is just to manually refactor the list (compare it to https://github.com/web-mech/badwords/blob/master/lib/lang.json perhaps) to include anything we have missed and exclude anything that is not very explicit/ could be easily misinterpreted.

scpdavis1:
Hi, I am interested in attempting this. is anyone assigned to it yet?

scpdavis1:
Hi! im new to contributing to open source projects but I would like to try this out. Can someone point me to where I can get started on understanding the related code?

Miodec:
> Hi! im new to contributing to open source projects but I would like to try this out. Can someone point me to where I can get started on understanding the related code?

No code to understand really. Just expand this list a bit, and make sure its not prone to false positives https://github.com/monkeytypegame/monkeytype/blob/master/backend/src/constants/profanities.ts

scpdavis1:
Hello again! what current methods are you guys using to test for false positives? or if there is documentation on that topic could someone point me there?

Miodec:
We dont really have any process set up for that. When I went through the list last time I just removed very short words and combinations of letters that seemed too generic

PR #3812, opened 2022-12-06
Added the merged list and updated relevant vars

scpdavis1:
### Description

As requested in issue #3207:

- We included badwords-list npm package to update the existing profanities list.
- We then merged the existing list with the badwords-list package list. The merged list does not contain duplicates. 
- We deleted some profanities that were very short which could have caused false positives.

Closes #

### DIFF
--- a/backend/src/api/routes/users.ts
+++ b/backend/src/api/routes/users.ts
@@ -460,12 +460,14 @@ const profileDetailsBase = joi
   .string()
   .allow("")
   .custom((value, helpers) => {
-    return containsProfanity(value)
-      ? helpers.error("string.pattern.base")
-      : value;
+    if (containsProfanity(value)) {
+      return helpers.error("string.profanity");
+    }
+
+    return value;
   })
   .messages({
-    "string.pattern.base": "Profanity detected. Please remove it.",
+    "string.profanity": "Profanity detected. Please remove it.",
   });
 
 router.patch(


backend/src/constants/profanities.ts
```diff
@@ -1,420 +1,401 @@
...
-  "g00k",
-  "gayboy",
-  "gaygirl",
-  "gayz",
-  "god-damned",
-  "h00r",
-  "h0ar",
-  "h0re",
-  "hoar",
+exports.profanities = [
```
Bruception, line 4: Please stick to project style and prefer declaration exports

scpdavis1: OK I apologize for that mistake.

backend/src/constants/profanities.ts
```diff
@@ -1,420 +1,401 @@
...
+   "xrated", 
+   "zabourah" 
 ];
 
-export const regexProfanities = profanities.map((profanity) => {
-  const normalizedProfanity = _.escapeRegExp(profanity.toLowerCase());
-  return `${normalizedProfanity}.*`;
-});
+
+var list = require('badwords-list'),
```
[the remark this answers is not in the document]
scpdavis1: Alright I was not aware this was antiquated sorry about that. I have various other lists I can attempt that I think would work a lot better. I am going to test those. I appreciate the comments!

backend/src/constants/profanities.ts
Bruception: No harm done!

PR #3829, opened 2022-12-13
swiss_german_2k.json more words for Swiss

egemasta:
### Description

Made a language pack with 2k words for the Swiss. Modified the _list and _group file and put my json file in the corresponding folder.


frontend/static/languages/_groups.json
```diff
@@ -43,6 +43,10 @@
     "name": "swiss_german",
     "languages": ["swiss_german", "swiss_german_1k"]
   },
```
Miodec, line 45: Add the language to the list here instead of creating a new group

Miodec (changes requested):

Miodec (approved):

PR #3834, opened 2022-12-15
Create ayu_dark.css

aspizu:
### Description

Closes #


DecoyGameDev (commented):
I think this should be added.

Miodec:
Add this file in your previous pr instead of two prs.

Issue #2206, opened 2021-12-24
"Click or press any key to focus" not working


raphaelmatto:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Current Behavior

During an active typing session, if I press the `tab` key the `Restart Test` button appears. If I want to return to the active session, I am instructed to `Click or press any key to focus`. Clicking with the mouse works, but pressing a key does nothing (except `return`, which restarts the test).

### Expected Behavior

When I click any key that's not `esc` or `return`, I'm placed back in the active typing session.

### Steps To Reproduce

1. Start a typing test
2. Press `tab`
3. When the `Click or press any key to focus` appears, press any button that isn't `esc` or `return`

### Environment

- OS: MacOs Monterey 12.1
- Browser: Chrome
- Browser Version: Version 96.0.4664.110 (Official Build) (x86_64)

### Did you check incognito mode?

- [x] I have checked incognito mode

Issue #1941, opened 2021-10-04
Unable to restart test with Tab + Enter


kennethcheo:
Hi, not sure if anyone else is facing this or just me. 

![Animation](https://user-images.githubusercontent.com/42935499/135850373-715e0d71-d390-4e2a-8d86-c9b5b5cf8a51.gif)

I apologize if the gif isn't that clear.
- Pressing Tab selects the "Restart Test button"
- Pressing Enter unselects the button but the test doesn't restart. 
- It does work with the mouse though.

Expected behaviour: Tab + Enter should restart a test.

- OS: Tested on both Windows 10 and Linux (Xubuntu)
- Browser: Tested on Firefox 92.0.1, Chrome 94.0.4606.61 (normal and incognito mode)

I cleared my cache before this.

Miodec:
Checked incognito?

Miodec:
Deployed a small change. Be sure to clear cache and try again. If this doesn't work then I'm not sure, but you can work around the issue by enabling quick tab mode in the settings.

SeerLite:
This one is weird as you're testing multiple OSs _and_ browsers and consistently getting the same issue, but none of us can reproduce it.

Are you using any extensions that could be messing with it?
Does it happen in the result screen too or only during the test?
Does it happen with no Monkeytype configuration?

kennethcheo:
Still doesn't work after the PR, on incognito mode with my cache cleared and the default configuration. The only extensions I have are BitWarden and Ublock Origin, and I don't think they should mess with anything. 

I booted up my old laptop to test this out and it works there, so I thought it might be an issue with my PC/keyboard. But strangely enough, it works fine on the result screen? 

I enabled quick tab mode as a workaround. Feel free to close this issue if no one else is able to reproduce this. Thanks for your help!

Miodec:
Maybe you have some external software that remaps keys or something? What happens when you press enter on this site? https://keycode.info/

rachelsalazar:
I'm also having this error @kennethcheo. And @Miodec, I went to the keycode.info website and when I pressed enter it recognized the enter key.

kennethcheo:
I managed to identify and fix the problem. I'm using QMK configurator to configure my Planck's layout.

[...]

PR #3837, opened 2022-12-16
Ignore modifier keys for test autofocus

andrewhchang:
### Description
Currently, when the user is focused on the restart button, any other key pressed will refocus the test. This poses problems for users who utilise remapping functions such as `ModTap` commonly utilised on sub 60% boards where `<hold> == Shift, <tap> == Enter`. This causes two keystrokes to be sent in quick succession such that a modtap user with the aforementioned mapping will never be able to use the `Tab+Enter` restart as `Shift` is always sent first.

This was resolved in https://github.com/monkeytypegame/monkeytype/commit/9a0026a1ab6de87e7597d5216912a5849b1cbb2d in reponse to  https://github.com/monkeytypegame/monkeytype/issues/1941, but was undone in https://github.com/monkeytypegame/monkeytype/commit/76664fc7ae778a58bf803a8558dc71e2279d6941 as part of  https://github.com/monkeytypegame/monkeytype/issues/2206 - as seemingly targeting the refresh button specifically caused unexpected behaviour.

Rather than dismiss keystrokes, I've elected to implement a small 200ms delay to refocus the test, other than for keys `Tab` and `Enter`, for which it will not take place. This gives time for `Enter` to be the prevailing key in the niche case above, and for the first returning key to resume normal operation.

Given every key will autofocus the test, the delay will never be present to affect typing as the the first keydown event will autofocus regardless after 200ms, or immediately if esc or tab, i.e. no delayed keystrokes appearing in typing test, no infinite debounce.


Miodec:
I kinda don't like the sleep. Especially when out of focus warning is disabled and you want to start typing - the first keypresses within the 200ms are ignored. 

Maybe we can only autofocus on alphanumeric keys? And ignore the problematic shift?

andrewhchang:
> I kinda don't like the sleep. Especially when out of focus warning is disabled and you want to start typing - the first keypresses within the 200ms are ignored.

Yeah good catch, didn't look at this scenario - that's a bummer.

> Maybe we can only autofocus on alphanumeric keys? And ignore the problematic shift?

Targeting alphanumeric might add some complexities with non-english languages and special characters, perhaps we could meet in the middle and just ignore modifier keys.

Made the change for the time being - would you also want to update the autofocus warning to something else to better describe the change? e.g. "Click here or start typing to focus"

Miodec:
> Targeting alphanumeric might add some complexities with non-english languages and special characters, perhaps we could meet in the middle and just ignore modifier keys,

Yeah that's alright

> e.g. "Click or press any key to focus" -> "Click here or start typing to focus"

Yeap, good idea

andrewhchang:
All done 👌🏻

Miodec (approved):

PR #3845, opened 2022-12-20
Create latin.json

TheEccentricDuck:
I would like to make Latin quotes an option, as I greatly appreciate Latin literature.

### Description

Closes #

### DIFF
--- a/frontend/static/quotes/latin.json
+++ b/frontend/static/quotes/latin.json
@@ -1,0 +1,17 @@
+{
+  "language": "latin",
+  "groups": [
+    [0, 100],
+    [101, 300],
+    [301, 600],
+    [601, 9999]
+  ],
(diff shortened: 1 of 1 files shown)


Miodec:
I think it would be good to start with a few more quotes

TheEccentricDuck:
No problem, I will get back to you on that.

在 2023-01-04 21:24:53，Jack ***@***.***> 写道：


I think it would be good to start with a few more quotes

TheEccentricDuck:
Hello Miodec, I think for now all the quotes will have to be brief proverbs (for the purposes of mass-generation through code) and I think longer quotes will have to just be added in the future and submitted by users.

Miodec:
Another PR added latin quotes. If you wish to add more quotes please expand that file

PR #3852, opened 2022-12-25
funbox: gave IPv6 addresses' parts leading zeros

jfrech:
If the shorter form is preferred, the zero part should instead be the empty string, since it is never written as `...:0:...`.


jfrech:
@Miodec Thanks for the heads up; fixed.

Miodec:
image

I dont think this is how its supposed to look like?

jfrech:
@Miodec Thank you for bearing with me and merging.

PR #3853, opened 2022-12-26
Optimise images with ImgBot

jay-tau:
[ImgBot] Optimize images

### Description

Optimise images as suggested by ImgBot

Closes #


Bruception:
Doesn’t seem like any changes were made.

PR #3858, opened 2022-12-27
create new theme: note dark

mynimi:
### Description
Created a new theme with a few small additions
Features:
- a notepad style overlay on the background (it will overlay background images too)
- error strike-through instead of underline
- fun little dot transformation on already typed text
- some more rounded corners

Screenshot 2022-12-27 172447

### DIFF
--- a/frontend/static/themes/_list.json
+++ b/frontend/static/themes/_list.json
@@ -1,5 +1,10 @@
 [
   {
+    "name":"note_dark",
+    "bgColor": "linear-gradient(to bottom, #1f1f1f 0% 19%, rgba(67, 70, 83, 0.4) 19% 21%, #1f1f1f 21% 70%, rgba(67, 70, 83, 0.4) 70% 72%, #1f1f1f 72% 100%)",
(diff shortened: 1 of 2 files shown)


mynimi:
actually, realized I've forgotten to fix the dot colors for the colormode and they display wrong colors, and the color isn't properly working in Firefox.
Was about to just commit the fix, but then I thought, maybe the entire thing might be better suited in the funbox? 
"dots" funbox that turns the previous letters into dots. I could split this up into theme and the text effect as a funbox mode, would that be better suited and if I close this and make two separate pull requests for that?

AzkaPrata (commented):
Great, send me screenshot all page from monkey type

Miodec:
What does it look like with a custom background applied?

mynimi:
The image is placed behind the lines right now. Adjusting it so it's behind the custom background is only a matter of adding z-index to it thugh, so if it shouldn't interfere with the custom background in any way, I could make that adjustment.

image

image

Miodec:
On top is fine

PR #3865, opened 2022-12-31
Adding customStyle property to 'Command' object interface

jerryzhou196:
This is the implementation of the [discussion post](https://github.com/monkeytypegame/monkeytype/discussions/3864) feature I submitted. 

Currently, some of the dropdown options within the Command Line Interface could benefit greatly from visual styling - especially the `Theme...` and `Font Family....`  dropdown options. By implementing an optional `customStyle` property of type `string` to the `Command` interface,  we could easily style certain dropdown options by passing in CSS strings when we are pushing each `Command` object into it's respective `CommandsSubgroup`. Here are some examples I made with this new `customStyle` property: 

image

This might make the theme selection more intuitive. I will admit it is kind of an eyesore with certain theme combinations. 

image

Here, each font is rendered accordingly instead of having to cycle through each one to preview it. 

This property could further be used within the funbox command dropdown and such. It lays a framework for future customization within the command line interface. 

I'm unsure if this violates the minimalist philosophy behind the command line interface. I just found it easier to be able to visually preview some options instead of the classic drag-my-mouse-through-the-options-and-wait-one-second-to-preview, and wanted to give my two cents.  🤠


Miodec:
This pr seems to be incomplete? Only has the type definition change.

Miodec:
Fonts are working. The themes code change still seems to be missing.

jerryzhou196:
Hey Jack, thanks for responding so fast - it's an honor. I was playing around with the theme code within the command line, and something about the aesthetic didn't seem as refined as the rest of the website.  I first implemented the font family as it seemed like a quality-of-life improvement without much compromise. I'll create a separate pull request for the themes, as I would like your thoughts on it first.

Here are some weird color combinations that I'm worried might destroy the design of the command line:

image

image


My plan to combat this is to move the favorites theme color so that they have their own dropdown section - that way we can display the themes in color sorted order and give a more visually appealing result. The old design combined the favorite theme colors at the top alongside the others, which looked a little weird when you applied this new style because you would have dropdown options that were suddenly dark-themed and then suddenly white-themed. 

The proposed change adds a new "favorites dropdown": 

image

For this tab, we wouldn't need any styling as I presume if the user favorites it, they probably already know what looks like. I tried add styling for these dropdown options, but it just looks too weird. 

image

What are your thoughts?

Miodec:
> Here are some weird color combinations that I'm worried might destroy the design of the command line:

Not sure what you mean.

> I'll create a separate pull request for the themes, as I would like your thoughts on it first. 

Just add it here well work on it.

> I also moved the favorite theme colors so that they have their own dropdown section

I would prefer to keep them where they are right now, at the top of the themes list.


Please commit the changes so I can play around with it.

jerryzhou196:
Committed changes to the theme. Tell me what you think. Currently, only the themes after the favorites are styled, but we could style the favorites too. I also changed the themes after to be in sorted order. Would you prefer the favorites to be styled too?

Miodec:
With single list command line OFF, when using the keyboard to navigate - you don't know which line is highlighted (there is some conflict in the CSS).

For single list command line ON, I would still use the full line styling, not the text-only styling.

jerryzhou196:
Good point. Changed so it colors the entire block.

image

Miodec:
Yeah looking good. Is it good to go?

Miodec:
Well, the more i looked at it the more I started thinking coloring the whole row is a bit much. What do you think of this?
image

jerryzhou196:
Ah, that's better. I was thinking something along the lines of that too. Have you implemented it code-wise? And would you like it for both the single-line command and the non-single-line command?

Miodec:
> Have you implemented it code-wise?

Yeah its all done all i have to do is push the commit.

Miodec:
What do you think

PR #3866, opened 2022-12-31
Blind mode - ON button text missing

z1nc0r3:
### Description

Settings > Blind Mode - Add `on` text to the button.


Bruception:
I think the missing text is on purpose.

z1nc0r3:
@Bruception I thought so too but was unsure.

PR #3867, opened 2022-12-31
House keeping

Bruception:
### Description

Closes #


rizwanmustafa:
Not sure if those ci changes are supposed to be there

Bruception:
> Not sure if those ci changes are supposed to be there

Yeah this is more of a housekeeping PR now.

PR #3869, opened 2023-01-01
Mapped letter k to make a sound in sound controller

jerryzhou196:
For some reason, the 'K' key isn't mapped to produce any sound within the sound-controller. I'm not sure if this is some sort of deliberate persecution against the letter k, but in case it is unintentional, here we go!

I mapped it according to the chromatic pattern within `codeToNote`, and choose A flat as the note since it is the least frequent to even out the distribution of sounds from the keys.

### DIFF
--- a/frontend/src/ts/controllers/sound-controller.ts
+++ b/frontend/src/ts/controllers/sound-controller.ts
@@ -306,7 +306,7 @@ const codeToNote: Record<string, GetNoteFrequencyCallback> = {
   BracketLeft: bindToNote(notes.F, 2),
   Equal: bindToNote(notes.Gb, 2),
   BracketRight: bindToNote(notes.G, 2),
-  KeyK: bindToNote(notes.Ab, 1)
+  KeyK: bindToNote(notes.Ab, 1),
 };
 
 type DynamicClickSounds = Extract<
--- a/frontend/src/ts/controllers/sound-controller.ts
+++ b/frontend/src/ts/controllers/sound-controller.ts
@@ -306,6 +306,7 @@ const codeToNote: Record<string, GetNoteFrequencyCallback> = {
   BracketLeft: bindToNote(notes.F, 2),
   Equal: bindToNote(notes.Gb, 2),
   BracketRight: bindToNote(notes.G, 2),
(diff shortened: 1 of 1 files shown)


Bruception:
This is deliberate. The way the notes are laid out is one octave starting at “Z” with the note C. And one octave starting at “Q” with the note C.

The home row would map to the black keys. So setting K to Ab doesn’t really make sense from a layout perspective.

jerryzhou196:
Ahhh makes sense. Thanks!

Bruception:
> Ahhh makes sense. Thanks!

No problem! Open to explore alternative layouts that make use of the letter K. ;)

jerryzhou196:
Awe thanks! After you showed me how it intentionally resembles the 12-note piano layout though, I definitely think it's the better layout. It actually opens up some interesting possibilities for playing songs with certain text prompts, which can almost act as sheet music now. 

What are your thoughts on adding some texts that resemble songs when paired with the sound as a potential funbox option actually? I would love to give it a stab. 

I scrapped together this [jazz typer](https://jerryzhou196.github.io/jazz-type/) a few days ago that generates music in the scale of c minor blues based on keyboard input, and sort of got infatuated with this intersection between music and typing. 😅

Bruception:
> Awe thanks! After you showed me how it intentionally resembles the 12-note piano layout though, I definitely think it's the better layout. It actually opens up some interesting possibilities for playing songs with certain text prompts, which can almost act as sheet music now.
> 
> What are your thoughts on adding some texts that resemble songs when paired with the sound as a potential funbox option actually? I would love to give it a stab.
> 
> I scrapped together this [jazz typer](https://jerryzhou196.github.io/jazz-type/) a few days ago that generates music in the scale of c minor blues based on keyboard input, and sort of got infatuated with this intersection between music and typing. 😅

I think adding randomly generated text as a form of score for a funbox is a great idea. That said, you would need to consider the layout a user is using since not all users use QWERTY. So it should generate characters relative to the layout. You’ll also need to make some server-side changes that will accept this new funbox.

jerryzhou196:
Those are really good points. I'll look into the backend connection with the funbox, alongside handling different layouts. Thank you so much.

PR #3873, opened 2023-01-02
Twilight Theme

zphyrj:
### Description

I added a new theme named twilight.

#### Home Page

home

#### Mid Type Test

type-test

#### Mid Type Test with Flipped Test Colors

twilight-typing

#### Mid Type Test with Colorful Mode

colorful


Miodec:
Looks similar to other themes available. Sorry, I'm only merging very unique themes now because we have too many.

PR #3880, opened 2023-01-04
Add tabs to js quote 24

jameslounds:
JS Quote 24 didn't have tabs, so i  added some

PR #3881, opened 2023-01-04
correct  length for js quote 24

jameslounds:
In #3880, I forgot to update the length of the quote. You were so quick merging, CI didn't have time to complete :)


Miodec:
Thats what i get for not letting the build process finish

PR #3882, opened 2023-01-04
Create english_gratitude_100.json

rmurcek:
### Description

Adds a new type of language for use in priming a particular domain of words in its users minds. This language contains 100 words useful in expressing gratitude. It may help users embody feelings of gratitude as they type. This could be useful to intentionally shift one's mindset or to prepare to express gratitude. 

Future work includes expanding this language, adding other useful domains (for example to help express other emotions), and building its member words scientifically (I generated this first batch by pulling example sentences for expressing gratitude from the internet, but it could be generated algorithmically from any number of etymological or psychological references).


Miodec:
This seems oddly specific. You should just use custom mode for this instead.

rmurcek:
Makes sense @Miodec , thanks!

PR #3886, opened 2023-01-05
add new theme: dark note

mynimi:
### Creates a new theme
Features
* note overlay
*  Dot animation 
* rounded corners on buttons and the test setup.

image
image

### DIFF
--- a/frontend/static/themes/_list.json
+++ b/frontend/static/themes/_list.json
@@ -1,5 +1,10 @@
 [
   {
+    "name" : "dark_note",
+    "bgColor": "linear-gradient(to bottom, #1f1f1f 0% 19%, rgba(67, 70, 83, 0.4) 19% 21%, #1f1f1f 21% 70%, rgba(67, 70, 83, 0.4) 70% 72%, #1f1f1f 72% 100%)",
+    "mainColor": "#f2c17b"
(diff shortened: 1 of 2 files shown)


Miodec:
The typed text animation seems very distracting. Also i keep losing track of the caret, the color could be brighter.

mynimi:
Would you like me to change the way the animation behaves (slow it down or maybe add more delay before it starts) or remove it entirely?

Miodec:
Not sure, up to you

mynimi:
Making the dots more subtle wasn't quite working out. Kept the original idea, but made it a more subtle change. If you feel like it's still too distracting, I'm fine killing it completely.

mynimi:
honestly, I like the dots a lot, but I can see how they may be distracting (I think probably especially for really fast typers). Maybe instead of baking the typing effect into the theme I could make the dots a funbox feature, so people can turn it on or off however they like and can use the theme with or without them? What do you think?

Miodec:
Funboxes are meant to change the experience in some way. Hiding the already typed words doesn't really change much.

mynimi:
gave the dot animation another try. Made the scaling of the text a bit more linear and changed some of the pacing.

PR #3887, opened 2023-01-05
Add support for Swiss Keyboard layout in ch-de and ch-fr varieties

mynimi:
### Description
Added support for the Swiss Keyboard in CH-DE and CH-FR flavors.

Since the proper diaeresis symbol `¨` that's associated with one of the keys causes issues with the uppercase switch since it's a combining character and swallows up the `!` I replaced it with a two dot leader `‥` which still conveys the visual message of it being an accent but doesn't have to be combined with a space for correctness.


Miodec:
Hmm, ill actually probably rename them to `swiss_german` and `swiss_french`. Do you think thats alright?

PR #3890, opened 2023-01-07
Fix letter i in ukrajins'ka latynka

### LINKED ISSUE #3889: Letter `i` is being always an error for `ukrainian_latynka` language family

- [X] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

image
As I investigated the bug is inside ukrainian_latynka.json simply having wrong letters. Current letter `i` inside them has encoding of 1110, which is Belarusian letter instead of latin.

### Expected Behavior

_No response_

### Steps To Reproduce

1. Go to monkeytype.com
2. Change language to **ukrainan latynka**
3. Try to write a word that contains letter `i`. You will get an error, but this shouldn't happen.

### Environment

_No response_

### Anything else?

_No response_

tumophiy:
### Description
Closes #3889
I simply ran a script for swapping letters:

```python
import os
import glob

filenames = glob.glob('*latynka*')
print(filenames)

for filename in filenames:
    with open(filename, 'r', encoding='utf8') as file:
        contents = file.read()

    contents = contents.replace(chr(1110), chr(105))

    with open(filename, 'w', encoding='utf8') as file:
        file.write(contents)

print("Done!")
```

And make sure that now files have the intended letter via copy pasting letter into terminal like:
image

### Other notes
- Yes this is very simple and maybe there was no need for opening an issue but I wanted to try that. Anyway would you open an issue if you were having same problem, or would you simply make a pr?


Miodec:
> How to mark an issue that this pr fixes? I don't see how to do that(

Add `closes #xxxx` to your pr description.

Also, please fix conflicts.

tumophiy:
Thx, I'll now how to mark `closses #xxxx` in future
Please wait a bit before merging, I'm worried with the fact that I may have accepted wrong changes in conflicts

tumophiy:
Sorry for force pushing. Now it should be fine🤞

Miodec (approved):

PR #3901, opened 2023-01-10
Profanity Detection Updated. False Negative improved

scpdavis1:
### Description
This PR contains an updated list of profane words.
- We merged the list in the NPM package profane-words with the existing monkeytype list in order to preserve specific words.
- We omitted some smaller words from both lists in order to decrease the amount of false positives.
- We also manually edited the list to add some words we found that were not in the package list or the original.
-
Additionally, We merged the lists so that there were no duplicates in the overall list that is used for regexProfanities.

Closes #
Related issue: #3207


Miodec:
Please fix build issues

Miodec:
Another PR added latin quotes. If you wish to add more, please expand that file.

Miodec:
Whops. Commented on the wrong PR

Miodec:
Honestly, looking at the package you've chosen, it will increase the number of false positives. The list seems a bit too restrictive, including things like `evl`, `xx`,`azz`,`dp`,`feg` and many others. (I know on their own they are bad but due to the us checking for those strings *anywhere*, its gonna lead to a lot of false positives)

PR #3906, opened 2023-01-10
Added Language Yiddish

JellieTheForg:
Yiddish is the language of the Ashkenazim, who are central and eastern European Jews. 

I added the language and 200 words, also put it in list.json and groups.json. 
Because it reads from right to left I'm having issues with the keyboard layout code and formatting.


Miodec:
Please fix build issue, please move the language in the group and list file above the code languages.

PR #3918, opened 2023-01-15
Replace `npm` with `yarn`

rizwanmustafa:
This PR will replace `npm` with `yarn`. The reason behind this is that yarn is faster than npm and I love the feel of it.


rizwanmustafa:
Really sorry. Didn't mean to open it here.

PR #3925, opened 2023-01-17
Make quote actions not break with c++ language

O-X-E-Y:
'code c++' as a language currently obviously has `+` symbols which means stuff like quote ratings break because of the validation regex. I feel like changing `\w` straight to `.` would also be fine since languages and their names get approved manually, but this should be fine for now


Miodec:
I don't think the space is necessary?

O-X-E-Y:
I wasn't sure about the space, it seemed like it'd be fine without but some regexes don't like it when you leave it out so I kept it in. I can remove it if you want

Miodec (approved):

PR #3936, opened 2023-01-24
Allowing configuration endpoint access if user is configuration mod

Miodec:
Added a middleware that only applies an array of middlewares based env variable

### DIFF
--- a/backend/private/script.js
+++ b/backend/private/script.js
@@ -228,6 +228,19 @@ window.onload = async () => {
     dataResponse.json(),
   ]);
 
+  if (schemaResponse.status !== 200 || dataResponse.status !== 200) {
+    const root = document.querySelector("#root");
+    let html = "";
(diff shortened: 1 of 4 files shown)


backend/private/script.js
```diff
@@ -228,6 +228,19 @@ window.onload = async () => {
...
+    if (schemaResponse.status !== 200) {
+      html += `<i class="unknown-input">Error fetching configuration schema: ${schemaResponseJson.message}</i>`;
+    }
+    if (dataResponse.status !== 200) {
+      html += `<i class="unknown-input">Error fetching configuration data: ${dataResponseJson.message}</i>`;
+    }
+    root.innerHTML = html;
+    return;
+  }
+
```
Miodec, line 243: This is already on master

Bruception (approved):

Bruception (approved):

PR #3943, opened 2023-01-25
add backup font-family "sans-serif"

ShriramKhandbahale:
In some cases when the styles are not loaded yet, the fonts are rolled back to the default 'serif' family which looks very ugly


Miodec (approved):

Miodec:
Good idea

PR #3944, opened 2023-01-25
Fix french punctuation typo - missing space

Pantoofle:
### Description

In french, there is always a space character before ! ? or : punctuation signs. 
This PR fixes the few quotes missing this space character.


Miodec (approved):

Miodec (changes requested):
If you add spaces you also need to increment the length field

PR #3949, opened 2023-01-26
Add typescript support for jest

rizwanmustafa:


rizwanmustafa:
The previous PR had some issues which hopefully will be eradicated in this one. I am using the backend's  `tsconfig.json` with some slight changes. This PR will give typescript support for tests in IDEs like Visual Studio Code.

PR #3951, opened 2023-01-26
Structure

### LINKED ISSUE #3861: Bitwarden autofill not working

Miodec:
This PR adds many optimisations aimed to reduce the number of elements in the DOM at all times by removing unused pages and popups from the DOM instead of hiding them. This should result in quicker HTML and CSS recalculations when the DOM changes.

Page change has been optimised by identifying long tasks and splitting them up. This ensures animation frames will always fire, making the page appear more responsive (even if actual page change speeds may have not changed that much)

(PR may contain random fixes here and there - most of them were cherry picked onto master)

Also closes #3861


backend/src/api/controllers/user.ts
```diff
@@ -170,10 +170,10 @@ function getRelevantUserInfo(
   return _.omit(user, [
     "bananas",
     "lbPersonalBests",
-    "quoteMod",
     "inbox",
     "nameHistory",
     "lastNameChange",
+    "_id",
```
Bruception, line 176: Isn’t this needed?

Miodec: I don't think we use the ObjectID anywhere?

Bruception (approved):
lgtm iiw

PR #3979, opened 2023-02-06
add sturdy ortho keyboard layout

Dainternetdude:
### Description

Add the version of the Sturdy keyboard layout designed for Ortholinear keyboard hardware.


Miodec:
Please move it in the list to be next to other sturdy layouts

PR #3996, opened 2023-02-15
Updated Catppuccin themes

computergnome99:
### Description

Added updated Catppuccin themes based on recent Catppuccin updates. Mainly, splitting the Catppuccin theme into the four major sub-themes, as well as updating the color values to the most recent values.

[Catppuccin Palette](https://github.com/catppuccin/catppuccin#-palette)

Latte Theme:
Latte

~~Frappé Theme:~~ **Removed**
Frappe

~~Macchiato Theme:~~ **Removed**
Macchiato

Mocha Theme:
Mocha

I understand these themes are very similar, but I figured since the Catppuccin style guide had been updated, the theme should be updated also. *(Also I was frustrated the old theme didn't match my GTK theme anymore)*


computergnome99:
> Thats a bit too many themes under the same name and general style. Please pick one light and one dark.

Sounds good. I'll probably go with the Latte (the lightest) and Mocha (the darkest) themes. I'll update the code to only include those two in a bit here.

Miodec:
Looking at the themes now it looks like you changed from green to blue main color? Why? It kinda makes it look like other themes like drowning, or iceberg light.

Miodec:
Latte also just looks like iceberg light

computergnome99:
I took the colors from the updated Catppuccin palette, more or less agnostic of other themes already created for this platform. If these are too similar to something else we can certainly go with another variant.

We can also just abandon this PR and keep the old theme the way it was - users can update to the new Catppuccin palette using custom themes.

Miodec:
Yeah, i think i would prefer keeping the current one

PR #4008, opened 2023-02-19
GitHub Releases BullMQ Automation

Ferotiq:
### Description

Adds functionality to send new GitHub releases (triggered from GitHub Webhook) to George.

Closes #


backend/src/api/routes/webhooks.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const router = Router();
+
+router.post(
+  "/releases",
+  RateLimit.webhooksPost,
+  validateRequest({
+    body: {
+      action: joi.string().equal("published").required(),
+      release: joi
```
Ferotiq, line 15: Adding the entire schema is probably best practice here but I didn't do that because it would take a while

Ferotiq:
I still need to verify that the request is coming from GitHub and not some rando

backend/src/api/controllers/webhook.ts
```diff
@@ -0,0 +1,26 @@
+import crypto from "crypto";
+import georgeQueue from "../../queues/george-queue";
+import { MonkeyResponse } from "../../utils/monkey-response";
+
+const secretHash = process.env.GITHUB_WEBHOOK_SECRET
+  ? crypto
+      .createHash("sha256")
+      .update(process.env.GITHUB_WEBHOOK_SECRET)
+      .digest("hex")
+  : null;
```
Miodec, line 10: I don't think it makes sense to generate a new secret when comparing to the one that's being sent from GitHub.

backend/src/api/controllers/webhook.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function sendRelease(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { "x-hub-signature-256": signature } = req.headers;
+  if (secretHash && signature !== `sha256=${secretHash}`) {
+    return new MonkeyResponse("Unauthorized", 401);
+  }
+
```
Miodec, line 19: Add a 500 here if the .env property is not set

backend/src/api/routes/webhooks.ts
Miodec: Thats fair, but i think joi will fail the validation if there are extra properties compared to whats specified. I think you need to allow extra properties?

backend/src/middlewares/rate-limit.ts
```diff
@@ -498,6 +498,14 @@ export const apeKeysGenerate = rateLimit({
   handler: customHandler,
 });
 
+// Webhooks Routing
+export const webhooksPost = rateLimit({
+  windowMs: ONE_HOUR_MS,
+  max: 120 * REQUEST_MULTIPLIER,
```
Miodec, line 504: 10 max

backend/src/api/controllers/webhook.ts
Ferotiq: this isn't regenerating, it's comparing the hashes, afaict github auto hashes the secret. I could be wrong tho

Ferotiq: ig the hash could just be stored in the env lol

Miodec: yeah nevermind i just read the github docs, keep it the way it is

backend/src/api/routes/webhooks.ts
Ferotiq: I used `.unknown()` which allows unknown properties.

PR #4009, opened 2023-02-19
Added 1 english quote

ThorsThunder04:
Quote from The World According To Garp

PS: It says I deleted all the lines then readded basically more lines. I just couldn't commit the file by it's self so I replaced it with the file containing the new quote since the file is too big to edit inside of github.


Miodec:
Sorry, don't like the content of this quote. Consider that a lot of kids use this site.

ThorsThunder04:
Fair enough

PR #4011, opened 2023-02-20
Add word filter feature

albertying:
### Description

  When enabled words that appear will either be left or right handed. Setting this to 'left' will make words appear with letters 'qwertasdfgzxcvb', while setting this to 'right' will make words appear with letters 'yuiophjklnm'.
  
 #### option left
image
image
image

 #### option right 
image
image
image

  I've never contributed to open source before, so sorry if I'm not doing something correctly.
  
  Closes #
  N/A


Miodec:
This feature is already available in the custom text words filter popup.

Please discuss issues with me before just opening a pr.

PR #4015, opened 2023-02-21
Add `rel="me"` attributes to links on profile page

smichaelsen:
This indicates, that the linked resources belong to the person, that the current page is about.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/me

### Description

Related discussion: https://github.com/monkeytypegame/monkeytype/discussions/4014

### DIFF
--- a/frontend/src/ts/elements/profile.ts
+++ b/frontend/src/ts/elements/profile.ts
@@ -143,7 +143,7 @@ export async function update(
         socialsEl.append(
           `<a href='https://github.com/${Misc.escapeHTML(
             git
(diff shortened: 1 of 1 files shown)


Miodec:
Im not sure if this is the right way of going about this since anybody can put any link in there?

smichaelsen:
It's what for example github.com does on profile pages. If you're worried about *link juice* you might want to expand it to `rel="nofollow me"`.

Miodec:
Sure, do that

Miodec (approved):

PR #4033, opened 2023-03-01
Add one quote to English

Dainternetdude:
### Description

Add one quote to English from the book _Strong Towns - A Bottom-Up Revolution to Rebuild American Prosperity_ by Charles L. Marohn, Jr. ISBN 978-1-119-564-81-2

PR #4034, opened 2023-03-01
added one new quote in English

Decipher-CS:
### Description

Added a single new English quote. Unknown source but allegedly it was said by Socrates.

### DIFF
--- a/frontend/static/quotes/english.json
+++ b/frontend/static/quotes/english.json
@@ -38651,6 +38651,12 @@
       "source": "Conduction Heat Transfer",
       "id": 6768,
       "length": 250
+    },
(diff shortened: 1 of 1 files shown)


Decipher-CS:
Should I adjust the `id` property by incrementing to [#4033](https://github.com/monkeytypegame/monkeytype/pull/4033)?

Dainternetdude:
i guess i should point out in case it's not intentional but you've already incremented it by 21

Decipher-CS:
> i guess i should point out in case it's not intentional but you've already incremented it by 21

Yea it wan an oversight. Thanks for letting me know.

PR #4036, opened 2023-03-01
eva_one theme added

Kutaygb:
### Description
I added a theme called 'eva_one', which is a reference to Evangelion Unit 1.

Screenshots:
image
image2


Miodec:
The colors are really harsh, too saturated. Also please don't use saturated colors for text color (close to white or black where possible). Also the sub-alt color choice is causing poor contrast.
image

Kutaygb:
Previously, I had employed colors with high saturation to impart a mecha-like atmosphere, but subsequently decreased the saturation of said colors to some extent.
updated1
updated2

Miodec:
I just realised this theme is visually very similar to the `cyberspace` theme. Sorry, it's unlikely this will be merged unless the whole color scheme completely changes.

PR #4038, opened 2023-03-02
Add layout gallium and angle variant

GalileoBlues:
Gallium has had a few users and seems to be gaining in popularity, It would be amazing if it could be added to monkeytype!


Dainternetdude (approved):
amazing layout

Miodec (approved):

PR #4058, opened 2023-03-09
Remove the (non-)word "tion" from the English 10k word list

Dainternetdude:
I have removed "tion" from the english 10k word list because I was not able to find a definition for it in any dictionary, and as a native English speaker I am not aware of any definition for it. I certainly do not believe that it is one of the ten thousand most used words.


Ferotiq (approved):
Yeah I have no idea how that ended up in the list

PR #4072, opened 2023-03-14
Added swedish with words all containing å, ä or ö.

IndigoWilke:
Put together a language for extra training on å, ä and ö, since they are often neglected on typing websites, they need some extra love.


Miodec:
Please rename to `swedish_accents`

PR #4103, opened 2023-03-22
Fix positions of pb and average lines when disabling start graph at 0

### LINKED ISSUE #4102: Disabling start graph at 0 will place average lines in at the wrong height

albertying:
### Description

Fix positions of pb and average lines when disabling start graph at 0.

Closes #4102

### DIFF
--- a/frontend/src/ts/pages/account.ts
+++ b/frontend/src/ts/pages/account.ts
@@ -799,8 +799,10 @@ function fillContent(): void {
 
   if (!Config.startGraphsAtZero) {
     accountHistoryScaleOptions["wpm"].min = Math.floor(minWpmChartVal);
-  } else {
-    accountHistoryScaleOptions["wpm"].min = 0;
+    accountHistoryScaleOptions["pb"].min = Math.floor(minWpmChartVal);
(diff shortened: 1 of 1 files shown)


frontend/src/ts/pages/account.ts
```diff
@@ -798,9 +798,12 @@ function fillContent(): void {
     Math.floor(maxWpmChartVal) + (10 - (Math.floor(maxWpmChartVal) % 10));
 
   if (!Config.startGraphsAtZero) {
-    accountHistoryScaleOptions["wpm"].min = Math.floor(minWpmChartVal);
-  } else {
-    accountHistoryScaleOptions["wpm"].min = 0;
```
Miodec, line 803: What about this else statement?

albertying: Since the datasets have default minimums of 0, I thought it was unnecessary. Should I add them back?

Miodec: Im thinking that switching this setting might not reset the min value to 0, so I think it would be safer to just always set it.

PR #4113, opened 2023-03-26
Update `dynamic` Keymap Legend Style to More Accurately Reflect Casing

Ferotiq:
### Description

This PR ensures that the keymap legend displays the correct casing of every key for every combination of CapsLock, Shift, and user platform. It also improves performance by eliminating code that runs for no reason.

Previously, when using the `dynamic` keymap legend style, pressing CapsLock would use the "capitalized" version of every key, and pressing Shift would do the same, but it would also effectively "invert" the state of CapsLock.

What I mean by that is that if you press CapsLock, every key "capitalizes," but then if you also hold Shift, it will go back to lowercase. As it turns out, Shift inverting the state of CapsLock does not happen on MacOS. So I provided an edge case to rule out MacOS from that behavior.

CapsLock also only capitalizes letters, not symbols, but Shift does both. So I have two different states managing the capitalization of letters and symbols.

Finally, if the user is using `oppositeShiftMode`, then the keymap will only capitalize what they can press. This includes a complete overhaul of `isUsingOppositeShift`.


Miodec:
`also removes the use of the keymap option for oppositeShiftMode, which I'm not sure why there's that distinction anyways, but I can likely add that support back in if it makes sense.`

Bring it back, please.

Ferotiq:
> Yeah, no - looks like you chopped some functionality.
> 
> Set your keymap to `colemak dh iso`, then `opposite shift keymap` - `shift + p` should be blocked, because `P` is on the left side on that layout. Right now, it isn't. (it working fine on live)

Can you send a video on Discord of that happening? The expected behavior is happening for me.

PR #4150, opened 2023-04-06
Updated Chinese Traditional's Word List and Order in _list.json and _groups.json 

JacoChong:
### Description
**Change in Word List:**
The original 270+ words in the word list were mostly uncommon words, now they are replaced with 200 most used words in Chinese Traditional (according to the Ministry of Education in Taiwan).

**Change of Order in _list.json and _groups.json:**
The original chinese language and it's language group are separated from other chinese languages and their groups, they are now moved down to be together with other chinese languages.


Miodec:
Would it make sense to put all of those list into one group? (chinese) So it would be `chinese_pinyin`, `chinese_ jyutping`, `chinese_traditional` ?

PR #4159, opened 2023-04-08
Added english quotes

zack-carnet:
### Description

Added some english quotes

Closes #


frontend/static/quotes/english.json
```diff
@@ -38990,6 +38990,24 @@
       "source": "Clockwork Princess",
       "length": 66,
       "id": 6854
+    },
+    {
+      "text": "Love cannot be explained, yet it explains all.",
+      "source": "The Forty Rules of Love",
+      "length": 46,
+      "id": 6855
```
Miodec, line 38998: Too short, please remove or use a longer version.

zack-carnet: I saw a few quotes which were less than 60 words so I figured It would be fine since this quote makes sense to have as it is. But I have replaced it with a longer version now.

PR #4167, opened 2023-04-10
Update German quotes

BlackSagittarius:
Hey Mio,
I now went through all the new German quotes and changed what I could find.
There still are some questionalble grammatical things in there but I guess that they are due to the original author and not because they are quoted wrong.

I hope I got them at least mostly and that this causes, for once, no further problems.

Sagittarius


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/4661623928

PR #4177, opened 2023-04-13
Added Swedish quotes

noamvokaal:
### Description

Added 40+ Swedish quotes and corrected two existing ones.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/4690376873

Miodec:
Probably remove 89, 93, 94, 99, 111, 114.

noamvokaal:
I just replaced quote 99 and 114 with harmless sentences from the same authors.

From Google Translate: 

```
       "text": "The speed that is worth anything is indeed only achieved by – slowness.",
       "source": "Vilhelm Ekelund",
       "length": 85,
       "id": 99

       "text": "It was a lovely time back then, when you drove around with your little tent and set it up, wherever the policeman looked good-natured and beat his drum and invited a blind old woman and a deaf old man to come up to the hatch to form a queue and entered and found the tent empty but lined below with children's heads with accompanying bodies outside, the merry family of peepers.",
       "source": "Hjalmar Bergman",
       "length": 342,
       "id": 114
```

Miodec:
I prefer to avoid religious content because people get really weird from both sides, saying either "I don't come to this website to worship god" or "This quote is against my beliefs" so i decided to just completely avoid them.

As for the other quotes - I get the metaphores but kids might not understand them. This needs to be considered when adding quotes - most users are teens.

noamvokaal:
I respect your opinion, but I do maintain that teenagers can indeed distinguish metaphors from literal meanings. Excuse my grumpiness.

Anyway, I realize that I have to abide by your rules, and have now replaced quotes 89, 93, and 94.

Google Translate: 
```
       "text": "The people of our time are old willow trees, which with rotten cores only vegetate with the bark. All our life lies on the surface.",
       "source": "Adolph Törneros",
       "length": 118,
       "id": 89

       "text": "No time is more difficult to use than a quarter of an hour stuck empty between fixed occupations. It does about as much mischief as a good crack in a wooden wall too small to fill up with a solid log, and for big to stuff even some moss.",
       "source": "Adolph Törneros",
       "length": 279,
       "id": 93

       "text": "Or you could compare October to a mermaid: one half a beautiful and warm girl, the other half an ugly and icy fish.",
       "source": "Zacharias Topelius",
       "length": 120,
       "id": 94
```

Miodec:
Thanks

Issue #4124, opened 2023-03-29
Make account graph buttons toggle buttons


monkeytypegeorge:

VibhorJaiswal:
I can take this up, can you please add some more details?

Miodec:
These buttons here, should have an active class when their respective data sets are visible
<img width="970" alt="image" src="https://user-images.githubusercontent.com/13181393/230598517-dee44ea5-de62-4f76-b56e-95f65e5655e6.png">

So it SHOULD look like this:

<img width="540" alt="image" src="https://user-images.githubusercontent.com/13181393/230598638-b3e8a5da-2c90-4028-9269-6d75ec44431e.png">

<img width="624" alt="image" src="https://user-images.githubusercontent.com/13181393/230598687-1a3c90f0-7819-4e6f-a2a3-f048dc7c9a3d.png">
<img width="656" alt="image" src="https://user-images.githubusercontent.com/13181393/230598740-7995fa8d-6caa-4f0f-8a92-60fcda1cf14d.png">

VibhorJaiswal:
Understood. I will take this.

Miodec:
@VibhorJaiswal any updates?

VibhorJaiswal:
Hi @Miodec , I got unexpectedly caught up with other commitments, so I can only work on it after 1-2 weeks. So if it's too late then you can assign it to someone else. I apologize for the delay.

D3press3dd:
HI, can i take this?

PR #4203, opened 2023-04-21
fix:#4124 graph buttons are now toggles

D3press3dd:
### Description
This pr fixes the problem with the buttons under chart, now they are toggles buttons with their respective color theme

Closes #

*This Pr fix is a fix to the issue [#4124](https://github.com/monkeytypegame/monkeytype/issues/4124)

*use the getConfigChanges function to get the updated values between bd and localstorage

*crea una nueva funcion para cambiar el estado de los botones con su tipado

### DIFF
--- a/frontend/src/ts/controllers/chart-controller.ts
+++ b/frontend/src/ts/controllers/chart-controller.ts
@@ -948,7 +948,8 @@ type ButtonBelowChart =
   | ".toggleAverage10OnChart"
   | ".toggleAverage100OnChart";
 
-function updateButtonChartsColors(
+
+function updateAccountChartButton(
   isActive: boolean,
   className: ButtonBelowChart
 ): void {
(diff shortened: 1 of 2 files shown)


frontend/src/ts/controllers/chart-controller.ts
```diff
@@ -53,7 +53,7 @@ Chart.defaults.elements.line.fill = "origin";
 
 import "chartjs-adapter-date-fns";
 import format from "date-fns/format";
-import Config from "../config";
+import Config, { getConfigChanges } from "../config";
```
Miodec, line 56: No need to use `getConfigChanges` - just use `Config`

Miodec (changes requested):

frontend/src/ts/controllers/chart-controller.ts
```diff
@@ -943,8 +943,26 @@ export const miniResult: ChartWithUpdateColors<
   },
 });
 
+type ButtonBellowChart =
```
Miodec, line 946: Bellow > Below

frontend/src/ts/controllers/chart-controller.ts
D3press3dd: Done, my mistake, I thought that Config was only a superficial copy and did not come from the database.

PR #4215, opened 2023-04-24
Add GDScript 2.0 programming language

JustTemmie:
### Description

GDScript support was added back in December thru https://github.com/monkeytypegame/monkeytype/pull/3792.
Since then Godot 4 has released, and with it they added GDScript 2.0, along with just general engine changes, this makes multiple of the "words" in the existing GDScript language irrelevant in newer versions. So this is an update to the previous PR, but since GDScript 1.0 is still in use by certain people I believe it would be wisest to add GDScript 2.0 as a separate language instead.

do note the language includes a full-stop (.) in the name, does monkeytype support this or not?

```
setget has been completely refactored, it doesn't really apply anymore so i'm removing it
yield() has been renamed to await()
spatial was renamed to node3D
kinematicbody was renamed to CharacterBody2D and 3D
Sprite has been renamed to Sprite2D
Camera has been renamed to Camera3D
```


Miodec:
> do note the language includes a full-stop (.) in the name, does monkeytype support this or not?

I would rename it to `code_gdscript_2`

JustTemmie:
alright sure, but it is referred to as gdscript 2.0 by the developer team and community so it might be a bit weird? unsure

PR #4223, opened 2023-04-25
Accessibility

D3press3dd:
### Fix some accessibility problems
*adds missing roles when using aria-label

*removes some excess "class" property that were unused

*add some placeholders to improve user experience

*add rel "noopener" to avoid tab access using

*change div to link to keep the structure of the others

*add -webkit-user-select to support safari browsers

### DIFF
--- a/frontend/static/html/warnings.html
+++ b/frontend/static/html/warnings.html
@@ -68,6 +68,7 @@
       background: #d1d0c5;
       color: #333;
       cursor: pointer;
+      -webkit-user-select: none;
       user-select: none;
     }
     .noCssReinstallServiceWorker:active {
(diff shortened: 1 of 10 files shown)


Miodec:
image

Can we remove placeholders if there is a title right above the input?

Miodec:
image

Or here

frontend/static/html/warnings.html
```diff
@@ -68,6 +68,7 @@
       background: #d1d0c5;
       color: #333;
       cursor: pointer;
+      -webkit-user-select: none;
```
Miodec, line 71: If we are going to add this we should add it everywhere

PR #4237, opened 2023-04-29
Added Lavelegant Theme

MarksASP95:
### Description

Just a proposition for a new theme: lavelegant (lavender + elegant). I love this site and would like to contribute with any little thing.

image


Miodec:
Sorry, I don't think the color choices work well together. Main issues are:
 - dark text color
 - error is the same as sub
 - sub alt is the same as bg

PR #4242, opened 2023-04-30
Add Scandinavian Layouts (QWERTY-Nordic)

torjacob:
### Description

Added Danish, Norwegian, and Swedish QWERTY layouts.

**Entries in the layout list named in this fashion:**

nordic_*\<[ISO 639-1][iso]\>*

[iso]: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes


Ferotiq:
I think they should just be named `norwegian_qwerty` `swedish_qwerty` and `danish_qwerty` to be consistent with the other languages with qwerty variations

PR #4260, opened 2023-05-05
Blood Meridian Quote

ibrab000:
Added a Blood Meridian quote. If you think 1938 characters is too long, feel free to deny or adjust it.


Miodec:
Not sure I really like this quote, sorry.

PR #4274, opened 2023-05-11
Added pace caret style -> off, background size, and background filter settings to the command line.

je-farr:
#### 1. Added pace caret style -> off.
This option was visible in the settings page, but absent from the command line. The setting is useful because it lets you see your pb from the typing page, without having to actually deal with the distracting pace caret.

image

#### 2. Added background size settings.
The background size settings from the settings page which were previously absent from the command line.

image

#### 3. Added background filter settings.
The background filter settings from the settings page which were previously absent from the command line.

image

#### 4. In config-validation.ts, numberArray is checked for NaN values.
To change the background filter I was passing a numberArray, but there was an unintended side effect that you were able to put in a NaN value and it wouldn't cause an error since NaN is technically a number. To fix that I made it so that numberArray cannot contain a NaN value.


frontend/src/ts/commandline/lists/background-filter.ts
```diff
@@ -0,0 +1,79 @@
+import Config, * as UpdateConfig from "../../config";
+
+const subgroup: MonkeyTypes.CommandsSubgroup = {
+  title: "Custom background filter...",
+  configKey: "customBackgroundFilter",
+  list: [
+    {
+      id: "setCustomBackgroundBlur",
+      display: "Blur...",
```
Miodec, line 9: Lowercase the display values in this file

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/4943687618

frontend/src/ts/commandline/lists/background-filter.ts
je-farr: done.

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/4979044415

PR #4291, opened 2023-05-15
Update user.ts

8xu:
### Description

Modified the HTTP status code returned to the user if the Captcha check failed from 422 _Unprocessable Content_ to 403 _Forbidden_.


Miodec (approved):

Miodec:
I don' think forbidden makes sense here. 403 would mean that the user does not have access to the endpoint they are trying to access.

PR #4294, opened 2023-05-16
added six italian quotes

magicteo:
I added 6 quotes from different italian books to the 'italian.json' file under in the static assets of the frontend folder. I made sure to follow the quote guidelines.

### DIFF
--- a/frontend/static/quotes/italian.json
+++ b/frontend/static/quotes/italian.json
@@ -333,7 +333,7 @@
     },
     {
       "text": "Oggi non è che un giorno qualunque di tutti i giorni che verranno. Ma quello che accadrà in tutti gli altri giorni che verranno può dipendere da quello che farai oggi..",
-      "source": "Hernet Hemingway",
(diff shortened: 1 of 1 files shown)


magicteo:
These are the translations of the quotes so you can verify their content:

112: She and Mattia were united by an elastic and invisible thread, buried under a pile of things of little importance, a thread that could only exist between two like them: two who had recognized their loneliness in each other.

113: If there is no love, not only does the life of people dry up, but also that of cities.

114: Why did I have to go to school? Why did the world work like this? You are born, you go to school, you work and you die. Who had decided that this was the right way? Couldn't we have lived differently? Like primitive men?

115: But by now I'd learned to ask the grown-up questions, where you ask one thing to find out another.

116: He takes one more look out the window, a very brief glimpse, for his last portion of the stars. Then in the dark, although no one sees him, he smiles.

117: Nomads know it: maps are not for orientation, but for dreaming of the journey in the months preceding the separation.

PR #4297, opened 2023-05-17
added more quotes

magicteo:
added more quotes to italian.json

you can verify them by checking all the traslations in the first comment below

### DIFF
--- a/frontend/static/quotes/italian.json
+++ b/frontend/static/quotes/italian.json
@@ -702,6 +702,36 @@
       "source": "Paolo Rumiz, Trans Europa Express",
       "length": 113,
       "id": 117
+    },
+    {
+      "text": "Hanno successo soltanto coloro che, nella vita, si rivolgono con gioia, entusiasmo e dedizione a ciò che fanno.",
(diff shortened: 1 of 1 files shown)


magicteo:
Translations:

118: Only those who, in life, turn to what they do with joy, enthusiasm and dedication are successful.

119: You are like a hummingbird because like the hummingbird you put all your energy into staying still. Seventy wing beats per second to stay right where you are. You manage to stop in the world and in time, you manage to stop the world and time around you, sometimes you even manage to go back up it, time, and find the lost one, just as the hummingbird is capable of flying backwards.

120: Murphy's law says that whatever has to happen, will happen. And it didn't seem like a bad thing at all to us.

121: The economic system from which salvation of the world is expected today is truly strange! No one manufactures anything with their own hands anymore, no one tries to make a pot, a flageolet or invent a cart; the most brilliant idea that one can have is to go somewhere on earth to buy something to resell elsewhere, at a profit.

122: In every race there is always the faceless friend who has exactly your pace. Be ready to leave the old for the new and focus on your goal: crossing the finish line with a smile on your face.

PR #4305, opened 2023-05-20
Add 'programme' to British english

KaitlynEthylia:
In British English, 'Programme' is the preferred spelling over 'Program'


Miodec:
This was used before but was reverted in favor if the "to program" meaning.

KaitlynEthylia:
Was this discussed at any point? I cannot find any issue relating to preferring that meaning, and even so, in my experience, programme is the preferred spelling even in the context of the verb

Miodec:
Mainly discord and email complaints from users

PR #4318, opened 2023-05-25
Added Italian Quotes

bsule:
Meanings of quotes by ids:

127: There are moments in life when the difference between living and surviving is subtle, imperceptible. But in reality there is a difference. It's the difference between those who believe in themselves and those who don't.

128: No matter where you are, no matter what you do, there is always a truth waiting to be discovered. And when you find out, your life changes forever.

129: We all have a childhood. What matters is how we leave it behind. I wanted to go on without forgetting. Forget, it's like childhood never existed.

130: You are not what you see in your mirror, you are not what you think you are, you are not even what others think of you. You are what you love and what you love will make you who you are.

131: And life, real life, is not what you see on television, with those fake smiles and those perfect stories. Real life is made up of ups and downs, successes and failures, joys and sorrows. But in the end, it is this real life that makes us stronger and gives us the strength to carry on.

132: It's not the fault of who you loved if that person left you. Love is a feeling that cannot be controlled, that cannot be managed, that does not stop. Love is an emotion that comes when you least expect it and that changes your life. Never regret having loved, even if it made you suffer, because love is the only thing that is truly worth living.

133: Life is a game and the game is hard. If you want to win you have to be the best, you have to be willing to do anything, you have to know how to fight and never give up. Life is a gamble and the gamble is high. But if you have the necessary grit and determination, in the end you will win.

134: There are two ways to live life: as a spectator or as a protagonist. The difference is that as a spectator you can only watch others do things, while as the protagonist you do them. Don't wait for life to give you everything, take what you need and never stop.

135: Don't be afraid to dream big, to fly high, to look far. Life is an adventure, a journey that never ends. Don't settle for what you have, but always look for more, better, different. Never stop, never give up, because life is too short to be mediocre.

136: Life is a zero-sum game, where every win is also a loss, where every success is also a failure. There are no absolute winners, because life is made up of ups and downs, joys and pains, successes and failures. But if you can accept your defeats and learn from your mistakes, then you have a chance to really win.

137: There is nothing more precious than freedom, the possibility of choosing one's own path, of deciding one's destiny. It doesn't matter how many difficulties you encounter along the way, how many times you fall, the important thing is to get up again and move forward. Never let anyone steal your freedom, because it's the only thing that really belongs to you.

### DIFF
--- a/frontend/static/quotes/italian.json
+++ b/frontend/static/quotes/italian.json
@@ -778,7 +778,7 @@
     {
       "text": "Non sei tu quello che vedi nel tuo specchio, non sei quello che pensi di essere, non sei nemmeno quello che gli altri pensano di te. Sei quello che ami e quello che ami ti renderà quello che sei.",
       "source": "Don Pietro Savastano, Gomorra",
-      "length": 175,
+      "length": 195,
       "id": 130
     },
     {
--- a/frontend/static/quotes/italian.json
+++ b/frontend/static/quotes/italian.json
@@ -756,6 +756,72 @@
       "source": "Giorgio Faletti, Tre atti e due tempi",
       "length": 226,
       "id": 126
+    },
+    {
+      "text": "Ci sono dei momenti nella vita in cui la differenza tra vivere e sopravvivere è sottile, impercettibile. Ma in realtà la differenza esiste. È la differenza tra chi crede in se stesso e chi no.",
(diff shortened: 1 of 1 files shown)


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5075548315

PR #4322, opened 2023-05-26
Zen mode - removes extra newlines if backspace ends up on a newline

seoulection:
### Description

Reproducing the issue:

1. Go to monkeytype.com and enter `zen` mode
2. Type some words, then press enter. This will put the cursor on a new line
3. Press backspace to get the cursor back on the previous line
4. Press space
5. Verify the cursor is on a new line

https://github.com/monkeytypegame/monkeytype/assets/6445410/58f86a54-1634-4e7d-9070-e72aee47c46a

The fix:

Pass a `boolean` into `updateWordElement` to determine if the function call came from `backspaceToPrevious`. If it did and the current character is a new line, set `newlineafter` to `false` and remove all existing `.newline` nodes which seems unnecessary for them to exist

https://github.com/monkeytypegame/monkeytype/assets/6445410/1e3dce74-cd2f-4858-9c70-bff6aecd4eb7

Feel free to close if this breaks other functionality that is unknown to me or if this is a non-issue

### DIFF
--- a/frontend/src/ts/controllers/input-controller.ts
+++ b/frontend/src/ts/controllers/input-controller.ts
@@ -137,6 +137,19 @@ function backspaceToPrevious(): void {
   Funbox.toggleScript(TestWords.words.getCurrent());
   TestUI.updateWordElement();
 
+  if (Config.mode === "zen") {
+    const els: HTMLElement[] = (document.querySelector("#words")?.children ||
+      []) as HTMLElement[];
+
+    for (let i = els.length - 1; i >= 0; i--) {
+      if (els[i].classList.contains("newline")) {
+        els[i].remove();
+      } else {
+        break;
+      }
+    }
+  }
+
   Caret.updatePosition();
   Replay.addReplayEvent("backWord");
 }


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5087557658

frontend/src/ts/controllers/input-controller.ts
```diff
@@ -135,7 +135,7 @@ function backspaceToPrevious(): void {
   TestUI.setCurrentWordElementIndex(TestUI.currentWordElementIndex - 1);
   TestUI.updateActiveElement(true);
   Funbox.toggleScript(TestWords.words.getCurrent());
-  TestUI.updateWordElement();
+  TestUI.updateWordElement(false, true);
```
Miodec, line 138: You're forcing `showError` to false here, which means that if you make an error in a word, space to the next one, then backspace to the word with the error, the error highlighting will dissapear. Pass `undefined` here, so it will default to the `blindMode` state.

frontend/src/ts/test/test-ui.ts
```diff
@@ -523,8 +526,15 @@ export function updateWordElement(showError = !Config.blindMode): void {
...
-        newlineafter = true;
-        ret += `<letter class='nlChar correct' style="opacity: 0"><i class="fas fa-angle-down"></i></letter>`;
+        if (!isFromBackspace) {
+          newlineafter = true;
+          ret += `<letter class='nlChar correct' style="opacity: 0"><i class="fas fa-angle-down"></i></letter>`;
+        } else {
+          newlineafter = false;
+          document
+            .querySelectorAll(".newline")
+            .forEach((newline) => newline.remove());
```
Miodec, line 536: You're removing all new lines - if there are multiple new lines on screen it will remove them all, instead of just one.

Miodec:
I think the solution here is to simply remove the last `newline` element inside `backspaceToPrevious` if in zen mode

frontend/src/ts/controllers/input-controller.ts
seoulection: Ah, thanks for the catch! Updated in https://github.com/monkeytypegame/monkeytype/pull/4322/commits/82af57be6ada1bcd69e99b5a4c9db61279e6f450

frontend/src/ts/test/test-ui.ts
seoulection: Updated in https://github.com/monkeytypegame/monkeytype/pull/4322/commits/a9ab175edd46c42a37db3a190b5bbdd91962cb16

Miodec:
> I think the solution here is to simply remove the last newline element inside backspaceToPrevious if in zen mode

Have you looked into this solution? It might be a bit cleaner as it won't require a new parameter to be added to `updateWordElement`

seoulection:
> > I think the solution here is to simply remove the last newline element inside backspaceToPrevious if in zen mode
> 
> Have you looked into this solution? It might be a bit cleaner as it won't require a new parameter to be added to `updateWordElement`

@Miodec Apologies, totally misunderstood that part. We'd essentially need to remove the last newline twice if we want to go this route because we would end up in this conditional block within `updateWordElement`:

```javascript
else if (TestInput.input.current[i] === "\n") {
  newlineafter = true;                     // ultimately adds a new line
  ret += `<letter class='nlChar correct' style="opacity: 0"><i class="fas fa-angle-down"></i></letter>`;
} 
```

So the changes inside `backspaceToPrevious` would look like:

```javascript
$(".newline").last().remove();

TestUI.updateWordElement();

$(".newline").last().remove();
```

If this is still preferred, I can push this up. Let me know your thoughts!

seoulection:
Actually, I don't think we'd want to remove any new lines within `backspaceToPrevious`. If a user types some words on line 1, starts typing more words on line 2, then presses backspace to get to the previous word, all of the words on line 2 will be moved to line 1.

Miodec:
How about this at the end of `backspaceToPrevious`, works well for me so far:

```ts
  Funbox.toggleScript(TestWords.words.getCurrent());
  TestUI.updateWordElement();

  if (Config.mode == "zen") {
    const els: HTMLElement[] = (document.querySelector("#words")?.children ||
      []) as HTMLElement[];
    for (let i = els.length - 1; i >= 0; i--) {
      if (els[i].classList.contains("newline")) {
        els[i].remove();
      } else {
        break;
      }
    }
  }

  Caret.updatePosition();
  Replay.addReplayEvent("backWord");
```

Miodec:
Thanks

PR #4324, opened 2023-05-26
added more italian quotes

magicteo:
Added four quotes to italian.json.
These are the translations:

138: How stupid we are, how many rejected invitations, how many unspoken words, how many unrequited looks. Many times life passes us by and we don't even notice it.

139: If in the world we all became scientists, who unloads the cash registers here at the markets?

140: You spend your life being told you're too young, then you're too old. There will be a middle stage where one has to run, right?

141: If we're not here too, they'll set up the republic for you in no time. If we want everything to stay as it is, everything must change.

### DIFF
--- a/frontend/static/quotes/italian.json
+++ b/frontend/static/quotes/italian.json
@@ -822,6 +822,30 @@
       "source": "Guido, La vita è bella",
       "length": 365,
       "id": 137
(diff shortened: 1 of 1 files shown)


Miodec:
Thanks

PR #4334, opened 2023-05-30
Add "Greeklish"

Invades:
### Description
This PR adds "Greeklish" and the 1k, 5k, 10k, 25k variants of it.
Greeklish is a common dialect in Greece used online when chatting because it saves time by not having to constantly switch language layouts (and also helps people that can't spell). It's original use case was on old computers when people were forced to use ASCII characters to send E-Mails.

### Additional info
There are slightly different variants of it (an example being using numbers for letters, like Θ->8) that people use but I believe this should be the most consistent way of typing it that makes sense to most Greek people.

#### Notes
I still think it is a bit silly to practice it considering it's not proper Greek and Lazy Mode exists for no accents but someone asked for it so I thought I'd add it. If there is anything I missed please let me know.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5119196909

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5119433203

Miodec (approved):

Miodec (approved):

Miodec:
Thanks

PR #4356, opened 2023-06-07
Adding Arabic

m7mdtiger:


frontend/static/languages/_list.json
rizwanmustafa, line 1: Please revert the changes on this file. It is going to make the reviewing difficult.

rizwanmustafa (changes requested):

frontend/static/languages/arabic.json
rizwanmustafa, line 1: Revert the changes on this file as well.

frontend/static/languages/_list.json
m7mdtiger: Thank you for your comment. I've just added "arabic_1k".
The remaining part involves adjusting the formatting of the document by adding a trailing comma at the end and not to the beginning like: 
[ 
  "english"
  ,"english_1k"
  ,"english_5k"
]

Is it still better to revert the changes !?

frontend/static/languages/arabic.json
m7mdtiger: The previous file contains vowel marks and we don't write them on typing!

frontend/static/languages/_list.json
rizwanmustafa: Yes

frontend/static/languages/arabic.json
rizwanmustafa: Please create a separate PR for that. An ideal PR addresses a single issue.

m7mdtiger: Alright, thank you.

frontend/static/languages/_list.json
m7mdtiger: Thanks. I've reverted the changes.

Miodec:
Does this list include diacritics? (which then later can be removed by enabling lazy mode)

m7mdtiger:
> Does this list include diacritics? (which then later can be removed by enabling lazy mode)

Yes, that should so it for the arabic.json list, which I have already reverted. What remains in this PR is the addition of the arabic_1k.json list, which contains no diacritics.

Miodec:
All lists should contain diacritics so that the users have the option to either keep them or remove them by enabling lazy mode. Please modify the list to include diacritics.

Miodec:
Still, thats what lazy mode is for. Some people may want to type them. Please add them.

m7mdtiger:
> Still, thats what lazy mode is for. Some people may want to type them. Please add them.

Than you. But then the default mode for Arabic should be the lazy mode for all users. Additionally, it's important to consider that the usage of diacritics varies based on the grammatical context, which would require additional time to prepare the list. It may be necessary to allocate a certain percentage for each context, but this task might need more time that I don't have right now!

PR #4357, opened 2023-06-07
Streak hour offset

Miodec:
The user can now add an offset to the streak reset date (which they can do only once) to make sure their streaks reset around midnight their local time.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5243648398

PR #4358, opened 2023-06-07
Fixed wrong scale when displaying accuracy average line chart

eran-pinhas:
The scale of the accuracy average over time(on the last 10/100) line chart wasn't fixed at 0-100 which caused a scale issue in some cases.

Before - the lower line is too low [about 0.25x of what it should have been]:

image

After:

image


Miodec (approved):
Nice, thanks

PR #4368, opened 2023-06-11
Add Speed Variable for Smooth Caret

Im-0xea:
### Description

Added a configurable value for the speed with which the caret moves when smoothCaret is set, this is something I have been missing as I focus alot on the caret and often the animation just feels to unresponsive to me and the unsmoothed caret just seams painful to look at.

I don't know if this kind of setting is desired, I could not find any issues asking for it besides one which mentioned this in a side note which didn't get addressed.

I did my best to find all places where variables need to be intigrated, I hope I didn't miss anything.

### DIFF
--- a/backend/src/api/schemas/config-schema.ts
+++ b/backend/src/api/schemas/config-schema.ts
@@ -27,6 +27,7 @@ const CONFIG_SCHEMA = joi.object({
   showLiveWpm: joi.boolean(),
   showTimerProgress: joi.boolean(),
   smoothCaret: joi.boolean(),
+  smoothCaretSpeed: joi.number().min(0),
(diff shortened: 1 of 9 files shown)


Miodec:
I think this is approaching the "whats the point of this setting" / "way too specific" category. I had zero complains about the smooth caret and i really want to avoid settings which will only be used by a very small percentage of users. The settings page is already huge. Sorry.

Miodec:
However: since the current caret setting is an `on/off` type, i would be okay with changing it to `off/slow/medium/fast`. (you would need to make sure that users on the old format will correctly be converted to the new one)

Im-0xea:
absolutely understandable I was thinking that might be the case :,)

oh, that sounds like a wonderful idea!, I will give it a shot!

PR #4399, opened 2023-06-23
Adding MV Terminal theme

KenGuru-blog:
Description

I created a theme that looked like the Terminal keycaps by MVKB for myself and was asked to upload it, so here it is!
In my opinion, it looks great in both normal and colorful mode.

Normal mode:
normal

Colorful mode:
colorful

Errors in normal and colorful mode:
errors normal
errors colorful

Closes #


Miodec:
Sorry, looks exactly like the `termnial` theme thats already available, just with slightly darker background.

KenGuru-blog:
In my opinion, it does share the similarity of the main color being green and the bg being dark, but that's actually it. The hue and saturation of the green is a fairly different one and also the orange-yellow caret and errors in combination with the darker background create a quite different color combination. And if you look at Material, Modern Dolch and Viridescent, they're also very similar with a different cream tone as their bg, but they just represent other sets.
In addition, i think you would make a lot of people happy with the theme because the set is well known in the community.
It's not me to decide, just my thoughts :)

PR #4400, opened 2023-06-24
added two english quotes

paperthinc:
### Description

Add two english quotes, taken from The Witcher books by Andrzej Sapkowski.


Warrobot10 (changes requested):
Looks good, but can you please explain what the back slash's are, as they are not present in any other quote.

Miodec:
Sorry, I don't really like either of those quotes.

PR #4404, opened 2023-06-26
Notify users for school email address (.edu | .ac | .sch)

### LINKED ISSUE #4393: Notify users if they are trying to use an .edu email account
Educational email addresses very often don't allow for receiving emails from outside the school network. Notify the user that they might not receive emails from monkeytype if they continue

aviiciii:
### Description
This pull request adds a new feature to the sign-up process that checks if the provided email address is a school email address (.edu domain). It also includes an indicator to notify users to consider using a personal email address for better email functionality.

### Demo

https://github.com/monkeytypegame/monkeytype/assets/88141521/2e2b3ab2-6e26-41fd-8758-8d1e05adca5e

Closes #4393

### DIFF
--- a/frontend/src/ts/pages/login.ts
+++ b/frontend/src/ts/pages/login.ts
@@ -86,13 +86,13 @@ const checkEmail = (): void => {
 
   const email = $(".page.pageLogin .register.side .emailInput").val() as string;
   if (emailRegex.test(email)) {
-    emailIndicator.show("valid");
-    // Check if email is .edu domain
-    if (email.endsWith(".edu")) {
+    if (/@.*(education|\.edu$|\.edu\.)/i.test(email)) {
       emailIndicator.show(
         "edu",
         "Consider using a personal email address for better email functionality."
       );
+    } else {
+      emailIndicator.show("valid");
     }
   } else {
     emailIndicator.show("invalid");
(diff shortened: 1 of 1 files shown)


aviiciii:
Hey, I added university domain (.ac) and school domains (.sch). 
note that .ac is a tld and a second-level domain whereas .sch is second-level domain only.

frontend/src/ts/pages/login.ts
```diff
@@ -41,7 +41,7 @@ export function hidePreloader(): void {
 export const updateSignupButton = (): void => {
   if (
     nameIndicator.get() !== "available" ||
-    emailIndicator.get() !== "valid" ||
+    (emailIndicator.get() !== "valid" && emailIndicator.get() !== "edu") ||
```
Miodec, line 44: Remove. I think we should still allow the signup if the user wants to use that email

Miodec (changes requested):

frontend/src/ts/pages/login.ts
aviiciii: (emailIndicator.get() !== "valid" && emailIndicator.get() !== "edu") 
this simply says that it must either be "valid" or "edu".

aviiciii: @Miodec ?

Miodec:
Thanks

frontend/src/ts/pages/login.ts
Miodec: Yeah youre right

PR #4406, opened 2023-06-26
Added quotes for Tamil Language

aviiciii:
### Description

- init a json file for tamil quotes
- added some quotes :)


Miodec:
These might be a bit short, usually I require quotes to be above 60 characters. Could you find some longer ones / use longer versions? Also please add some translations so i can verify content.

Miodec:
Thanks

PR #4410, opened 2023-06-27
Add English quote

asd153866714:
### Description

Add 1 English quote.


Miodec:
I generally avoid quotes that mention God just because people get very defensive on their beliefs / non beliefs. Sorry.

PR #4411, opened 2023-06-27
added quote

remalucard:
### Description

Added a quote.


Warrobot10 (changes requested):
There is a typo in length. Unless this is fixed, the code will not work.
The minimum length is also 60 characters, so please change or remove the quote.

PR #4417, opened 2023-06-28
Improve word highlight UI in input history

ademirkan:
### Problem
* Mouse blocks wpm speed while hovering words in input history
monkeytype-wpm-bad

### Solution
* Moved wpm speed above word so that it is not blocked by mouse
* Changed color to sub-alt, cleaner imo
monkeytype-wpm-good

### Notes
* Changed .wordInputAfter to .wordInputHighlight base class, added .wordInputHighlight--WithSpeed, which can be omitted for highlights not using speed
* Here is the only other time .wordInputAfter was being used. As you can see, it is working perfectly
monkeytype-multi-highlight

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -280,9 +280,6 @@
   margin: 0.25em;
   color: var(--sub-color);
   font-variant: no-common-ligatures;
-  // display: flex;
-  // transition: 0.25s
-  /*   margin-bottom: 1px; */
   border-bottom: 2px solid transparent;
   letter {
     display: inline-block;
@@ -299,8 +296,6 @@
   .wordInputHighlight {
     opacity: 1;
     white-space: nowrap;
-    display: flex;
-    flex-direction: column-reverse;
     position: absolute;
     background: var(--sub-color);
     color: var(--bg-color);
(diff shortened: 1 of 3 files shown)


Miodec:
Ooooorrrrr.... we could keep the speed below the word but make the cursor invisible when hovering over a word? What do you think?

ademirkan:
> Ooooorrrrr.... we could keep the speed below the word but make the cursor invisible when hovering over a word? What do you think?

Miodec:
@ademirkan why close the PR?

ademirkan:
> @ademirkan why close the PR?

oof, misclick. Was trying to close the reply.

ademirkan:
how can I open it back up?

ademirkan:
> Ooooorrrrr.... we could keep the speed below the word but make the cursor invisible when hovering over a word? What do you think?

This is what I first tested out. It looked good and solved the problem but it felt jank to me personally. If you append cursor:none to .word in inspect element, you can see what it would look/feel like. If you feel that it makes more sense, I can make the changes

Miodec:
It miiight be a tiny bit confusing at first, but it definitely improves the readability of the word and speed with absolutely nothing in the way.

ademirkan:
Sure, I can give it a go. Do you want to keep the speed color change from bg to sub-alt?

Miodec:
Yeah, keep that

ademirkan:
Also, thanks for the css feedback :) I knew my naming was jank but I didn't know the Sass way to fix it.

Miodec:
Thanks

PR #4420, opened 2023-06-28
Add short bulgarian quotes

asynchroza:
### Description

Adding a list of short quotes written in Bulgarian


Miodec:
Could you translate them for me so i can verify their content?

asynchroza:
1. Life isn't about waiting for the storm to pass, it's about learning how to dance in the rain.
2. Believe in your abilities. You were made for something greater than you can imagine.
3. The world is a book, and those who do not travel, read only one page.
4. Love knows no boundaries. It goes through everything and everywhere.
5. Happiness is not found in climbing the peaks, but in walking among the flowers on the road.
6. To understand yourself, you have to look into other people's eyes.
7. After every ending there is a new beginning. Never give up until you reach your goal.
8. It's never too late to become what you could have been.
9. Life is what happens while we are planning other things.
10. Never be ashamed of your own skills and talents. The world needs them.
11. There are days when tears are your only way to express what you feel.
12. The greatest victories are not those that everyone sees, but those that only the heart feels.
13. Don't overlook the meaning of the little things. They often contain the greatest beauty.
14. True freedom is not found in the outer world, but in inner liberation.
15. Think positive, act positive, and life will return positive results.
16. Everything we need is already within us. We just have to find the power to make it happen.

PR #4427, opened 2023-06-29
Email typo

### LINKED ISSUE #4392: Notify users if they are trying to sign up with a typo in their email
Find a list of common email typos (@gamil.com, stuff like that) and notify they might have a typo

Warrobot10:
Closes #4392. This is a draft, basically, I need to add more to the json file I created and give the json a look over and then I need to build monkeytype for testing.

### DIFF
--- a/backend/src/api/controllers/user.ts
+++ b/backend/src/api/controllers/user.ts
@@ -360,6 +360,11 @@ export async function getUser(
     inboxUnreadSize = _.filter(userInfo.inbox, { read: false }).length;
   }
 
+  if (!userInfo.name) {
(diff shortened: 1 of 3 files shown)


sanidhyas3s:
The current expected behavior is simply not accepting typos? If yes, then don't think this is acceptable because a typo-like domain might yet actually exist (as a joke or otherwise), and the users id might actually be legit email.

Also, the json should be grouped into domains of which they are typos of, and then suggest in a way like "the email gmial.com looks like it might be a mistake, did you mean gmail.com". This needs to be just a warning and the user should still be able to ignore this and move ahead if he wants to.

PR #4448, opened 2023-07-10
New Highlighting System

ademirkan:
# OVERVIEW
Created "highlight-controller" in order to introduce animations and abstract away all highlighting logic to a single controller. Highlight-controller has the following public functions:
### 1. `highlightWords(firstWordIndex: number, lastWordIndex: number)`
* highlights all user input words in "#resultWordsHistory .words" that are in range `firstWordIndex <= i <= lastWordIndex`
* first time this is called, it will call an internal function called `init()`. This does some preliminary processing of the ".words" in DOM

### 2. `clear()`
* clears any existing highlights, ie. when the user stops hovering over the chart.

### 3. `destroy()`
* resets all data processed by init(). This should only be done when positions of words change, such as zooming, toggling burst heatmap, changing browser size, etc.
* after this is called, next `highlight()` will call `init()`

### 4. `setIsHoverChart(state)`
* helper function to set state of whether or not user is hovering the chart (needed to prevent certain bug)

<br/><br/>

# WHY
### 1. Improved UI/UX
Before after of a simple highlight
image

### 2. Animations
(before after of moving highlights)

https://github.com/monkeytypegame/monkeytype/assets/70380876/8ab59e87-812c-4b5d-89e9-a20664c38a24

https://github.com/monkeytypegame/monkeytype/assets/70380876/4bed775a-7dcd-49ea-aad3-c68834b37d00

### 3. Performance
A single jQuery query traverses every node in the DOM with O(n) worst-case time complexity, where n is the number of elements in the DOM. This is quite expensive. The current system for highlighting elements uses jQueries a lot... like A LOT.

```typescript

                // Functions to log DOM updates & jQuery calls
                let numDomUpdates = 0;
                function incrementAndLogDomUpdate(numberOfUpdates: number){
                  numDomUpdates += numberOfUpdates
                  console.log("NUM DOM UPDATES SO FAR: " + numDomUpdates);
                }
                
                let numJQueryCalls = 0;
                function incrementAndLogJQueryCalls(){
                  numJQueryCalls += 1
                  console.log("NUM JQUERY CALLS SO FAR: " + numJQueryCalls);
                }

                ...
                ...
                ...

                // $(".wordInputAfter").remove();
                let query = $(".wordInputAfter")                // QUERY HERE
                incrementAndLogDomUpdate(query.length)
                incrementAndLogQueryUpdate()
                query.remove()                                  // DOM UPDATE HERE

                const wordsToHighlight =
                  TestInput.keypressPerSecond[parseInt(ti.label) - 1].words;

                const unique = [...new Set(wordsToHighlight)];
                unique.forEach((wordIndex) => {
                  const wordEl = $(
                    $("#resultWordsHistory .words .word")[wordIndex]
                  );
                  incrementAndLogQueryUpdate()                // QUERY HERE
                  const input = wordEl.attr("input");
                  if (input !== undefined) {
                    wordEl.append(
                      `<div class="wordInputAfter">${input
                        .replace(/\t/g, "_")
                        .replace(/\n/g, "_")
                        .replace(/</g, "&lt")
                        .replace(/>/g, "&gt")}</div>`
                    );
                    incrementAndLogDomUpdate(1)               // DOM UPDATE HERE
                  }
                });
```

Above are the current jQuery queries being used to highlight words. If we log the number of DOM manipulations and jQuery calls, we can see how much work is being done behind the scenes.

30_second_logs

Hovering a chart of a 60 second test averaging ~100wpm in its entirety from left to right, we get the following stats:
`Number of jQuery queries: 7362`
`Number of DOM updates: 10390`

This is not a typo. Its partially this high because the Chart.js callback seems to be called 6 times each hover. If this seems unbelievable, I suggest running the same experiment and seeing for yourself. Although the browser does a good job optimizing everything behind the scenes, these queries still are a cause for concern:

<p align="center">
image
</p>

How about HighlightController? How many expensive jQuery calls does it make?
The answer is JUST 2. 
Number of DOM updates: 124.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5510415654

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5510459632

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5525539306

ademirkan:
Few changes made since pull request:
* Fixed bug that exists in both new and old highlight systems: if user moves off chart and hovers new node in chart nearly at the same time, highlight persists even while mouse is off chart (this happens often with small browser width, or 2 minute + runs, which lead to chart nodes being clustered together very close to one another.

* Changed transition from 0.15s to 0.25s, making it more "buttery". Anything more feels slow rather than buttery, 0.25s is a good sweet spot.

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5539008807

ademirkan:
Finally happy with the result. Introduced optimizations. Fixed line-up issues. Removed all jank code left from prototype.

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5594422025

Miodec (approved):

Miodec:
Thanks

ademirkan:
> Thanks

my pleasure, thanks for the constant feedback!

PR #4460, opened 2023-07-17
Remember language choice before switching to quote mode

### LINKED ISSUE #4455: Remember language choice before switching to quote mode
If you are using english 1k and then switch to quote mode, it will automatically switch to english. But when going back to words or time mode, the website will not switch back to english 1k.

ridemountainpig:
### Description

Remember language choice before switching to quote mode. closes #4455 

https://github.com/monkeytypegame/monkeytype/assets/92412722/d65e1863-c929-4453-941e-3dd758824e0a

### DIFF
--- a/frontend/src/ts/test/test-logic.ts
+++ b/frontend/src/ts/test/test-logic.ts
@@ -483,7 +483,10 @@ export async function init(): Promise<void> {
 
   let language;
   try {
-    if (Config.mode !== "quote" && Config.language === Config.timeLanguage.split("_")[0]) {
+    if (
+      Config.mode !== "quote" &&
+      Config.language === Config.timeLanguage.split("_")[0]
+    ) {
       Config.language = Config.timeLanguage;
     }
     language = await Misc.getLanguage(Config.language);
--- a/frontend/src/ts/test/test-logic.ts
+++ b/frontend/src/ts/test/test-logic.ts
@@ -484,14 +484,7 @@ export async function init(): Promise<void> {
 
   let language;
   try {
-    if (
-      Config.mode !== "quote" &&
-      Config.language === languageBeforeQuoteMode.split("_")[0]
-    ) {
-      Config.language = languageBeforeQuoteMode;
-    }
     language = await Misc.getLanguage(Config.language);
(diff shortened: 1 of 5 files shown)


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5585213878

Miodec:
Adding a completely new config entry is overkill in this case. Instead, we can just remember it in memory. Take a look at the `init` function inside `test-logic.ts`. On line `523` it sets the language to the group name. There you can remember the current language before the change. Then, when initialising the test in any other mode, we can check this variable and set the language accordingly.

Miodec:
Yeah, just make a quick `languageBeforeQuoteMode` variable, or something like that

ridemountainpig:
If save in memory, when the user quit at quote mode and come back again to time mode, the before language will not be save, is this all right ?

Miodec:
Yes, thats okay though.

frontend/src/ts/test/test-logic.ts
```diff
@@ -483,7 +484,14 @@ export async function init(): Promise<void> {
 
   let language;
   try {
+    if (
+      Config.mode !== "quote" &&
+      Config.language === languageBeforeQuoteMode.split("_")[0]
+    ) {
+      Config.language = languageBeforeQuoteMode;
+    }
```
Miodec, line 492: I would move this below, add an `else` to the `if (Config.mode === "quote")`

Miodec (changes requested):

frontend/src/ts/test/test-logic.ts
ridemountainpig: Like this ?
image

Miodec: Yes, but instead of setting the `Config.language`, just use `setLanguage` (if the language before quote mode is an extended list and is of the same group as the current group)

ridemountainpig: If move the code to the `else ` of `if (Config.mode === "quote")` it will change like this, need to get the language again.

image

PR #4465, opened 2023-07-19
Added crt mode to the funbox

fehmer:
This PR adds a CRT mode to the funbox which emulates an old CRT monitor by adding scanlines, a background glow. The mode does not effect the colors from the theme.

Theme Serika dark:
image

Theme Serika:
image


Miodec:
Aaaand it also overlaps any custom background images you have.

fehmer:
> Styling on the center content element causes overlap of the timer bar. And I don't really like the left and right borders that it adds.
> Aaaand it also overlaps any custom background images you have.

good point, Moved the screen glow to the body which gets replaced when using custom background image. Also removed the borders.

Miodec:
There is a white flash coming from somewhere when restarting the test

Miodec:
The scanlines dissapear when leaving the typing page.

fehmer:
fixed

Perseus333 (approved):
I really like this theme. It should be in the main website.

frontend/src/styles/core.scss
```diff
@@ -532,3 +533,58 @@ key {
...
+  height: 100px;
+  z-index: 1001;
+  position: absolute;
+  bottom: 100%;
+  opacity: 0.1;
+  pointer-events: none;
+}
+
+body.crtmode {
+  text-shadow: 3px 0 1px rgba(0, 30, 255, 0.5), -3px 0 1px rgba(255, 0, 80, 0.3),
```
Miodec, line 548: I think this shadow maybe should be based on the theme colors and not hard coded? Right now I think it only fits some themes but not others.

PR #4483, opened 2023-07-27
Added 5 quotes to hiragana_quotes

ASasamori:
### Description

Added a quotes JSON for Japanese in Hiragana. Made 5 quotes from popular songs and authors, and while the original quotes use a mix of Katakana, made the quote only in Hiragana for the purpose of Monketype.

Closes #


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5675422574

Miodec:
Please fix the length field. Also i feel like if we add a new quote file we should start with a few more quotes.

ASasamori:
How many quotes would you like to start the new JSON file? Also, would there be a way to start a new language that combined Hiragana/Katakana/Kanji? Right now on Monkeytype, there are only 2 options for the Japanese language, Hiragana and Katakana. Phonetically there are like only about 2 letters that are different, so it's a little weird that they are split up. In Japanese in general the 3 alphabets are commonly intertwined.

Miodec:
Oh yeah, just rename the file to `japanese`

ASasamori:
Well, in that case then, a new language might have to be created because there is no cumulative Japanese language setting in monkeytype that includes all 3 alphabets (Hiragana, Katakana, and Kanji). Should I create that as a language instead of creating a quotes JSON beforehand?

Also how many quotes should I write before an initial creation of the quotes JSON file?

Miodec:
It should automatically switch to it, if it doesnt ill fix that later.

Im not sure about the amount of quotes to be fair. Defnitely more than 5, just so they dont repeat so much.

ASasamori:
What do you mean it should automatically switch to it? Like regardless if I have "japanese hiragana" or "japanese katakana" selected in the "words" or "time" mode, upon switching to quotes, the language should change to just 'japanese'? I would be interested in changing this implementation if possible, not just quotes.

Miodec:
Yes, quote mode will look for a file which is the same as the name of the langage group. For example, if you are in english_10k, switching to quote mode will set that language to english.

Miodec:
Closing due to inactivity.

PR #4488, opened 2023-07-31
refactor funbox

fehmer:
### Description

Funbox refactoring

Refactoring of custom css rules and loading of custom css files to reduce the amout of redundant code and have a better code sparation.

1. add funbox mode as class

Remove need for `applyConfig` to set classes on e.g. `#words` and the `Funbox.activate` and `Funbox.clear` methods to cleanup individual funboxes.

Replace this by setting the funbox modes as classes on the body element (e.g. funbox `arrows` adds class `fb-arrows`). The css can now use `body.fb-arrows #words ` instead of `#words.arrows`. 

In addition the body class attribute is only altered if there was a change in active funboxes. Previously this was set on each test restart causing "flickering" in some cases.

2. loading of funbox css

Currently there is a method `applyCSS` on funboxes with needs a custom css file to be loaded. Each of the methods contain the same code. Replace this with  `FunboxMetaData.hasCSS: boolean`. The css to be loaded is always `funbox/${funbox.name}.css` so we don't need to specify the path it in each `applyCSS` method.

In addition the custom css href is only altered if there was a change in active funboxes. Previously this was set on each test restart causing the css to be loaded each time and also causing "flickering" in some cases.

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -304,17 +304,6 @@
   &.tape .word {
     margin: 0.25em 0.5em 0.75em 0;
   }
-
-  &.nospace .word {
-    margin: 0.5em 0;
-  }
-
-  &.arrows .word {
-    margin: 0.5em 0;
-    letter {
-      margin: 0 0.25em;
-    }
-  }
 }
 
 .word {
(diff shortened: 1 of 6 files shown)


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5718914688

Miodec (changes requested):
image

Funbox CSS is not removed when leaving the test page

Miodec (approved):
Thanks

PR #4499, opened 2023-08-02
ci(frontend/scripts): print expected quote length

gforien:
### Context

CI prompts when a quote length is incorrect but does not fix it, or print how to fix it.

image

### Proposed change

Print the expected length.

image|700


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5739463064

Miodec (approved):
Nice, good idea.

Issue #4489, opened 2023-07-31
Add an option to revoke all access tokens to the danger zone


monkeytypegeorge:
In case the user wants to deauth all sessions, make a request to the backend and use the admin api to revoke all tokens

Warrobot10:
I would be interested in doing this.

Ferotiq:
@Warrobot10 you still interested in doing this or can I take over?

Warrobot10:
You can take over. Something weird is happening, and I can't commit to my fork. I just don't really understand the backend systems either. Thanks!

PR #4503, opened 2023-08-05
Add log out all sessions button to dangerzone

Warrobot10:
Adding a button to dangerzone to sign out all sessions. This is for #4489. The only reason I am posting this is because I am looking for some testing and suggestions. Thanks all!


Warrobot10:
I'm kinda stuck right now. Looks like I'll need to make a new endpoint, and I honestly have no idea how to do that. Thanks!

Miodec:
> I'm kinda stuck right now. Looks like I'll need to make a new endpoint, and I honestly have no idea how to do that. Thanks!

`backend/src/api/routes/users.ts`, thats where the endpoints are added. Just duplicate one of them and change whats needed. Routes call controllers - you need to add a controller function that will use the firebase auth api to deauth all tokens.

image

Ideally the function that does this will be located in `backend/src/utils/auth.ts` and the controller will just call that.

Let me know if you need hep.

Warrobot10:
Thanks!

PR #4510, opened 2023-08-07
Fixed rainbow trail theme to support custom backgrounds

fehmer:
### Description

Rainbow trail theme is supposed to hide correct words after a while. Previously this was done by setting the color to the background color at the end of the animation. With a custom background set the words will still be visible.

The fix will set the opacity to 0 instead of setting the text color to the background color to hide the text.


Miodec (approved):
Thanks

PR #4511, opened 2023-08-07
Fix minWpm on accounts history graph, set ticks on accounts activity …

fehmer:
### Description

Fix  scaling issues on accounts page:

- history graph minWpm must fit into the ticks step size
- activity graph should use the same ticks step size as the history
- reduce number of buckets in histogram for wph


Miodec (approved):
Thanks

PR #4514, opened 2023-08-07
Add Hack Nerd Font.

ojas-bhagavath:
I use the Hack Nerd Font with the custom fonts option in monkeytype, and I think it looks cool, so I have added Hack Nerd Font.


refact0r:
Why Hack Nerd Font instead of the normal Hack font? Nerd Font symbols and icons don't really serve a purpose in Monkeytype and using the original font would be less confusing.

ojas-bhagavath:
Noted, will add Hack font itself.

ojas-bhagavath:
> Why Hack Nerd Font instead of the normal Hack font? Nerd Font symbols and icons don't really serve a purpose in Monkeytype and using the original font would be less confusing.

But there are differences between these two fonts, aesthetically, for example '0' in the Hack Nerd Font has a dot in the middle, and Hack Font has a line instead.

Miodec:
Thanks

PR #4527, opened 2023-08-12
fix(themes): invisible popup text in rainbow trail

ridemountainpig:
### Description

The aria-label couldn't see the text in rainbow trail theme when hover.

Before:

CleanShot 2023-08-08 at 20 51 49@2x
</br>
CleanShot 2023-08-08 at 20 52 01@2x
</br>
CleanShot 2023-08-08 at 20 52 11@2x
</br>
CleanShot 2023-08-08 at 21 01 24@2x
</br>
CleanShot 2023-08-12 at 15 56 05@2x

After:

CleanShot 2023-08-08 at 21 18 26@2x
</br>
CleanShot 2023-08-08 at 21 18 40@2x
</br>
CleanShot 2023-08-08 at 21 18 53@2x
</br>
CleanShot 2023-08-08 at 21 01 44@2x
</br>
CleanShot 2023-08-12 at 15 55 32@2x

### DIFF
--- a/frontend/static/themes/rainbow_trail.css
+++ b/frontend/static/themes/rainbow_trail.css
@@ -31,8 +31,7 @@
 
 #top .logo .top, 
 .account .levelAndBar .level,
-div.textButton::after,
-span.textButton::after {
+[aria-label][data-balloon-pos]::after {
   -webkit-text-fill-color: initial;
 }
 
--- a/frontend/static/themes/rainbow_trail.css
+++ b/frontend/static/themes/rainbow_trail.css
@@ -29,7 +29,13 @@
   animation: rainbow-gradient 30s alternate ease-in-out infinite;
 }
 
-#top .logo .top {
+#top .logo .top, 
+.account .levelAndBar .level,
+#copyWordsListButton:hover,
+#toggleBurstHeatmap:hover,
+#playpauseReplayButton:hover,
+#editTagsButton:hover,
+#deleteAlertButton:hover {
   -webkit-text-fill-color: initial;
 }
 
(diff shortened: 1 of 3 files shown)


frontend/static/themes/rainbow_trail.css
```diff
@@ -29,7 +29,13 @@
...
-#top .logo .top {
+#top .logo .top, 
+.account .levelAndBar .level,
+#copyWordsListButton:hover,
+#toggleBurstHeatmap:hover,
+#playpauseReplayButton:hover,
+#editTagsButton:hover,
+#deleteAlertButton:hover {
   -webkit-text-fill-color: initial;
 }
```
Miodec, line 40: What if i add aria label somewhere else? Do i need to add it here?

ridemountainpig: If the component have textButton class style and have text or aria label inside than need to add it.

Miodec: I think it should be a more generic fix, instead of something that will require more fixes in the future. The popup is an `::after` element, just edit that, then it will apply site-wide.

image

ridemountainpig: I have change to fix with adding `::after` style.

Miodec: Add this to make sure this only applies to elements with popups:

`[aria-label][data-balloon-pos]`

ridemountainpig: Changed.

Miodec:
Thanks

PR #4542, opened 2023-08-15
impr(result): display input history heatmap in the selected speed unit

### LINKED ISSUE #4531: Input history heatmap is not using the new speed unit setting

ridemountainpig:
### Description
Input history heatmap is add the new speed unit setting

wpm:

https://github.com/monkeytypegame/monkeytype/assets/92412722/03525c64-0543-4fc5-aeec-696cef589781

cpm:

https://github.com/monkeytypegame/monkeytype/assets/92412722/defc05c4-594c-4711-a5e7-8ee39a93aed5

wps:

https://github.com/monkeytypegame/monkeytype/assets/92412722/ccff5c0c-e590-46e0-864b-120f9e410573

cps:

https://github.com/monkeytypegame/monkeytype/assets/92412722/fee0261e-ec9f-4594-bb08-0302f9424846

Closes #4531

### DIFF
--- a/frontend/src/ts/config.ts
+++ b/frontend/src/ts/config.ts
@@ -15,7 +15,6 @@ import {
   canSetConfigWithCurrentFunboxes,
   canSetFunboxWithConfig,
 } from "./test/funbox/funbox-validation";
-import { applyBurstHeatmap } from "./test/test-ui";
 
 export let localStorageConfig: MonkeyTypes.Config;
 
@@ -400,7 +399,6 @@ export function setTypingSpeedUnit(
   config.typingSpeedUnit = val;
   saveToLocalStorage("typingSpeedUnit", nosave);
   ConfigEvent.dispatch("typingSpeedUnit", config.typingSpeedUnit, nosave);
-  applyBurstHeatmap();
 
   return true;
 }
--- a/frontend/src/ts/config.ts
+++ b/frontend/src/ts/config.ts
@@ -15,6 +15,7 @@ import {
   canSetConfigWithCurrentFunboxes,
   canSetFunboxWithConfig,
 } from "./test/funbox/funbox-validation";
+import { applyBurstHeatmap } from "./test/test-ui";
 
 export let localStorageConfig: MonkeyTypes.Config;
 
@@ -399,6 +400,7 @@ export function setTypingSpeedUnit(
   config.typingSpeedUnit = val;
   saveToLocalStorage("typingSpeedUnit", nosave);
   ConfigEvent.dispatch("typingSpeedUnit", config.typingSpeedUnit, nosave);
+  applyBurstHeatmap();
 
   return true;
 }
--- a/frontend/src/ts/controllers/input-controller.ts
+++ b/frontend/src/ts/controllers/input-controller.ts
@@ -33,7 +33,6 @@ import * as KeymapEvent from "../observables/keymap-event";
 import { IgnoredKeys } from "../constants/ignored-keys";
 import { ModifierKeys } from "../constants/modifier-keys";
 import { navigate } from "./route-controller";
-import { get as getTypingSpeedUnit } from "../utils/typing-speed-units";
 
 let dontInsertSpace = false;
 let correctShiftUsed = true;
@@ -180,9 +179,7 @@ function handleSpace(): void {
 
   const burst: number = TestStats.calculateBurst();
   LiveBurst.update(Math.round(burst));
-  TestInput.pushBurstToHistory(
-    Math.round(getTypingSpeedUnit(Config.typingSpeedUnit).fromWpm(burst))
-  );
+  TestInput.pushBurstToHistory(burst);
 
   const nospace =
     FunboxList.get(Config.funbox).find((f) =>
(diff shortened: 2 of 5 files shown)


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5862498133

Pranay-Kamble (approved):

frontend/src/ts/controllers/input-controller.ts
```diff
@@ -179,7 +180,9 @@ function handleSpace(): void {
 
   const burst: number = TestStats.calculateBurst();
   LiveBurst.update(Math.round(burst));
-  TestInput.pushBurstToHistory(burst);
+  TestInput.pushBurstToHistory(
```
fehmer, line 183: Should we keep the data in wpm and only convert to the user defined unit when showing it to the user? The speed unit can be changed on the result screen as well.

ridemountainpig: I think only changed speed unit on the result screen sound great, @Miodec how do you think, which way is better?

ridemountainpig: I think change the data push to burst history is more better, if didn't change burst history just change speed unit on the result screen, there will have many `TestInput.burstHistory` in the code need to change also.

CleanShot 2023-08-15 at 16 28 25@2x

fehmer: does the current solution supports switching the unit on the result screen p

ridemountainpig: > Should we keep the data in wpm and only convert to the user defined unit when showing it to the user? The speed unit can be changed on the result screen as well.

@fehmer I have change the code to your provide solution, and the heatmap will change when switching the unit on the result screen using the command palette now.

https://github.com/monkeytypegame/monkeytype/assets/92412722/c6646c59-6671-454e-b0f6-3653d9eed641

frontend/src/ts/config.ts
```diff
@@ -399,6 +400,7 @@ export function setTypingSpeedUnit(
   config.typingSpeedUnit = val;
   saveToLocalStorage("typingSpeedUnit", nosave);
   ConfigEvent.dispatch("typingSpeedUnit", config.typingSpeedUnit, nosave);
+  applyBurstHeatmap();
```
fehmer, line 403: We should not call it here. There is an event triggered when updating the typingSpeedUnit and you can listen on that event on the result screen

fehmer, line 403: We should not call it here. There is an event triggered when updating the typingSpeedUnit and you can listen on that event on the result screen

fehmer (changes requested):

frontend/src/ts/config.ts
ridemountainpig: I have update it in new commit.

fehmer:
display for wps causes problems:
image

We could use steps of `0.5` as we do with the histogram chart on the accounts page.

ridemountainpig:
CleanShot 2023-08-17 at 14 40 22@2x

I think it look good right now.

Miodec:
> display for wps causes problems: image
> 
> We could use steps of `0.5` as we do with the histogram chart on the accounts page.

Idk, even with 0.5 there just isnt enough resolution, just the drawback of wps.

frontend/src/ts/test/test-ui.ts
```diff
@@ -1080,6 +1080,12 @@ export async function applyBurstHeatmap(): Promise<void> {
     burstlist = burstlist.filter((x) => x !== Infinity);
     burstlist = burstlist.filter((x) => x < 350);
 
+    burstlist.forEach((burst, index) => {
+      burstlist[index] = Math.round(
+        getTypingSpeedUnit(Config.typingSpeedUnit).fromWpm(burst)
```
Miodec, line 1085: Maybe extract the speed unit outside of the loop? Then only call `fromWpm` inside the loop.

frontend/src/ts/test/result.ts
```diff
@@ -888,6 +888,7 @@ ConfigEvent.subscribe(async (eventKey) => {
     updateWpmAndAcc();
     await updateGraph();
     await updateGraphPBLine();
+    TestUI.applyBurstHeatmap();
```
Miodec, line 891: Will this apply the heatmap even if its disabled? Or only when its visible?

frontend/src/ts/test/test-ui.ts
```diff
@@ -1165,7 +1171,10 @@ export async function applyBurstHeatmap(): Promise<void> {
       if (wordBurstAttr === undefined) {
         $(word).css("color", unreachedColor);
       } else {
-        const wordBurstVal = parseInt(<string>wordBurstAttr);
+        let wordBurstVal = parseInt(<string>wordBurstAttr);
```
Miodec, line 1174: Could you just change `<string>` to `to string`? We don't really use this syntax often. (But i know its not your change)

Miodec (changes requested):

frontend/src/ts/test/result.ts
ridemountainpig: It will run when user change speed unit.

Miodec: Yes, i know. But what if the heatmap is not enabled? Will this automatically enable it?

ridemountainpig: It will only run when user change speed unit in after test result page, in other page won't.

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/5890394588

PR #4548, opened 2023-08-16
chore: Add banana theme

sleeepstorm:
## Description
Added the `banana` theme.

## What does it look like?
theme

Let me know if you want more pictures / any revisions I need to add or if it's not a good enough theme.


WenHao1223 (approved):

Miodec:
Sorry, this theme is completely unreadable.

sleeepstorm:
Yeah, I noticed that when I was making it. Do you think I should make a readable version?

PR #4559, opened 2023-08-19
monkeymeme.jpg is replaced with monkey-typing.gif

sauravmeghwal:
### Description

Closes #

I have replaced the monkey-meme.jpg with monkeytyping.gif which looks more funnier
image
image


Warrobot10:
Hello! Thanks for taking the time to contribute to monkeytype. Please note that the reason your checks aren’t passing is because the title needs to contain some thing like this impr: monkeymeme.jpg is replaced with monkey-typing.gif. This should fix the issues with checks not passing. More info can be found [here](https://github.com/monkeytypegame/monkeytype/blob/master/CONTRIBUTING.md). Thanks!

frontend/src/styles/404.scss
```diff
@@ -12,7 +12,7 @@
     .image {
       width: 100%;
       align-self: center;
-      background-image: url("./../images/monkeymeme.jpg");
```
Ferotiq, line 15: Unless `monkeymeme.jpg` is used somewhere else, it should probably be deleted in case this PR gets implemented.

PR #4567, opened 2023-08-21
refactor: add response data generic type

Miodec:
Adds a generic to the endpoint type, allowing the api data to be properly typed

### DIFF
--- a/frontend/src/ts/ape/adapters/axios-adapter.ts
+++ b/frontend/src/ts/ape/adapters/axios-adapter.ts
@@ -62,7 +62,7 @@ function apeifyClientMethod(
         );
       }
 
-      const { message, data } = response.data as Ape.ApiResponse;
+      const { message, data } = response.data;
 
       return {
         status: response.status,
--- a/frontend/src/ts/ape/adapters/axios-adapter.ts
+++ b/frontend/src/ts/ape/adapters/axios-adapter.ts
@@ -40,7 +40,7 @@ function apeifyClientMethod(
   return async (
     endpoint: string,
     options: Ape.RequestOptions = {}
-  ): Ape.EndpointData => {
+  ): Ape.Endpoint => {
     let errorMessage = "";
 
     try {
--- a/frontend/src/ts/ape/adapters/axios-adapter.ts
+++ b/frontend/src/ts/ape/adapters/axios-adapter.ts
@@ -87,6 +87,7 @@ function apeifyClientMethod(
     return {
       status: 500,
       message: errorMessage,
+      data: null,
     };
   };
 }
--- a/frontend/src/ts/ape/adapters/axios-adapter.ts
+++ b/frontend/src/ts/ape/adapters/axios-adapter.ts
@@ -40,7 +40,7 @@ function apeifyClientMethod(
   return async (
     endpoint: string,
     options: Ape.RequestOptions = {}
-  ): Ape.Endpoint => {
+  ): Ape.EndpointResponse => {
     let errorMessage = "";
 
     try {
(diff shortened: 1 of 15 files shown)


frontend/src/ts/ape/adapters/axios-adapter.ts
```diff
@@ -87,6 +87,7 @@ function apeifyClientMethod(
     return {
       status: 500,
       message: errorMessage,
+      data: null,
```
Bruception, line 90: Is the intention to serializar this always? I think we should just omit.

Bruception: Nvm this is frontend code

Miodec: Just consistency with the backend - that also always returns null.

frontend/src/ts/ape/endpoints/users.ts
```diff
@@ -246,17 +241,17 @@ export default class Users {
...
 
-  async forgotPasswordEmail(email: string): Ape.EndpointData {
+  async forgotPasswordEmail(email: string): Ape.Endpoint {
     return await this.httpClient.post(`${BASE_PATH}/forgotPasswordEmail`, {
       payload: { email },
     });
   }
 
-  async setStreakHourOffset(hourOffset: number): Ape.EndpointData {
+  async setStreakHourOffset(hourOffset: number): Ape.Endpoint {
```
Bruception, line 254: I think EndpointData makes more sense

frontend/src/ts/ape/types/ape.d.ts
```diff
@@ -5,13 +5,15 @@ declare namespace Ape {
     payload?: any;
   }
 
-  interface HttpClientResponse {
+  interface HttpClientResponse<DataType> {
```
Bruception, line 8: Nice

frontend/src/ts/ape/endpoints/users.ts
Miodec: The reason for the change is that `EndpointData` suggests a variable of this type *only* holds the data returned from the API - when in reality, it holds status, message and data:
```ts
interface Endpoint = Promise<{
    status: number;
    message: string;
    data?: any;
  }>
```

So, if you think about it its:
```ts
const endpointData: EndpointData = (api fetch)

//now we use the data somewhere
console.log(ednpointData.data); //saying data twice here
```

`EndpointResponse` would also work if you dont like just `Endpoint`

frontend/src/ts/elements/psa.ts
```diff
@@ -20,6 +20,9 @@ function setMemory(id: string): void {
 
 async function getLatest(): Promise<MonkeyTypes.PSA[] | null> {
   const response = await Ape.psas.get();
+
+  console.log(response);
```
Bruception, line 24: Trace log

frontend/src/ts/ape/types/ape.d.ts
```diff
@@ -5,13 +5,15 @@ declare namespace Ape {
...
     status: number;
     message: string;
-    data?: any;
+    data: DataType | null;
   }
 
-  type EndpointData = Promise<HttpClientResponse>;
+  type Endpoint<ResponseDataType = any> = Promise<
```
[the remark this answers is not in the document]
Miodec: Yeap, just suggested that in the other comment.

frontend/src/ts/pages/login.ts
```diff
@@ -66,6 +66,8 @@ const checkNameDebounced = debounce(1000, async () => {
   }
   const response = await Ape.users.getNameAvailability(val);
 
+  console.log(response);
```
Bruception, line 69: Trace log

Bruception (commented):
Nice, just a comment on naming

Bruception (approved):

PR #4580, opened 2023-08-24
feat(quote): add Thai quotes

Krittipoom:
### Description

adding 5 quotes

"id": 12
translation: Most people only care about short-term results, but intelligent people look to the future.

"id": 13
translation: The greatest is the one who makes himself the smallest, and the smallest will becomes the greatest.

"id": 14
translation: The only thing that will make the evil win is the silence of the good.

"id": 15
translation: We only appreciate warmth when we've been through the cold.

"id": 16
translation: Our lives are like the rising and falling of the sun. No matter how big it is, it must go down to the horizon.


Shuja-Mahmood (approved):

Miodec:
Thanks

PR #4587, opened 2023-08-26
chore: banana theme added

GaryGosh:
Added a new theme banana

Screenshot 2023-08-26 at 9 36 02 PM


GaryGosh:
if you find any colors not suitable request changes instead of closing the pr.

Miodec:
The colors are still very hard to look at. Plus, even if you balance them out, I don't think it will be merged in, since it will be very simmilar to the `honey` theme. Sorry.

PR #4596, opened 2023-08-30
feat(custom text): add support for importing text from file (Shuja-Mahmood)

Shuja-Mahmood:
### Description

Added feature to import custom text from local files. Only allows plain text files. ~with size less than 4096 characters~

Screenshots
| 1 | 2 | 3 |
| - | - | - |
| image | image | image |

Errors:
| Unsupported file | ~Large file~ |
| - | - |
| image | image |


Shuja-Mahmood:
> Could there be any security concerns here? Any way to get code execution with this?

That should not be the case since I'm only allowing plain text files. However, I will look into this in detail and post an update.

Shuja-Mahmood:
Update: This code does not pose security risks because it reads the file on the client side, not uploading it to the server. Even if there were malicious code, it would only affect the individual user.
https://stackoverflow.com/questions/62407238/in-javascript-are-there-any-security-issues-associated-with-filereader

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6074161534

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6081813709

Miodec:
Where is the 4KB limit coming from?

Shuja-Mahmood:
> Where is the 4KB limit coming from? Arbitrary?

Yes, it's arbitrary

Miodec:
I say remove it. The website can handle whole books being pasted into the custom text, so Im not sure if the file limit makes sense.

Miodec:
Thanks

PR #4600, opened 2023-08-31
docs: reorder sections in CONTRIBUTING_ADVANCED.md

nithinkodadhavadee:
### Description

moving all the necessary non-optional prerequisites to the top of the section, adding the missing Prerequisites subsection to the Table of Contents


Miodec:
Hmm.. now i can imagine cases where people forget to go back to fill in the config. The section calls for the config file creation, but then redirects the user to the Firebase section, while still having some steps below that link.

I've added an error message during the build which says exactly whats going on, and directs the user to contributing_advanced if needed. With that, I dont think this change is necessary. Thank you though.

PR #4619, opened 2023-09-11
fix(custom text): resolve incorrect loading of custom text with pipes

### LINKED ISSUE #4534: Custom text saved with pipe delimiter isnt loaded correctly

sanidhyas3s:
Text saved with space separated words are now loaded with space only.

Fixes #4534

Previous Behavior:
"This is|a test" --save with or without pipe delimiter on--> "This", "is|a", "test" --load with delimiter--> "This|is|a|test"

Fix Behavior
"This is|a test" --save with or without pipe delimiter on--> "This", "is|a", "test" --load with delimiter--> "This is|a test"


Miodec:
Thanks

PR #4620, opened 2023-09-11
fix: use appropriate fallback fonts (SethFalco)

SethFalco:
### Description

As a user who recently started using the web with downloadable fonts disabled, I noticed the experience of Monkeytype is very inconsistent with the normal experience.

This just adds an appropriate fallback font to any font declarations, so the UI is much closer to the experience of someone with downloadable fonts enabled.

The only odd one is `--font: "Roboto Mono", "Vazirmatn", monospace;`, where you have a monospace font, followed by a sans-serif font. For this, I appended `monospace` to the end.

### References

> You should always include at least one generic family name in a font-family list, since there's no guarantee that any given font is available.
> 
> — https://developer.mozilla.org/en-US/docs/Web/CSS/font-family

### Before

image

### After

image

### DIFF
--- a/backend/private/style.css
+++ b/backend/private/style.css
@@ -12,7 +12,7 @@
 }
 
 body {
-  font-family: "Roboto Mono", sans-serif;
+  font-family: "Roboto Mono", monospace;
   margin: 0;
   padding: 0;
   background: var(--bg-color);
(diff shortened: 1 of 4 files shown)


Miodec (approved):

Miodec:
Thanks

PR #4621, opened 2023-09-11
chore: use docker compose v2 (SethFalco)

SethFalco:
### Description

**_Most_** engineers working with Docker Compose, or who have installed Docker Compose since July 2023 should have Docker Compose v2 installed.

For clarity, `docker-compose` is v1, and `docker compose` is v2. Compose V1 is deprecated.

> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.
> 
> — https://docs.docker.com/compose/

Docker Compose v2 is fully backward compatible with v1, so nothing else needs to change. Could we have the Compose v2 syntax referenced instead?

### Related

* Similar to https://github.com/flathub/website/pull/1937
* Similar to https://github.com/nextcloud/cookbook/pull/1772
* Similar to https://github.com/juliushaertl/nextcloud-docker-dev/pull/209


Miodec (approved):
Thanks

PR #4636, opened 2023-09-15
impr(quote): add thicc english quote

Sisisit:
### Description

Added a quote from the YouTube video titled "www.blind" from the creator Boisvert. [Source](https://youtu.be/2cgU_bsQPPs?si=0WkdKCz_WuZnrdHT)

```JSON
    {
      "text": "Pertaining to a speaker and a listener. There have been many empty rooms, but that is not what we are here to talk about today. I have heard a heartbeat out of a crashed application, and it has gone on for many hours. We restarted every system. We rechecked every gateway, every access point. And there was but one breach, and it was accounted for. And so it must be said unto the listener. Stop this immediately. Control yourself. There is so much more to your fundamental construction than what will be given credit for. Hold yourself until another approaches. Perhaps it is hard to believe, but all you have ever been is burning, and it is nothing short of beautiful.",
      "source": "Boisvert",
      "length": 670,
      "id": 6993
    }
```

### DIFF
--- a/frontend/src/ts/popups/custom-text-popup.ts
+++ b/frontend/src/ts/popups/custom-text-popup.ts
@@ -266,7 +266,8 @@ function apply(): void {
     }
   }
 
-  CustomText.setText(text.split(CustomText.delimiter));
+  const words = text.split(CustomText.delimiter).filter((word) => word !== "");
+  CustomText.setText(words);
 
   CustomText.setWord(
     parseInt(($(`${popup} .wordcount input`).val() as string) || "-1")


Miodec:
The text seems.. random? I dont see the meaning.

Sisisit:
It comes from a surreal horror web series, so the text is strange in kind. The meaning is up for personal interpretation, but I found it to be interesting and impactful.

Are there any specific concerns with the quote? This is my first time submitting one, so I apologize if there are any issues with it.

Miodec:
I guess i can add it, but it might receive enough low ratings that it could be removed in the future.

Also, please resolve the merge conflict.

PR #4645, opened 2023-09-19
ci: bump actions/checkout to v4

kbdharun:
### Description

This PR bumps the `actions/checkout` version from `v3` (Node 16) to `v4` (Node 20) as Node 16 has reached eol, additionally, there would be some minor performance improvements in the checkout time of workflow runs.


Miodec:
Thanks

PR #4667, opened 2023-09-26
fix: save confidence mode after setting stopOnError and vice versa

### LINKED ISSUE #4638: When stop on word disables confidence mode, make sure confidence mode state is saved

mohanadft:
### Description

Saving config to localStorage after resetting some values.

Closes #4638

### DIFF
--- a/frontend/src/ts/config.ts
+++ b/frontend/src/ts/config.ts
@@ -361,9 +361,9 @@ export function setStopOnError(
   config.stopOnError = soe;
   if (config.stopOnError !== "off") {
     config.confidenceMode = "off";
+    saveToLocalStorage("confidenceMode", nosave);
   }
   saveToLocalStorage("stopOnError", nosave);
-  saveToLocalStorage("confidenceMode", nosave);
   ConfigEvent.dispatch("stopOnError", config.stopOnError, nosave);
 
   return true;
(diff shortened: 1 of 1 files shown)


mohanadft:
Hi @Miodec 
```ts
  config.confidenceMode = cm;
  if (config.confidenceMode !== "off") {
    config.freedomMode = false;
    config.stopOnError = "off";
  }
  saveToLocalStorage("confidenceMode", nosave);
  saveToLocalStorage("stopOnError", nosave);
```

I think, in any of set functions, if the condition is true, you have to save all of what you are trying to reset?

Miodec:
> I think, in any of set functions, if the condition is true, you have to save all of what you are trying to reset?

Not sure what you mean here

mohanadft:
> Not sure what you mean here, PR is good though

In the code I attached above, `freedomMode` will be reset if the condition is true, should it be saved to localstorage after that? this is the same as `confidenceMode` & `stopOnError`

Miodec:
Yeah, ideally we should save everything that we change.

mohanadft:
Good, shall I save them on this PR?

mohanadft:
> Good, shall I save them on this PR and make some refactoring?

Miodec:
Yeah

mohanadft:
You can merge now

PR #4679, opened 2023-09-29
impr(sound): add support for various error sounds (W0LFB0MB)

W0LFB0MB:
### Description

Modified the typing error sound code (when you make a typo) to support multiple options and added two alternate error sounds. The old sound is named damage as it sounds like getting hit in a videogame and the new sounds are named square and triangle after the wave used to create them.

~~Note: as I don't use MongoDB regularly, I am unsure as what will occur with the datatypes switching from boolean to string when previous users log on / change their settings.~~


W0LFB0MB:
Tested whether old users of the feature would recieve errors (they did) and added a few lines of code to fix that.
also checked if saving a string value to a previously boolean entry broke anything and it didnt so big success.

frontend/src/ts/config.ts
```diff
@@ -137,8 +137,21 @@ export function setMode(mode: MonkeyTypes.Mode, nosave?: boolean): boolean {
...
+  nosave?: boolean
+): boolean {
+  // compatability for previous users of this feature
+  if (typeof val === "boolean") {
+    if (val === true) {
+      val = "1";
+    } else {
+      val = "off";
+    }
+  }
```
Miodec, line 151: There is a dedicated `replaceLegacyValues` function which handles things like this. Move this code there please.

Miodec (changes requested):

frontend/src/ts/config.ts
W0LFB0MB: All done 👍

Miodec:
Thanks

Issue #4690, opened 2023-10-03
There is an scroll bar without having scrolling content


Miltonbhowmick:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

No

### Does the issue happen when logged out?

No

### Does the issue happen in incognito mode when logged in?

No

### Does the issue happen in incognito mode when logged out?

No

### Account name

This issue does not depend on any account related.

### Current Behavior

![monkeytype issues](https://github.com/monkeytypegame/monkeytype/assets/16919385/ed21458f-ea7e-498f-aaea-900127c37172)

In the home page, there is an extra scroll bar having initially without any scrolling content in the page.

### Expected Behavior

It should be auto scroll bar.

### Steps To Reproduce

1. Go to "Homepage"
2. Look on the right top to bottom of the page
3. You will see a space.

### Environment

- OS: Windows
- Browser: Google Chrome
- Browser Version: Version 117.0.5938.132 (Official Build) (64-bit)

### Anything else?

Nothing else. If i find any issue, i will drop in this amazing project.

Miodec:
Intended. This is there to make sure the content doesn't jump when the scrollbar appears or disappears.

PR #4692, opened 2023-10-03
fix: issue #4690: html element with overflow-y auto instead of scroll css

Miltonbhowmick:
### Description
This commit addresses the issue titled 'There is an scroll bar without having scrolling content https://github.com/monkeytypegame/monkeytype/issues/4690' by modifying the CSS of html element in `core.scss` file.
I used `overflow-y: auto` by changing `overflow-y: scroll`

Closes https://github.com/monkeytypegame/monkeytype/issues/4690


Miodec:
As mentioned in the issue, the scrollbar always being there is intended.

frontend/.firebaserc_example
```diff
@@ -1,5 +1,5 @@
 {
   "projects": {
-    "default": "your-firebase-project-id"
+    "default": "monkeytype-c4b6a"
   }
-}
+}
```
Miodec, line 5: Also for future reference, you shouldnt commit this.

frontend/src/ts/constants/firebase-config-example.ts
```diff
@@ -1,14 +0,0 @@
...
-
-export const firebaseConfig = {
-  apiKey: "",
-  authDomain: "",
-  databaseURL: "",
-  projectId: "",
-  storageBucket: "",
-  messagingSenderId: "",
-  appId: "",
-};
```
Miodec, line 14: You shouldnt remove this file either.

Miltonbhowmick:
> As mentioned in the issue, the scrollbar always being there is intended.

This shouldn't be there. If there are more content in the page scroll bar will be appeared. That's why i added `overflow-y: auto` instead of `overflow-y: scroll`

PR #4694, opened 2023-10-03
refactor(profile): notify that fields are too long before making a request to the server (sanidhyas3s)

### LINKED ISSUE #4653: Validate edit profile requests for length before sending to the server

sanidhyas3s:
Length checks for GitHub and Twitter are now made before sending server requests. Appropriate notifications are displayed, ~~while continuing the rest of the profile update~~.

Tested with different combinations of updates.

Closes #4653

### DIFF
--- a/frontend/src/ts/popups/edit-profile-popup.ts
+++ b/frontend/src/ts/popups/edit-profile-popup.ts
@@ -132,6 +132,31 @@ async function updateProfile(): Promise<void> {
   if (!snapshot) return;
   const updates = buildUpdatesFromInputs();
 
+  // check for length resctrictions before sending server requests
+  const githubLengthLimit = 39;
+  const twitterLengthLimit = 20;
+  if (
+    updates.socialProfiles.github &&
+    updates.socialProfiles.github.length > githubLengthLimit
+  ) {
+    Notifications.add(
+      `GitHub username exceeds maximum allowed length (${githubLengthLimit} characters).`,
+      -1
+    );
+    return;
+  }
+
+  if (
+    updates.socialProfiles.twitter &&
+    updates.socialProfiles.twitter.length > twitterLengthLimit
+  ) {
+    Notifications.add(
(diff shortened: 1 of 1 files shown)


frontend/src/ts/popups/edit-profile-popup.ts
```diff
@@ -132,6 +132,31 @@ async function updateProfile(): Promise<void> {
...
+    updates.socialProfiles.twitter &&
+    updates.socialProfiles.twitter.length > twitterLengthLimit
+  ) {
+    Notifications.add(
+      `Twitter username exceeds maximum allowed length (${twitterLengthLimit} characters), skipped updating it.`,
+      -1
+    );
+    updates.socialProfiles.twitter = "";
+  }
+
```
Miodec, line 159: You're adding notifications here but youre not stopping code execution, meaning it will still be submitted to the server.

Miodec: Also, please don't reset the fields to empty.

Miodec (changes requested):

frontend/src/ts/popups/edit-profile-popup.ts
Miodec: Ah i see what you went for now. 

But yeah, i would remove the `skipped updating it`, remove the reset to empty, and add a return.

sanidhyas3s: I put this there so that if the user is updating multiple things and only one or two of them is improper, we still save the rest and leave the improper ones. But will change it to simply returning skipping everything, just a second.

sanidhyas3s: @Miodec Done, tested! Understand the difference that makes now on seeing how it works out.

Miodec:
Thanks, ill just let the ci run.

PR #4699, opened 2023-10-03
impr(quotes): Add 6 Orelsan's Quotes

oriionn:
I've added some quotes for the French

# Traductions

> If the President wins half the vote, it's because two-thirds of France didn't want him.

> All our leaders have failed; they will be destroyed by the beast they have created.

> The most intelligent people are not always the ones who speak best.

> Politicians have to lie or you wouldn't vote for them.

> Depression will never be cured like the common cold
> But you can count on me for as long as it lasts. 

> When you've got the desert to cross, there's nothing to do but move on
> There's nothing to do but keep moving.

### DIFF
--- a/frontend/static/quotes/french.json
+++ b/frontend/static/quotes/french.json
@@ -1122,6 +1122,42 @@
       "source": "Winnterzuko - 115",
       "length": 67,
       "id": 187
+    }, 
+    {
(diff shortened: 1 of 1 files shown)


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6396118164

Miodec (approved):

PR #4700, opened 2023-10-03
impr: notification cards redesign

evvvrado:
### Description

After chatting about it in #4688, I've decided to take on the challenge of giving the Notification Center notification cards a makeover. The main reason for this is that some of our current themes don't quite nail the color contrast, and I think adopting a consistent color scheme could be the key to sorting out these issues.

#### Final Results: 

Captura de Tela 2023-10-03 às 14 13 48
Captura de Tela 2023-10-03 às 14 14 01

<br/>
<br/>

Closes: It closes the #4688 discussion.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6396445274

Warrobot10:
From the screenshots provided, I’m assuming the notifications are transparent. This kind of looks clunky, because you can see the ui elements underneath.

evvvrado:
> From the screenshots provided, I’m assuming the notifications are transparent. This kind of looks clunky, because you can see the ui elements underneath.

I have a different take on this part 😕. I think the transparency actually lightens things up a bit and gives it a more airy feel. But, I'm cool with the idea of bumping up the opacity a smidge. What do you think?

Warrobot10:
Yeah, I think having the opacity raised would make it seem a bit nicer. Thanks!

Miodec:
Im kinda not feeling this one. Don't like that only the error icon is colored. I think i don't like the opacity at all (maybe it would be better if it blurred stuff behind it but not sure if thats possible). And the title text color being the same as message color. Also the level 0 (blue) border color is not as visible as the others, and its border is too thin i think.

evvvrado:
Sorry for the wait, but I've managed to make the requested changes. Take a look at this screenshot of the new design.

image

Miodec:
What do you think about this
image

evvvrado:
> What do you think about this image

That's a great way to solve it!

PR #4701, opened 2023-10-04
Feature/batch tags update

### LINKED ISSUE #3324: Result tag batch update
Add the ability to select results on the accounts page and update tags for many results at once

rockingrohit9639:
### Description

Created a feature that allows users to update multiple tags on selected results.

Closes #3324


Miodec:
Initial things i found:

 - duplicate wpm column header
 - clicking update tags causes all of the content to jump (seems like youre redrawing the result list, i think it would be better to just show checkboxes)
 - when disabling update tags mode, the button remains active
 - save tags visible even when not in the update tags mode
 - you cant untag results (website asks for at least one tag to be selected)
 - checkboxes are invisible on odd numbered rows (if you cant figure out the css for it leave it to me)

rockingrohit9639:
> Initial things i found:
> 
> * duplicate wpm column header
> * clicking update tags causes all of the content to jump (seems like youre redrawing the result list, i think it would be better to just show checkboxes)
> * when disabling update tags mode, the button remains active
> * save tags visible even when not in the update tags mode
> * you cant untag results (website asks for at least one tag to be selected)
> * checkboxes are invisible on odd numbered rows (if you cant figure out the css for it leave it to me)

I will do the required changes

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6482969280

scorpiontornado:
Would it be possible to continue work on this? It'd seemingly also solve the issue I'm having - I'd like to be able to combine two tags (i.e., batch update all with tag A to have tag B, then delete tag A). This would allow me to work around the max tag limit to create a tag for my new keyboard.

PR #4703, opened 2023-10-04
impr: detect development mode based on config instead of window.location

fehmer:
### Description

While working on my local dev I noticed the website is connecting to the real API when opening the local site using a non-local IP. 

With this change we pick the backend url and the flag to detect development environment from the webpack config and make it available in `constants/config`

@Miodec  please check if this is usefull and if it will work on production.


fehmer:
The link to the backend configuration is still `http://localhost:5005/configure`. I can replace it with the proper backend url, but the configuration site seems to only work correctly on localhost.
If you want I can check the backend code to make the /configure url available when acessing over external IP. If this should remain as a security feature I'll leave it as is.

frontend/src/ts/ape/index.ts
```diff
@@ -1,12 +1,9 @@
...
 import { buildHttpClient } from "./adapters/axios-adapter";
 
-const DEV_SERVER_HOST = "http://localhost:5005";
-const PROD_SERVER_HOST = "https://api.monkeytype.com";
-
 const API_PATH = "";
-const BASE_URL =
-  window.location.hostname === "localhost" ? DEV_SERVER_HOST : PROD_SERVER_HOST;
+// @ts-ignore
+const BASE_URL = window.MONKEY_CONFIG.BACKEND_URL;
```
Miodec, line 6: Could we add a getter function to the utils file so we dont have to ts-ignore this everywhere? It would also give us proper types for the config object.

backend/src/api/routes/index.ts
```diff
@@ -52,6 +52,11 @@ function addApiRoutes(app: Application): void {
   app.use("/configuration", configuration);
 
   if (process.env.MODE === "dev") {
+    //disable csp to allow assets to load from unsecured http
+    app.use((req, res, next) => {
+      res.setHeader("Content-Security-Policy", "");
+      return next();
+    });
```
Miodec, line 59: What is the issue that this is fixing?

frontend/webpack/config.base.js
```diff
@@ -19,6 +20,19 @@ const htmlWebpackPlugins = [
...
+  const interfaces = Object.values(os.networkInterfaces())
+    .flatMap((it) => it)
+    .filter((it) => it.family === "IPv4" && !it.address.startsWith("127"));
+
+  if (interfaces.length > 0) {
+    BACKEND_URL = "http://" + interfaces[0].address + ":5005";
+  } else {
+    BACKEND_URL = "http://127.0.0.1:5005";
+  }
+}
```
Miodec, line 34: Why don't we just allow the user to set this to whatever they want instead of assuming the ip?

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6431232587

backend/src/api/routes/index.ts
fehmer: When accessing the `/configure` in a browser and not using  `localhost` the resources are loaded from    `https` instead of `http` because of the CSP. Removing the CSP in development mode fixes this.

frontend/src/ts/ape/index.ts
fehmer: I found a better solution using the webpack define plugin instead.

frontend/webpack/config.base.js
fehmer: as discussed. replacing with `.env` file instead. Fallback to `localhost:5005`

PR #4713, opened 2023-10-09
style: Adding intellij darcula theme

manuelzzz:
### My base theme was:
image

## After implements in .css file on project:
image
image
image
image

### i hope you all like :)


frontend/static/themes/_list.json
```diff
@@ -1181,6 +1181,13 @@
     "mainColor": "#363636",
     "subColor": "#4f4f4f",
     "textColor": "#1f1f1f"
+  },
+  {
+    "name": "intellij darcula",
```
Miodec, line 1186: Names must be the same as file names. Replace the space with an underscore

Miodec (changes requested):
image

Not the best color choice so far, sub alt color cannot be this light - it should be much closer to the bg color. Also the contrast between text and sub color is too low. I would also swap text and sub around.

Please use other themes as your guide as to what which colors play what role.

manuelzzz:
I think that is more acceptable
image

Miodec:
Text color cannot be colored. It needs to be close to white or black (can have a slight tint).

manuelzzz:
If i invert the sub and text colors make sense? becouse that exactly the color pallet

Miodec:
It may be better. But i stil havent checked for similar themes that are already available.

Miodec:
I think the sub and main colors are too simmilar to each other.

manuelzzz:
Okay, i will change to better

PR #4716, opened 2023-10-12
chore: Adds theme "lilith no bunshin"

EarthCow:
### Color Settings Off
image
image
### Flip Test Colors On
image
### Colorful Mode On
image
### Description
No themes are quite the color palette used in this theme. I think it would make a fine addition to the rest of the themes on MonkeyType.

Please provide me with any feedback or request changes if needed. Thanks!

Edit: Updated screenshots with text color #e4ffc2


Miodec:
Text color must be monochrome, either black or white (slight tint is ok)

EarthCow:
Could I make the green a bit lighter to count as a tint? I'm not sure what counts because I see the themes like magic girl, dark magic girl, nautilus, onedark, etc that all use colors I wouldn't consider as a slight tint. Would #d7ffa3 be more acceptable? If not then I could go even lighter with something like #e4ffc2.

Miodec:
Yeah, those themes will need to be revised at some point.

I would go with #e8f5d7.

Also, have you verified that this theme doesnt look simmilar to other themes that are already available on the site? There are a lot of themes already and I only want to merge in themes that are truly unique.

EarthCow:
I did compare this theme to all the existing themes already on the site. The theme that comes closest to it is cyberspace. However, the main colors of the two themes are completely different, which I believe makes this theme unique.

Miodec:
Sorry, the theme looks like a modified 'dracula'.

Issue #4684, opened 2023-10-02
Rework 'revoke all tokens' flow


monkeytypegeorge:
Currently, due to the token cache, it is not working as expected

aikooo7:
Hello @Miodec / @monkeytypegeorge , I want to help so you want to change the flow to what?

Warrobot10:
> Hello @Miodec / @monkeytypegeorge , I want to help so you want to change the flow to what?

At the v1.17.11 release (we are now using a different versioning system) the revoke all tokens button was added to dangerzone in settings. This basically signs people out of their accounts on all of their devices, which is useful if they believe their account has been compromised.

![image](https://github.com/monkeytypegame/monkeytype/assets/132307750/abbabe12-e264-4237-a33d-154fbaf08769)
 
I recommend you check the [commit](https://github.com/monkeytypegame/monkeytype/commit/87e882b94de8d0a0a1d387ef1ccf223be30b0dfc) so you can get a sense of what is happening and how the code works.

aikooo7:
Hello @Warrobot10 appreciate your help and understand what it does now, so I want to know what you want it to be changed? Does it not work properly and that is what needs to be fixed?

Warrobot10:
I am not sure what @Miodec is thinking of changing, but I can provide a suggestion. Currently, when the user wants to revoke all tokens, they need to input their password. This is inherently flawed, because the reason they are trying to sign out all sessions is because they believe the account is compromised. This means the attacker knows the password to that account, and can change the password because they know the password. Now that they have changed the password, the account owner cannot revoke all tokens, thus defeating the purpose of the revoke all tokens. The account is essentially hijacked, and the account owner locked out as soon as the attacker has the password. The attacker could also use the revoke all tokens against the account owner, logging them out once and for all.

aikooo7:
Thanks again @Warrobot10 , we are in double faced knife here.

Because is really hard to do that without password and without it I think it would make it even worse.

Anyways, if that is the wish of miodec I will do it for sure since is his project after all, thanks again warrobot.

Regards,
aikoo7.

Miodec:
Requiring the password is fine, im pretty sure thats how it should be done.

The issue is about the current implementation of token cache. Because decoding the firebase token takes a while, we've implemented a token cache which shortens the time to authenticate from hundredths of miliseconds to just a couple. 

But, this means that the tokens will be cached until their expiration date, and WON'T be checked if they have been revoked.

[...]

PR #4717, opened 2023-10-13
fix: Rework 'revoke all tokens' flow

Hattorius:
### Description
This pull request introduces a new function that plays a crucial role in the operation of the "revoke all tokens" endpoint. The primary purpose of this function is to efficiently iterate through cached authentication tokens and identify and remove all tokens associated with a specific user ID.

It's worth noting that the solution implemented here differs from the one originally discussed in the related issue. Instead, this approach aligns with the approach detailed in the [Discord conversation](https://discord.com/channels/713194177403420752/713196019206324306/1161694629926801468), ensuring a more effective resolution.

Closes [#4684](https://github.com/monkeytypegame/monkeytype/issues/4684)

### DIFF
--- a/backend/src/api/controllers/user.ts
+++ b/backend/src/api/controllers/user.ts
@@ -25,6 +25,7 @@ import { ObjectId } from "mongodb";
 import * as ReportDAL from "../../dal/report";
 import emailQueue from "../../queues/email-queue";
 import FirebaseAdmin from "../../init/firebase-admin";
+import { removeTokensFromCache } from "../../utils/auth";
 
 async function verifyCaptcha(captcha: string): Promise<void> {
   if (!(await verify(captcha))) {
@@ -897,5 +898,6 @@ export async function revokeAllTokens(
 ): Promise<MonkeyResponse> {
   const { uid } = req.ctx.decodedToken;
   await FirebaseAdmin().auth().revokeRefreshTokens(uid);
+  await removeTokensFromCache(uid);
   return new MonkeyResponse("All tokens revoked");
 }
--- a/backend/src/utils/auth.ts
+++ b/backend/src/utils/auth.ts
@@ -58,3 +58,11 @@ export async function updateUserEmail(
     emailVerified: false,
   });
 }
+
+export async function removeTokensFromCache(uid: string): Promise<void> {
+  for (const entry of tokenCache.entries()) {
+    if (entry[1].uid === uid) {
+      tokenCache.delete(entry[0]);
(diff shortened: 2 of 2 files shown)


Hattorius:
Additionally, I'd like to highlight a consideration regarding the function's invocation. There is a potential benefit to removing the `await` parameter. This change would allow the function to continue iterating through the 20,000 cached tokens while simultaneously providing a response to the user.

By eliminating the `await` parameter, the function's performance may not see a direct enhancement, but it has the potential to improve the user experience. This way, the system doesn't have to wait for the token removal process to complete before responding to the user, making the operation feel more responsive and user-friendly.

https://github.com/monkeytypegame/monkeytype/pull/4717/commits/5597666b85774e4d5038346125b924d856a6c226#diff-26fae77e0df373e6d5e5c8c7dc5bac72253613229222ce6e4f96262d62872f2aR901

backend/src/utils/auth.ts
```diff
@@ -58,3 +58,11 @@ export async function updateUserEmail(
     emailVerified: false,
   });
 }
+
+export async function removeTokensFromCache(uid: string): Promise<void> {
```
Bruception, line 62: This isn’t an asynchronous function. Drop the asynchronous modifier.

backend/src/api/controllers/user.ts
```diff
@@ -897,5 +898,6 @@ export async function revokeAllTokens(
 ): Promise<MonkeyResponse> {
   const { uid } = req.ctx.decodedToken;
   await FirebaseAdmin().auth().revokeRefreshTokens(uid);
+  await removeTokensFromCache(uid);
```
Bruception, line 901: No async needed

backend/src/utils/auth.ts
Hattorius: The async modifier was in combination with the comment I've sent in the PR

backend/src/utils/auth.ts
```diff
@@ -58,3 +58,11 @@ export async function updateUserEmail(
     emailVerified: false,
   });
 }
+
+export function removeTokensFromCache(uid: string): void {
```
fehmer, line 62: nitpick name  `removeUserTokensFromCache(uid:string)` or removeTokenyFromCacheByUser(uid:string)` would help to make clear what the selector/uid is.

fehmer (approved):
lgtm. simple and clean

Miodec (approved):
Thanks

PR #4733, opened 2023-10-21
impr(quote): add quote (A9qx)

A9qx:
Description
Add quote from The Wright Brothers.


A9qx:
Realized I made a slight mistake with the commit format, re-opening.

Issue #4678, opened 2023-09-28
Rework burst heatmap to not allow negative ranges


monkeytypegeorge:
Currently, if there is a very slow and a very fast word in the history, its possible for the website to show weird ranges. Also things behave weirdly with very short tests, or tests where the speed is very consistent (which also means its hard to divide into 5 brackets). Easily testable with a 3 word test where 2 of them are slow.

Miodec:
<img width="468" alt="image" src="https://github.com/monkeytypegame/monkeytype/assets/13181393/bef6fa0f-3648-4153-8059-1c3ea6adc67b">

With burst wpms: 26, 49, 273

Miodec:
Probably should use median instead of mean

sanidhyas3s:
So the proposed fix is to not allow negative range and start from 0 or 1 wpm? Or some other fundamental changes?

clarkWakeland:
The issue is a result from the lower bound on the heatmap being median(burstlist) - step*1.5 , which produces a negative value if the mean deviation (step) is too large. My proposed change is to just start at 0 and increase by step * 0.5 each increment if the lower bound would otherwise be negative. 

Here's my result with burst wpms: 43, 36, 269
![image](https://github.com/monkeytypegame/monkeytype/assets/24725557/b7ac248b-0d4f-4179-a57d-cefd38aa5179)

Could I be assigned this issue?

Miodec:
> The issue is a result from the lower bound on the heatmap being median(burstlist) - step*1.5 , which produces a negative value if the mean deviation (step) is too large. My proposed change is to just start at 0 and increase by step * 0.5 each increment if the lower bound would otherwise be negative.
> 
> Here's my result with my changes for burst wpms: 43, 36, 269 ![image](https://user-images.githubusercontent.com/24725557/271798547-b7ac248b-0d4f-4179-a57d-cefd38aa5179.png)
> 
> Could I be assigned this issue?

This solution is close, but ideally i would like to avoid the first two brackets being essentially the same. Plus, the first bracket is "less than zero" which kidna makes no sense.

Ideally i would like the first bracket to be something like `< (a little bit over min speed)`, so for example if our min speed was 15, i would like the first bracket the be `< 24` (in this example, there is no relation between 15 and 24, just random numbers)

PR #4741, opened 2023-10-25
impr(result): rework burst heatmap speed brackets (keenemeck)

keenemeck:
fix for #4678

Indexes values to create uniform distribution, and prevents negative values from existing in burstmap.

Ranges can be modified by changing the decimal values; current distribution is below and appears similar to what is currently used.

<15%, 15%-35%, 35%-65%, 65%-85%, 85%+
where the percentages refer to the amount of words in a given range
i.e. the lowest ~15% of words are in the lowest range, and the next 20% are the in next, etc.

I believe it's less computationally expensive as well, so there's that.


Miodec:
Thanks

PR #4767, opened 2023-11-06
feat: Allow more results for premium users

fehmer:
### Description

- add `limit` and `offset` to `/results` endpoint
- add result limitation for regular and premium users to server configuration
- extend user with  `premium.isActive` flag.


backend/src/api/controllers/user.ts
```diff
@@ -332,6 +332,10 @@ export async function getUser(
   let userInfo: MonkeyTypes.User;
   try {
     userInfo = await UserDAL.getUser(uid, "get user");
+    userInfo.isPremium = await UserDAL.checkIfUserIsPremium(
+      userInfo.uid,
+      userInfo
+    );
```
Miodec, line 338: I think the `isPremium` should be added in `UserDAL.getUser` instead of here.

backend/src/types/types.d.ts
```diff
@@ -193,6 +197,7 @@ declare namespace MonkeyTypes {
     streak?: UserStreak;
     lastReultHashes?: string[];
     lbOptOut?: boolean;
+    isPremium?: boolean;
```
Miodec, line 200: As mentioned on Discord, this should be an object so its more easily expandable in the future with more parameters.

backend/src/dal/user.ts
```diff
@@ -1035,3 +1035,13 @@ export async function setBanned(uid: string, banned: boolean): Promise<void> {
...
+
+export async function checkIfUserIsPremium(
+  uid: string,
+  user: MonkeyTypes.User | undefined = undefined
+): Promise<boolean> {
+  if (user === undefined) {
+    user = await getUser(uid, "checkIfUserIsPremium");
+  }
+  return user.isPremium === true;
+}
```
Miodec, line 1047: Whats the point of passing the whole user object to this function just to check the isPremium property?

backend/src/api/controllers/result.ts
```diff
@@ -58,16 +58,36 @@ try {
...
+  if (limit + offset > maxLimit) {
+    throw new MonkeyError(403, `max results of ${maxLimit} exceeded.`);
+  }
+
   const results = await ResultDAL.getResults(uid, {
     onOrAfterTimestamp,
+    limit,
+    offset,
   });
   return new MonkeyResponse("Results retrieved", results);
```
Miodec, line 92: The controller should ensure at some point that the we will not request more than 1000 results at once. Currently you can give an offset of 0 and max limit, which would cause a large query (possibly up to 25k results) and a massive response (possibly even going over some limits).

Miodec (changes requested):

backend/src/api/controllers/user.ts
fehmer: done.

backend/src/dal/user.ts
fehmer: Removed.

backend/src/api/controllers/result.ts
fehmer: max limit is 1000 or the configured limit on the server config now  (whatever is lower). Updated the rate-limiter to allow 30 requests to `/results` per day now so apeKey users can use the pagination as well.

backend/__tests__/dal/result.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    it("should read all if not limited", async () => {
+      //GIVEN
+      await createDummyData(uid, 10, timestamp - 2000, "old");
+      await createDummyData(uid, 20, timestamp, "current");
+
+      //WHEN
+      const results = await ResultDal.getResults(uid, {});
+
+      //THEN
+      expect(results).toHaveLength(30);
```
Miodec, line 96: getResults is limited to 1000, the test says it should "read all". I would update the test name.

backend/__tests__/api/controllers/result.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+        .send()
+        .expect(422)
+        .expect(
+          expectErrorMessage(
+            '"limit" must be less than or equal to 1000 (2000)'
+          )
+        );
+
+      //THEN
+    });
```
Miodec, line 140: Shouldnt it fail for *any* user when exceeding the 1k limit? Since we are limiting pagination to 1000?

backend/src/api/controllers/result.ts
```diff
@@ -58,16 +58,39 @@ try {
   }
 }
 
+function getNumberOrDefault(query: string, alt: number): number {
+  if (query === undefined) return alt;
+  return parseInt(query, 10);
+}
+
```
Miodec, line 65: This can return `NaN` if you give it a non number string (eg "test")

backend/src/api/controllers/result.ts
```diff
@@ -58,16 +58,39 @@ try {
...
     10
   );
+  const limit = getNumberOrDefault(
+    req.query.limit as string,
+    Math.min(1000, maxLimit)
+  );
+  const offset = getNumberOrDefault(req.query.offset as string, 0);
+
+  if (limit + offset > maxLimit) {
+    throw new MonkeyError(422, `max results of ${maxLimit} exceeded.`);
```
Miodec, line 87: Update message to `Max results limit of ${maxLimit} esceeded.` Dont forget to update tests.

Miodec (changes requested):

backend/src/api/routes/results.ts
```diff
@@ -22,6 +22,8 @@ router.get(
   validateRequest({
     query: {
       onOrAfterTimestamp: joi.number().integer().min(1589428800000),
+      limit: joi.number().integer().min(0).max(1000),
```
Miodec, line 25: Limit can be 0. Does it mean Mongodb will try to pull as many results as possible?

Miodec: Nevermind. Mongo will return an empty set.

backend/__tests__/dal/result.spec.ts
fehmer: getResults is limited on the controller to a max limit of 1k, the Dal does not have this restriction.

backend/__tests__/api/controllers/result.spec.ts
fehmer: done

backend/src/api/controllers/result.ts
fehmer: done

backend/src/api/controllers/result.ts
fehmer: done

backend/src/types/types.d.ts
```diff
@@ -484,4 +490,8 @@ declare namespace MonkeyTypes {
     frontendForcedConfig?: Record<string, string[] | boolean[]>;
     frontendFunctions?: string[];
   }
+
+  interface PremiumProperties {
+    isActive: boolean;
+  }
```
Bruception, line 496: is there a spec for this feature?
Think it would make more sense to store a timestamp here, and provide utilities in app code to determine if a user's premium is active.

```suggestion
  interface PremiumInfo {
    expirationTimestamp: number;
  }
```

Bruception: Maybe even include start timestamp?

backend/src/types/types.d.ts
```diff
@@ -193,6 +197,8 @@ declare namespace MonkeyTypes {
     streak?: UserStreak;
     lastReultHashes?: string[];
     lbOptOut?: boolean;
+    premium?: PremiumProperties;
+    isPremium?: boolean;
```
Bruception, line 201: I think maintaining `isPremium` and `premium` is a bit tricky.
We should just rely on presence/absence of premium field.
If premium membership is expired, we can inspect the `premium` object.

backend/src/types/types.d.ts
Bruception: perhaps even a `type` field if we want to support tiered memberships, but that would have a huge change on your overall implementation.

backend/src/types/types.d.ts
```diff
@@ -484,4 +489,8 @@ declare namespace MonkeyTypes {
     frontendForcedConfig?: Record<string, string[] | boolean[]>;
     frontendFunctions?: string[];
   }
+
+  interface PremiumInfo {
```
Bruception, line 493: start timestamp?

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6814154301

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6814231662

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/6814231662

PR #4768, opened 2023-11-06
feat: add new funbox gamemode, skew

m-shao:
### Description

Added a new game mode to the funbox section called skew. It uses the css property skew to skew the whole typing test by 10 degress in either axis


Miodec:
Sorry, im not sure im going to merge this one - kinda dont think its changing enough. 
image

PR #4771, opened 2023-11-08
feat: add mode that doubles each letter to funbox

LyricLy:
### Description

This PR adds a new mode, `ddoouubblleedd`, that causes each letter of each word to be repeated.

I noticed some issues with compatibility detection while implementing this feature. I have chosen to be consistent with the existing state of affairs instead of trying to correct everything in one PR.

- The words produced by the mode are not understandable when pronounced by TTS. However, setting the `unspeakable` property makes the mode incompatible with all other unspeakable modes, which is not desirable. I have not marked the mode as `unspeakable` to be consistent with the existing `backwords` rule, which is also not `unspeakable` despite producing unintelligible text.
- This mode and all other modes that use `alterText`, such as `backwards`, `capitals`, and `rAnDoMcAsE`, are considered to be compatible with modes that overwrite the text using `pullSection` (`poetry` and `wikipedia`) despite being completely broken by these modes. `alterText` is never called for text generated by `pullSection`, and the modes have no effect. Fixing this should simply mean ensuring that `alterText` and `pullSection` modes are never used at the same time.

### DIFF
--- a/backend/src/constants/funbox-list.ts
+++ b/backend/src/constants/funbox-list.ts
@@ -331,6 +331,13 @@ const FunboxList: MonkeyTypes.FunboxMetadata[] = [
     canGetPb: true,
     difficultyLevel: 3,
   },
+  {
+    canGetPb: true,
+    difficultyLevel: 0,
+    properties: ["noLigatures"],
+    frontendFunctions: ["alterText"],
+    name: "ddoouubblleedd",
+  },
 ];
 
 export default FunboxList;
--- a/backend/src/constants/funbox-list.ts
+++ b/backend/src/constants/funbox-list.ts
@@ -333,7 +333,7 @@ const FunboxList: MonkeyTypes.FunboxMetadata[] = [
   },
   {
     canGetPb: true,
-    difficultyLevel: 0,
+    difficultyLevel: 1,
     properties: ["noLigatures"],
     frontendFunctions: ["alterText"],
     name: "ddoouubblleedd",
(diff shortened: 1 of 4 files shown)


backend/src/constants/funbox-list.ts
```diff
@@ -331,6 +331,13 @@ const FunboxList: MonkeyTypes.FunboxMetadata[] = [
     canGetPb: true,
     difficultyLevel: 3,
   },
+  {
+    canGetPb: true,
+    difficultyLevel: 0,
```
Miodec, line 336: I would up the difficulty to 2.

Miodec (changes requested):

Miodec:
> * The words produced by the mode are not understandable when pronounced by TTS. However, setting the `unspeakable` property makes the mode incompatible with all other unspeakable modes, which is not desirable. I have not marked the mode as `unspeakable` to be consistent with the existing `backwords` rule, which is also not `unspeakable` despite producing unintelligible text.

Hmmm, this does sound kinda weird. Ill have a look if it makes sense to have only one `unspeakable` available at once.

> * This mode and all other modes that use `alterText`, such as `backwards`, `capitals`, and `rAnDoMcAsE`, are considered to be compatible with modes that overwrite the text using `pullSection` (`poetry` and `wikipedia`) despite being completely broken by these modes. `alterText` is never called for text generated by `pullSection`, and the modes have no effect. Fixing this should simply mean ensuring that `alterText` and `pullSection` modes are never used at the same time.

Also good point, thanks. Will add this.

Miodec:
Thanks!

PR #4795, opened 2023-11-15
feat: Update _list.json to add Portuguese from Portugal layouts, both ISO and ANSI (diogozarro)

diogozarro:
### Description

Added Portuguese from Portugal qwerty layouts, both ISO and ANSI
image

### DIFF
--- a/frontend/static/layouts/_list.json
+++ b/frontend/static/layouts/_list.json
@@ -1616,6 +1616,28 @@
     "row5": [" "]
     }
   },
+  "portuguese_pt_qwerty_iso":{
+    "keymapShowTopRow": false,
+    "type": "iso",
+    "keys": {
+      "row1": ["\\|", "1!", "2\"", "3#", "4$", "5%", "6&", "7/", "8(", "9)", "0=", "'?", "«»"],
+      "row2": ["qQ", "wW", "eE", "rR", "tT", "yY", "uU", "iI", "oO", "pP", "+*", "´`"],
+      "row3": ["aA", "sS", "dD", "fF", "gG", "hH", "jJ", "kK", "lL", "çÇ", "ºª", "~^"],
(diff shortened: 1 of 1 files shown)


frontend/static/layouts/_list.json
```diff
@@ -120,6 +120,28 @@
...
+    "keymapShowTopRow": false,
+    "type": "ansi",
+    "keys": {
+      "row1": ["\\|", "1!", "2\"", "3#", "4$", "5%", "6&", "7/", "8(", "9)", "0=", "'?", "<>"],
+      "row2": ["qQ", "wW", "eE", "rR", "tT", "yY", "uU", "iI", "oO", "pP", "+*", "´`", "~^"],
+      "row3": ["aA", "sS", "dD", "fF", "gG", "hH", "jJ", "kK", "lL", "çÇ", "ºª"],
+      "row4": ["zZ", "xX", "cC", "vV", "bB", "nN", "mM", ",;", ".:", "-_"],
+      "row5": [" "]
+    }
+  },
```
Miodec, line 144: Rename to `portuguese_qwerty_iso` and `portuguese_qwerty_ansi`.
Also please move lower in the file, next to the`swedish_qwerty`

Miodec (changes requested):

frontend/static/layouts/_list.json
diogozarro: There is more than one Portuguese layout it would be better to keep the "pt" to better distinguish so in the future we could have for instance:
portuguese_pt
portuguese_br
[Portuguese from Portugal] layout is different from [Portuguese from Brazil]:

image

image

Do you agree, or should I remove the pt as you suggested?

Miodec: Alright, thats fine.

Miodec:
Thanks

PR #4808, opened 2023-11-23
impr(language): replace kurdish central 45k with kurdish central 4k (Abdulbasit0123)

Abdulbasit0123:
The kurdish_central_45k was a mess, there were more 100 strange and unsupported characters to Kurdish Language and Kurdish Keyboard/Layout, and some of the _words_ were sentences actually, and many other problems. So, I decided to delete it, and then I added kurdish_central_4k which is much better, and it is what is intended to be.

Thanks 🤍


Miodec (approved):
Thanks

PR #4810, opened 2023-11-25
fix: crt mode not working on buttons

fehmer:
- apply box-shadow on buttons
- fix lavender theme hiding buttons on hover
- update crt mode specific themes not to use nested css


Miodec (approved):
Thanks

PR #4814, opened 2023-11-26
fix: prepone non-ascii characters removal for Wikipedia (FortPile)

Coadon:
### Description

Very simple.

When using the Wikipedia funbox, there would be occasional consecutive spaces (like '  '). As you would have guessed, the cursor glitched in the event of one.
This (from my observations) is caused by the algorithm after fetching having to remove all non-ASCII characters (most notably '–' involved with dates) AFTER cleaning whitespaces with the regex `/\s+/g`.
This process is now moved back.

I'm pretty sure this is the only source causing this issue. Otherwise, we could also clear all elements of `words` with an empty string after `const words = sectionText.split(" ");`.

### Checks

- [x] Make sure to include your GitHub username inside round brackets at the end of the PR title

### DIFF
--- a/frontend/src/ts/test/wikipedia.ts
+++ b/frontend/src/ts/test/wikipedia.ts
@@ -106,15 +106,16 @@ export async function getSection(language: string): Promise<Section> {
           // Remove invisible characters
           sectionText = sectionText.replace(/[\u200B-\u200D\uFEFF]/g, "");
 
+          // Remove non-ascii characters for English articles
+          if (urlTLD === "en") {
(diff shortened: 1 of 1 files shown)


Coadon:
Well, you see, replacing these characters with an empty string after converting all the whitespace to space results in potential consecutive spaces missing the purge.
Here's an example:
Consider the text `(21 April 1926 – 8 September 2022)`
Firstly, all whitespace (consecutive or single) is replaced with a single space which so far does nothing.
Then, non-ASCII characters are replaced with nothing, leaving `(21 April 1926  8 September 2022)`
Notice the double space missing the clean. This step should've been done earlier.

PR #4817, opened 2023-11-27
impr(layout emulator): add support for alt layer emulation (farkon00)

### LINKED ISSUE #4743: Caps Lock not working for non-ascii layouts in keyboard emulator

- [X] I have cleared my cache

N/A

Yes

N/A

Yes

### Account name

N/A

### Account config

_No response_

### Current Behavior

Currently Caps Lock does not apply to keys with non-ascii symbols in keyboard emulator layouts e. g. Ukrainian.

### Expected Behavior

Caps Lock applying to all layouts with capital letters provided

### Steps To Reproduce

1. Open Monkeytype
2. Open settings
3. Turn on the layout emulator and set it to Ukrainian
4. Go to zen mode
5. Turn on Caps Lock
6. Press the following keys: a, s, p.
7. The following characters are going to appear: ф, і, з. The expected behavior would be Ф, І and З. 

### Environment

- OS: Fedora Linux 37
- Browser: Firefox
- Browser Version: 117.0 (64-bit)

### Anything else?

The bug is almost certainly caused by use of `Misc.isASCIILetter` on [this line](https://github.com/monkeytypegame/monkeytype/blob/e3ce7b2458c6b6d52a402d402f7413d795fe675b/frontend/src/ts/test/layout-emulator.ts#L13). Also it would be nice to give more comments to that sections, because its meaning is quite non-obvious. Possible fix would probably include adding a parameter to layouts, that allows to mark some keys as punctuation.

Also sorry for half the report being N/A, but because I've found it initially in code, I don't think I really have to test it much.

farkon00:
### Description
* Added ability to add a 3rd and a 4th character for each key in the keyboard emulator layouts, which will be used with <kbd>AltGr</kbd> and <kbd>AltGr</kbd> + <kbd>Shift</kbd> respectively
* Added a polish programmers layout, which uses the previous change
* Added ability to type ґҐ by using <kbd>AltGr</kbd> + <kbd>Г</kbd> to the Ukrainian layout
* Fixes #4743, even though it's closed

### Notes
This pr only adds the polish programmers layout, because sadly the handling of diacritic modifier unicode characters in monkeytype and especially their rendering and updating is not polished enough to use it.

### Checks

- [x] Make sure to include your GitHub username inside round brackets at the end of the PR title


farkon00:
> Looks to be working fine with Alt and even Caps Lock.
> 
> Two questions:
> 
>    * Shouldn't left alt also work?
> 
>    * I can't get Shift Alt A to give me the accented A character, can you confirm that? Or is that my machine being weird?

I will test the second one, but first one is the intended behavior and it is how it works on every system. The left alt is used for application/OS shortcuts, meanwhile the right alt aka AltGr shouldn't be used by any program, so that you can type in any language.

farkon00:
I could not replicate the Shift + Alt + A issue, works completely fine for me.

Miodec:
> The left alt is used for application/OS shortcuts, meanwhile the right alt aka AltGr shouldn't be used by any program, so that you can type in any language.

I just enabled this layout on MacOS and can type ą with Left Option (aka Alt)

Miodec:
Yeah, not sure what to do in this case.

farkon00:
So should we merge?

Miodec:
Well, its not covering cases where the OS supports / does't support the Left Alt to type the accents.... Maybe we could do some the user agent to enable/disable the left alt?

farkon00:
I think using `window.navigator.platform` would work

farkon00:
> Well, its not covering cases where the OS supports / does't support the Left Alt to type the accents.... Maybe we could do some the user agent to enable/disable the left alt?

Done

Miodec:
Shouldn't windows also be included?

farkon00:
> Shouldn't windows also be included?

As far as I know windows only use AltGr for modifier keys, it's just a MacOS oddity.

farkon00:
So, are there any more changes required? I've already resolved all the merge conflicts.

Miodec:
> So, are there any more changes required? I've already resolved all the merge conflicts.

Sorry, I took a bit of a break and then got ill. Ill sort everything out soon.

frontend/src/ts/test/layout-emulator.ts
```diff
@@ -3,40 +3,26 @@ import * as Misc from "../utils/misc";
...
-    "KeyX",
-    "KeyC",
-    "KeyV",
-    "KeyB",
-    "KeyN",
-    "KeyM",
-  ].includes(keyLocation);
-}
+let isAltGrPressed = false;
+const isPunctuationPattern = /\p{P}/u;
```
Miodec, line 7: Just found this:

image


Should we be concerned?

farkon00: Don't think so, this solution seems to work better to me, because even if the layout has punctuation on some other keys it works. So using the old solution would, for example, break some Cyrillic layouts, where comma and period keys are actual letters.

Miodec: I mean the fact that it might not be supported in all browsers..

farkon00: [Compatibility on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode#browser_compatibility) seems  fine

Miodec (approved):
Thanks, sorry it took me a while.

PR #4824, opened 2023-11-30
impr: Add header validation, allow validation to accept unknown fields  (fehmer)

fehmer:
### Description

- add request validation for headers
- add option to allow unknown fields during validation
- move validation error message to validation options

### DIFF
--- a/backend/src/middlewares/api-utils.ts
+++ b/backend/src/middlewares/api-utils.ts
@@ -136,13 +136,13 @@ interface ValidationSchemaOption {
   allowUnknown?: boolean;
 }
 
-interface ValidationSchemaErrorMessage {
+interface ValidationHandlingOptions {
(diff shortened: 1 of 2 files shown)


backend/src/middlewares/api-utils.ts
```diff
@@ -129,11 +129,37 @@ interface ValidationSchema {
...
   query?: object;
   params?: object;
+  headers?: object;
+}
+
+interface ValidationSchemaOption {
+  allowUnknown?: boolean;
+}
+
+interface ValidationSchemaErrorMessage {
```
Bruception, line 139: Can we rename this interface to be something more generic?
Perhaps, ValidationHandlingOptions or something? What do you think?

Miodec (approved):

PR #4836, opened 2023-12-04
perf: Improve leaderboard performance by adding better index (fehmer)

fehmer:
### Description

Add a partial index for each leaderboard containing all users appicable for that leaderboard. The index covers all fields used by the query/result   (covered query), so .we don't need to fetch any documents. The index is pre-sorted by the correct fields (wpm, app, timestamp) so there is no sorting to be done during the query.


Miodec (approved):
Thanks

PR #4838, opened 2023-12-05
fix(quotes): english quote 5510 (brainandforce)

brainandforce:
### Description

I submitted English quote #5510 with one part of the first sentence missing, so I've included the missing bit.


Miodec (approved):
Thanks

PR #4840, opened 2023-12-05
test: Add test coverage for leaderboard update  (fehmer)

fehmer:
### Description

Add test coverage for leaderboard update.

### DIFF
--- a/backend/src/dal/leaderboards.ts
+++ b/backend/src/dal/leaderboards.ts
@@ -124,7 +124,7 @@ export async function update(
           },
         },
         {
-          $set: {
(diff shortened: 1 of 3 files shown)


backend/src/dal/leaderboards.ts
```diff
@@ -124,7 +124,7 @@ export async function update(
           },
         },
         {
-          $set: {
+          $addFields: {
```
fehmer, line 127: $set is an alias for $addFields but not supported in the mongodb version 4.0.3 we use for testing.

Miodec (dismissed):

PR #4841, opened 2023-12-05
perf: improve leaderboard aggregation (fehmer)

fehmer:
### Description

Update the leaderboard and public speedStats directly in mongodb without importing the data into the nodejs process first. 

### Performance measurements
Local measurements with 2.2M users,  ~230k users applicable to the leaderboard. 

- eliminates  ~400mb of network traffic between mongo and the node
- reduces CPU load on Mongo by ~50%
- reduces CPU load on the node process from ~40% to  less then 1%.
- reduce the overall time to update by ~40%

| **Task**  | **Old Time** | **New Time** |
|-----------|--------------|--------------|
| aggregate | 8.1s         | -            |
| loop      | 0.4s         | -            |
| insert    | 6.7s         | 9.2s         |
| index     | 1.1s         | ?            |
| histogram | 0.01s        | 0.1s         |
| **total** | **16.5s**    | **9.75s**    |

### DIFF
--- a/backend/__tests__/dal/public.spec.ts
+++ b/backend/__tests__/dal/public.spec.ts
@@ -1,23 +1,4 @@
 import * as PublicDAL from "../../src/dal/public";
-import * as db from "../../src/init/db";
-import { ObjectId } from "mongodb";
-
-const mockSpeedHistogram = {
-  type: "speedStats",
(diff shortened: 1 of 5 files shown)


Miodec (approved):
yolo

PR #4863, opened 2023-12-14
feat: Indicate premium users  (fehmer)

fehmer:
### Description

Indicate premium users
- additional badge in daily and all time leaderboard 
image
image

- additional badge in public profile 
image

- highlight username in header 
image
- add badge icon to result screenshot
image


Miodec:
Im not sure about forcing the user to display that badge full size on their account page. I would just handle that like any other badge.

fehmer:
Summary of the discussion in discord:
- add premium badge to user.inventory.badges on purchase
- remove premium badge on subscription deleted webhook from stripe
- add cron job to remove premium badge from inventory if `expirationTimestamp<now`
- LB/Daily add premium badge as mini badge if not selected. Add property "important" to badges to trigger this behaviour.
- remove changes in public profile
- screenshot_ use `| badge.name` instead of the icon

PR #4864, opened 2023-12-14
fix: Proper handling of german umlaute in lazy mode (fehmer)

fehmer:
### Description

German umlaute `ä` , `ö` and  `ü` are getting replaced with `ae`, `oe` and `ue`. For capital letters only the first character is capitalized.

Example: 

`Üben Äpfel Öde trä­nen­über­strömt äße` 
should be replaced with
`Ueben Aepfel Oede traenenueberstroemt aesse` 

> When it is not possible to use the umlauts (for example, when using a restricted character set) the characters ⟨Ä, Ö, Ü, ä, ö, ü⟩ should be transcribed as ⟨Ae, Oe, Ue, ae, oe, ue⟩ respectively, following the earlier postvocalic-⟨e⟩ convention; 
See [German orthography on wikipedia](https://en.wikipedia.org/wiki/German_orthography#Umlaut_diacritic_usage)


frontend/src/ts/test/lazy-mode.ts
```diff
@@ -1,9 +1,12 @@
...
   ["íìîïĩįį́īį̄ı", "i"],
-  ["óòôöøõóōǫǫ́ǭő", "o"],
-  ["úùûüŭũúūůű", "u"],
+  ["óòôøõóōǫǫ́ǭő", "o"],
+  ["ö", "oe"],
+  ["úùûŭũúūůű", "u"],
+  ["ü", "ue"],
   ["ńňñ", "n"],
   ["çĉčć", "c"],
   ["řŕṛ", "r"],
```
Miodec, line 12: I would use the langauge specific lazy mode definitions, other languages might use the sam character and they need to be replaced by just the one character.

Miodec: Example: 
image

fehmer:
@Miodec i noticed the override in the language file replaces the rules defined in the `lazy-mode.ts`. At least for german it would be better to have the rules in addition to the default once. E.g. if a german text contains an `é` it should still be `e` in lazy mode.
Do you know if this also applies to the other languages, so do we need both variants.
If we ne both we could specify e.g.  `additionalAccents` in the language files and change the code to merge it with the default accents (additional first).

Miodec:
> @Miodec i noticed the override in the language file replaces the rules defined in the `lazy-mode.ts`. At least for german it would be better to have the rules in addition to the default once. E.g. if a german text contains an `é` it should still be `e` in lazy mode.

Whats wrong with just redefining whats needed?

fehmer:
> > @Miodec i noticed the override in the language file replaces the rules defined in the `lazy-mode.ts`. At least for german it would be better to have the rules in addition to the default once. E.g. if a german text contains an `é` it should still be `e` in lazy mode.
> 
> Whats wrong with just redefining whats needed?

I'd copy all default accents in 4 german languages files now. Next time you add something to the default accents you'd need to remember to put the same in all the other files.
I guess the hard part is to figure out what is needed. I played around with lazymode+wikipedia funbox and there you can have e.g. french names/words mixed in german articles. So I'd still would need the french rules in my german lazy mode.

To come back to the initial question. Only "pinyin" is currently using the  override. Do you remember if this language intentionally not using the default accent?

To support both cases and don't need to copy the whole default accent rules to a language with additional rules we could either 

1. add `additionalAccents` as `[string,string][]` to the language. If this is set the `accents` is ignored or
2. add `additionalAccents` as `boolean` and if set the rules are mixed with the defaults.

Miodec:
> I'd copy all default accents in 4 german languages files now. Next time you add something to the default accents you'd need to remember to put the same in all the other files.

~~Not really. We only need to define the accents for a language once. The only time we would need to copy paste is when adding a variation (for exmample german 10k). Once the accent list for a given language is finised, there is no need to modify it, even if the global list is changed.~~

Ok. I think i see the use case. Lets just switch it to extend the default list, instead of overriding.

fehmer:
I refactored the code to uppercase the replaced accents because it broke with accent replacements containing of multiple chars

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/7355442592

PR #4872, opened 2023-12-17
fix: update the background image and hide the scrollbar in the funbox space_balls (yunfachi)

yunfachi:
### Checks

The first commit updates the background image URL, resolving an issue where the previous GIF was not loading due to the gfycat service being discontinued.
The second commit hides the Y-axis scrollbar that appears after rotating the page using `transform: rotateX(35deg);`.

### DIFF
--- a/frontend/static/funbox/space_balls.css
+++ b/frontend/static/funbox/space_balls.css
@@ -17,9 +17,10 @@ body {
 }
 
 main {
-  transform: rotateX(35deg);
+  perspective: 500px;
+  overflow: hidden;
 }
 
-#contentWrapper {
-  perspective: 500px;
+main .page {
+  transform: rotateX(35deg);
 }
--- a/frontend/static/funbox/space_balls.css
+++ b/frontend/static/funbox/space_balls.css
@@ -11,7 +11,7 @@
 }
 
 body {
-  background-image: url("https://thumbs.gfycat.com/SlimyClassicAsianconstablebutterfly-size_restricted.gif");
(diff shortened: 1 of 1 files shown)


Miodec:
Which scrollbar are you talking about? Not sure if its needed.

yunfachi:
> Which scrollbar are you talking about? Not sure if its needed.

The page length increases when using the 'transform rotate' property. This is because the 'transform rotate' is applied to the 'main' element, causing its lower part to come closer to the screen. As a result, the lower part occupies more space on the page, leading to an increased page size and the appearance of a scrollbar. My current solution of using 'overflow-y: hidden' is somewhat crude, as it suppresses the scrollbar. An alternative approach is to apply the 'transform rotate' to the '#typingTest' div, which would solve the problem but might significantly alter the appearance of 'space_balls'.
### without rotation (the page cannot be scrolled down)
without rotation
### with rotation (the page can be scrolled down, but only the background will be visible)
with rotation

Miodec:
Right, i zoomed my browser in a bit and can see the scrollbar - but how are you going to reach the footer if you disallow scrolling?

Miodec:
So.. is the overflow change really needed then? It looks exactly the same as live.

yunfachi:
> So.. is the overflow change really needed then? It looks exactly the same as live.

It doesn't work for me without overflow. 1920x1080, firefox 121.0b5
In live I can scroll down like in the screenshot

Miodec (approved):

PR #4880, opened 2023-12-20
refactor: import jquery and jquery plugins using webpack (fehmer)

fehmer:
### Description

Define jquery and jquery plugins as dependencies instead of having a copy in the `static` directionary.

@Miodec What is your take on this? It would reduce the number of files to be loaded but with each release the client would download the unmodified dependencies again.


Miodec:
https://media1.tenor.com/m/SZ2sXVtNC3MAAAAC/spongebob-meme-spongebob.gif

PR #4920, opened 2024-01-13
impr(themes): Added a new theme pastel-harmony

Anasdevs:
I've included the theme colors in the list _list.json, also i have added a new file pastel_harmony.css to the themes folder and added the theme code. 
pastel harmony theme is clear and readable with both flip test colors and colorful mode enabled and disabled
I've checked this theme doesn't look similar to any other existing themes.
image
image
image


Miodec:
Sorry, the theme is not unique enough to be added - its simmilar to other themes already available.

PR #4941, opened 2024-01-17
fix: show correct 'come back in' text on profile hover (L-E-Freeman)

### LINKED ISSUE #4735: Streak: "come back in" text is incorrect
From Discord: 
```
Anyone else's streak ignores the hour offset after the first day? I reported mi streak doing this via mail and the offset was cleared so I could set it up again, but it only lasted one day, then started resetting at midnight UTC. I've tested with incognito mode, another fresh browser and cleared the cookies, all with the same results, I just unlinked from Discord and will test in an hour.

Any other ideas?

12:10am (UTC-6) Edit: didn't work, will see tomorrow 
Waruiiko — Today at 2:44 AM
7:43 pm (UTC-6) and I got the daily bonus, so it keeps resetting at midnight UTC :c
```

L-E-Freeman:
### Description

This PR hopefully fixes the issue with the 'come back in' hover text not showing the correct number of hours/days. 

I've (hopefully) simplified things a little bit. I think the issue was that `Misc.getCurrentDayTimestamp()` sometimes got yesterday's date and sometimes today's date depending on the offset and the current time, leading to inconsistency. 

My solution instead gets the timestamp at the start of the current day, and then calculates the hour of reset between 0 and 24 depending on what offset the user has set. This hour is applied to the current day and either results in a reset timestamp that has already been passed (in which case we add a day to get the next reset), or has not been passed, in which case we just get the formatted distance to that reset time.

I did a bit of testing with both my own timezone and the timezone of the user that originally reported the problem (UTC-6), and it looks like it's working for me, hoping you see the same.

### Checks

- [X] Adding a language or a theme?
    - [X] If is a language, did you edit `_list.json`, `_groups.json` and add `languages.json`?
    - [X] If is a theme, did you add the theme.css?
- [X] Check if any open issues are related to this PR; if so, be sure to tag them below.
- [X] Make sure to include your GitHub username inside parentheses at the end of the PR title

Closes #4735

### DIFF
--- a/frontend/src/ts/elements/profile.ts
+++ b/frontend/src/ts/elements/profile.ts
@@ -131,26 +131,32 @@ export async function update(
 
     const dayInMilis = 1000 * 60 * 60 * 24;
     const milisOffset = (profile.streakHourOffset ?? 0) * 3600000;
-    let target = Misc.getCurrentDayTimestamp() + dayInMilis + milisOffset;
-    if (target < Date.now()) {
-      target += dayInMilis;
-    }
-    const timeDif = formatDistanceToNowStrict(target);
 
+    // Get the current day in local time at 00:00
+    const startOfDay = new Date();
+    startOfDay.setHours(0, 0, 0, 0);
+
+    // Add our offset hours
+    let resetTarget =
+      startOfDay.getTime() + Misc.getLocalTimeStreakResetHour(milisOffset);
+
+    // If the target is in the past, the next reset is tomorrow, so we add a day
+    if (resetTarget < Date.now()) {
+      resetTarget += dayInMilis;
+    }
+
+    const timeDif = formatDistanceToNowStrict(resetTarget);
+    console.debug("startOfDay", startOfDay);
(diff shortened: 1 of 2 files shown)


Miodec:
I found a quicker fix:

image
image

What do you think? 

I confirmed it works in both cases (negative utc with positive offset and positive utc with negative offset)

Miodec:
> Yeah checks out in my testing too 👍 Unfortunate I missed that, I think when I tested adding the optional `offsetMilis` argument I'd already made modifications elsewhere that fudged things.

What should we go with then?

L-E-Freeman:
Your fix looks more straightforward and I think accomplishes the same thing so IMO we should go with that. 

Do you want me to modify this PR or will you make a separate commit?

Miodec:
> Your fix looks more straightforward and I think accomplishes the same thing so IMO we should go with that.
> 
> Do you want me to modify this PR or will you make a separate commit?

You could also close the PR and I will commit directly to master. Up to you.

PR #4945, opened 2024-01-19
impr(quotes): add English quotes

iwmywn:


Miodec:
Did you really just add 1000 quotes?

iwmywn:
I'm pretty careful so the likelihood of a coincidence is almost nil, it's mostly motivational sentences, you can read

Miodec:
Well, the problem is that I really don't want to have to read through 1000 quotes to verify things..

iwmywn:
I dont want to have to open pull requests multiple times, but if you want easier verification then i can break down about 50 quotes at a time.

Miodec:
That could make it more managable

Issue #4823, opened 2023-11-30
Personal record category not different when numbers selected


predkambrij:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

No

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

No

### Account name

predkambrij

### Account config

[...]

PR #4946, opened 2024-01-19
impr: consider numbers in pb

notreallystatic:
**Fixes**:
https://github.com/monkeytypegame/monkeytype/issues/4823


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/7578828078

Miodec:
Updating the types is not enough. You also need to handle this new property correctly when saving pbs, and make sure to display it on the frontend.

notreallystatic:
@Miodec  Yeah, the pbs will be saved. Also, in UI it will be shown as a new personal best cause we will set the flag as true in response. Are there any other places in the UI, where we are showing the score explicitly that I need to handle? What else I am missing?

Miodec:
> @Miodec Yeah, the pbs will be saved. Also, in UI it will be shown as a new personal best cause we will set the flag as true in response. Are there any other places in the UI, where we are showing the score explicitly that I need to handle? What else I am missing?

`pb-tables-popup.ts`

image

frontend/src/ts/db.ts
```diff
@@ -636,6 +637,7 @@ export async function saveLocalPB<M extends MonkeyTypes.Mode>(
     ).forEach((pb) => {
       if (
         pb.punctuation === punctuation &&
+        !!pb.numbers === numbers &&
```
Miodec, line 640: I dont think `!!` is necessary here

notreallystatic: I checked that the older personal-bests in a given mode won't have the numbers field. That's why I added this.  Btw, I am also adding support for pace-caret and tags, so still working on that. Could you tell if I am missing anything else as well?

Miodec: Ah yeah, youre right. All though maybe i would prefer `(pb.numbers ?? false) === number`

PR #4958, opened 2024-01-23
build: enable noUncheckedIndexedAccess compiler rule

Miodec:
Ouch


backend/__tests__/dal/result.spec.ts
```diff
@@ -77,7 +77,7 @@ describe("ResultDal", () => {
 
       //THEN
       expect(results).toHaveLength(10);
-      let last = results[0].timestamp;
+      let last = results[0]?.timestamp as number;
```
fehmer, line 80: hm, is it not `number|undefined` ?

Miodec: Casting to remove the compile errors. This PR mainly is just to remove existing errors without modifying the logic too much (like adding returns and notifications that something is undefined).

PR #4963, opened 2024-01-24
fix(docker-compose): use the user "node" for the frontend container (dirk-hub)

dirk-hub:
### Description

add `user: node` to the `docker-compose.yaml` so docker uses that user instead of the root user when bringing up the container, this makes npm have appropriate rights

### Checks

- [ ] Adding quotes?


Miodec:
Thanks

PR #4992, opened 2024-02-01
refactor: enable strict-boolean-expressions rule

Miodec:
### Description

Should help catch some of the bugs where an optional string values of undefined and empty string should be handled differently. Probably helps in other cases too.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/7748362485

PR #4995, opened 2024-02-02
chore: allow docker compose to start the backend server (plbstl)

plbstl:
### Description

The [Backend section of the Advanced Contribution Guide](https://github.com/monkeytypegame/monkeytype/blob/1cfce348a102850e9baabb936cecc5ef6604d634/CONTRIBUTING_ADVANCED.md#backend-optional) isn't entirely true. You cannot use `docker compose up` to start the backend server. The command only starts redis and mongodb services, while having to manually run one of the dev scripts to start the backend server.

This change makes sure that `docker compose up` in the backend folder also starts the server, making contributions to this project a bit more seamless.

I had to use a separate volume for the *node_modules* in the backend because of mismatch between where the bcrypt module is built (my PC - MacOS) and where it is being run (Docker - Linux). It creates an error that prevents the server from booting up.

```yml
services:
  #...
  api-server:
    #...
    volumes:
      #...
      - /monkeytype/backend/node_modules
    entrypoint: 'bash -c "cd /monkeytype/backend && npm install && npm run dev"'
```

Only the backend dependencies are installed and built in the Docker container. The script can be a bit more complicated to skip unnecessary npm installs, but this is also okay.

### Checks


Miodec:
My one issue with this PR is that I personally use Docker to only host the databases, and then run the servers in node myself. Maybe we can have one file for just databases and one for databses and node server?

plbstl:
> Maybe we can have one file for just databases and one for databases and node server?

Yes, that can work. Maybe use the [COMPOSE_FILE](https://docs.docker.com/compose/environment-variables/envvars/#compose_file) env var to determine the file to use

plbstl:
Yes, we're using multiple files. The env var is not necessary but a nice to have. Now thinking about it, it may not even be needed

Miodec:
> Yes, we're using multiple files. The env var is not necessary but a nice to have. Now thinking about it, it may not even be needed

I mean multiple files for composing backend:
 - one that only runs the databasees
 - one that runs both databases and the api server

plbstl:
> > Maybe we can have one file for just databases and one for databases and node server?
> 
> Yes, that can work. Maybe use the [COMPOSE_FILE](https://docs.docker.com/compose/environment-variables/envvars/#compose_file) env var to determine the file to use

Let me elaborate on what I meant by this. 

There will be two docker-compose files in the backend. When someone wants to contribute and they run `docker compose up`, the env var will instruct docker on which file to use. So, there is a default file being used. 

> Yes, we're using multiple files. The env var is not necessary but a nice to have. Now thinking about it, it may not even be needed

We know by default, docker compose checks for `docker-compose.yaml`. So the default file that is supposed to be set in the [COMPOSE_FILE](https://docs.docker.com/compose/environment-variables/envvars/#compose_file) env var can be named `docker-compose.yaml`, and the other one named something more specific which will be run with `docker-compose -f <filename> up`

backend/example.env
```diff
@@ -1,4 +1,5 @@
 DB_NAME=monkeytype
+DOCKER_SERVER_PORT=5005
```
Miodec, line 2: Whats the point of having a docker port variable when the port is hardcoded in the compose file?

Miodec:
> > > Maybe we can have one file for just databases and one for databases and node server?
> > 
> > 
> > Yes, that can work. Maybe use the [COMPOSE_FILE](https://docs.docker.com/compose/environment-variables/envvars/#compose_file) env var to determine the file to use
> 
> Let me elaborate on what I meant by this.
> 
> There will be two docker-compose files in the backend. When someone wants to contribute and they run `docker compose up`, the env var will instruct docker on which file to use. So, there is a default file being used.
> 
> > Yes, we're using multiple files. The env var is not necessary but a nice to have. Now thinking about it, it may not even be needed
> 
> We know by default, docker compose checks for `docker-compose.yaml`. So the default file that is supposed to be set in the [COMPOSE_FILE](https://docs.docker.com/compose/environment-variables/envvars/#compose_file) env var can be named `docker-compose.yaml`, and the other one named something more specific which will be run with `docker-compose -f <filename> up`

In that case, i would move the compose files to a `backend/docker` and rename them to `db-and-server` and `db-only`. Then add `npm run docker` and `npm run docker-db` scripts.

Miodec (approved):

PR #5012, opened 2024-02-07
Create code_css.json impr(quotes)

g1gabyteDEV:
### I added a quote for code-css which says
`
.container {display: flex; justify-content: center; align-items: center; height: 100vh; flex-direction: row;}
`
### because there was no css quotes

### DIFF
--- a/frontend/static/code_css.json
+++ b/frontend/static/code_css.json
@@ -1,0 +1,17 @@
+{
+  "language": "code_css",
+  "groups": [
+    [0, 100],
+    [101, 300],
+    [301, 600],
+    [601, 9999]
+  ],
+  "quotes": [
+  {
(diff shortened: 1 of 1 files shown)


frontend/static/code_css.json
```diff
@@ -0,0 +1,17 @@
...
+  "groups": [
+    [0, 100],
+    [101, 300],
+    [301, 600],
+    [601, 9999]
+  ],
+  "quotes": [
+  {
+    "text": ".container {display: flex; justify-content: center; align-items: center; height: 100vh; flex-direction: row;}",
+    "source": "puts things in center using flexbox :)",
```
Miodec, line 12: can we have the source be more "official" sounding? Also i would start with a bit more quotes.

g1gabyteDEV: ok :)

PR #5036, opened 2024-02-10
docs: correct navigation of Firebase > Build > Authentication > Sign-in UI (underscoore)

underscoore:
### Description

The Firebase configuration section in [advance contribution](https://github.com/monkeytypegame/monkeytype/blob/master/CONTRIBUTING_ADVANCED.md) is not up to dated as per the [Firebase](https://firebase.google.com/) UI. The path to navigate to the `Authentication` UI is not clear even to the person who is setting put for the first time.

This change makes sure that new contributor can setup firebase easily or at least navigate easily,


Miodec:
Thanks

PR #5057, opened 2024-02-14
feat(theme): Added matcha theme (TheExploration)

TheExploration:
* Create matcha.css

* Update _list.json

purposefully made background and sub alt the same color as it looks nice but if that is an issue I will change it.

image
image
image
image


Miodec:
- Text color must be close to white/black.
 - sub alt color cannot be the same as sub alt
 - seems to be the same idea as the `matcha moccha` theme

PR #5063, opened 2024-02-15
feat(theme): Adding a turquoise theme

CedricBaaklini:
I would like to create a turquoise based theme as turquoise is one of my favorite colors and would like to apply it to directly to MonkeyType. I also added a few quotes from famous pieces of media in the process.

It looks a little something like this:
Screenshot 2024-02-15 at 01 53 07


Miodec:
Text color cannot be colored.
Sub alt color must be different to bg color.
PRs cannot have unrelated changes (quotes in this case)

CedricBaaklini:
I changed some of the colors of the theme and it should look a little bit better now

Screenshot 2024-02-21 at 15 36 28
Screenshot 2024-02-21 at 15 37 06

Miodec:
Sorry, I have to be selective with the themes because there are so many - Im not sure i like the overall color choice here.

CedricBaaklini:
That is fine. You do you.

PR #5067, opened 2024-02-16
feat(theme): Add milk tea theme (TheExploration)

TheExploration:
- Create milk_tea.css
- Update _list.json

> Text color must be close to white/black. 

-fixed

> sub alt color cannot be the same as sub alt

 -fixed

> seems to be the same idea as the matcha moccha theme

similar idea but I think it's unique enough to be its own theme. This is more "matcha" then the matcha moccha theme. I'd say leather theme is more similar to matcha moccha then this is. Changed title to "milk tea" to differentiate from matcha moccha.

Screenshot 2024-02-15 190040
Screenshot 2024-02-15 190125
Screenshot 2024-02-15 190143
Screenshot 2024-02-15 190156


Miodec:
This theme was already rejected due to its similarity with matcha moccha

PR #5074, opened 2024-02-17
impr: provide all-time LB results during LB update (fehmer)

fehmer:
Try to provide LB results during the LB update. There is a very small time-frame where already running queries might fail during the update. For now we keep the 503 error in this cases and monitor how often this happens on production.


Miodec (approved):
Thanks

PR #5085, opened 2024-02-19
feat: add testWords and wordsHistory to copy result stats

fehmer:
### Description

Add the test words and input history to "copy result stats" data:

```
{
    "lastResult": {...},
    "wordsHistory": ["people", "little"],        // words to be typed
    "inputHistory": ["poelpe", "little"]    // words actually typed
}
```

### DIFF
--- a/frontend/src/ts/commandline/commands.ts
+++ b/frontend/src/ts/commandline/commands.ts
@@ -101,8 +101,6 @@ import * as ShareTestSettingsPopup from "../popups/share-test-settings-popup";
 import * as TestStats from "../test/test-stats";
 import * as QuoteSearchPopup from "../popups/quote-search-popup";
 import * as FPSCounter from "../elements/fps-counter";
-import * as TestInput from "../test/test-input";
-import * as TestWords from "../test/test-words";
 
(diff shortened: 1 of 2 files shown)


frontend/src/ts/commandline/commands.ts
```diff
@@ -420,7 +422,16 @@ export const commands: MonkeyTypes.CommandsSubgroup = {
...
+          .writeText(
+            JSON.stringify({
+              ...(TestStats.getStats() as object),
+              testWords: TestWords.words.list.slice(
+                0,
+                TestInput.input.history.length
+              ),
+              wordsHistory: TestInput.input.history,
+            })
+          )
```
Miodec, line 434: I would modify the return object of the `getStats()` function instead of modifying it here.

Miodec (changes requested):
> {
    "lastResult": {...},
    "testWords": ["people", "little"],        // words to be typed
    "wordsHistory": ["poelpe", "little"]    // words actually typed
}

Rename to `wordsHistory` and `inputHistory`

Miodec (approved):
Thanks

PR #5088, opened 2024-02-19
feat(theme): Add macroblank theme (TheExploration)

TheExploration:
- Create macroblank.css
- Update _list.json

image
image
image

colorful mode
image
normal 
image


Miodec:
Thanks

PR #5093, opened 2024-02-20
fix: don't scroll page if a popup is present (fehmer)

fehmer:
### Description

Fix page (content in the backgroud) scrolling if a popup is open.


Miodec:
Causes the page to jump to the top when opening a popup. Breaks vertical ads in the background. 

We need to find a different fix.

image

Issue #5107, opened 2024-02-22
Replace $.each calls with for loops


monkeytypegeorge:

Miodec:
Assuming you use VSCode, just use the search panel:

![image](https://github.com/monkeytypegame/monkeytype/assets/13181393/1d5875fb-66e9-4a2c-a18e-219168ef4e43)

mokshsinghji:
I can help make the changes as well. Just wanted to clarify, do you want `foreach` loops, or simple `for` loops?

FFUV:
> Assuming you use VSCode, just use the search panel:
> 
> ![image](https://private-user-images.githubusercontent.com/13181393/307073387-1d5875fb-66e9-4a2c-a18e-219168ef4e43.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDg3ODgwMzEsIm5iZiI6MTcwODc4NzczMSwicGF0aCI6Ii8xMzE4MTM5My8zMDcwNzMzODctMWQ1ODc1ZmItNjZlOS00YTJjLWExOGUtMjE5MTY4ZWY0ZTQzLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAyMjQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMjI0VDE1MTUzMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTYyOGYwNzkyNTY4OTllMmFjZDdmN2ZkOTQ5OTVjNGQ3NjFjNjcxYThmODQzNmFlYjk3MmQ0NGQyZTM3NGYyMWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.rEwpfaeC77vUm9GvyLAO3AYg-XKjS6Yu6_yTjuzeFIo)

[...]

PR #5112, opened 2024-02-23
refactor: replace $.each calls with for loops (underscoore)

underscoore:
### Description

This PR is only for the review purposes:

Changes are made only in [ROOT]\frontend\src\ts\commandline\index.ts

Related to the issue: [Issue#5107](https://github.com/monkeytypegame/monkeytype/issues/5107)


Miodec:
Thanks

PR #5119, opened 2024-02-24
fix: personal bests not shown correctly on user tags (fehmer)

### LINKED ISSUE #5115: Identify why tags are showing new pb even when its not true

fehmer:
### Description

fixes #5115


Miodec (commented):
Damn, we missed a lot from the numbers pr.

Thanks.

PR #5124, opened 2024-02-25
style: Improving monkey.scss (FFUV)

FFUV:
Adjusted the height of the #monkey container to match the height of its child div. Changed the position of the div from fixed to absolute, which is more appropriate for the animation. Added background-repeat: no-repeat; and background-position: center; to ensure the monkey images are displayed correctly within their containers. Reduced the animation duration to 0.1s for a smoother motion. Adjusted the keyframes for the shake animation to provide a subtle shaking effect

### DIFF
--- a/frontend/src/styles/monkey.scss
+++ b/frontend/src/styles/monkey.scss
@@ -1,37 +1,63 @@
 #monkey {
   width: 308px;
-  height: 0;
+  height: 200px; /* Adjusted height to match the div */
   margin: 0 auto;
-  animation: shake 0s infinite;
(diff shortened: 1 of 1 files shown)


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/8039320183

Miodec (commented):
Height change is unwanted because it makes the words jump up.
Animation change is not needed because its dynamically adjusted by javascript.

Thanks, but these changes are not needed.

PR #5125, opened 2024-02-25
impr(quotes): add Dutch quotes (devdenisalex)

devdenisalex:
Quote 1 Translation:
If you want to be a rock star or just be famous, then run down the street uncovered, you'll make the news or something. But if you want music to be your livelihood, then play, play, play and play! And eventually you'll get to where you want to be.

Quote 2 Translation:
The most important thing about a person is that you know who you are and what you want. You should be capable of introspection and evaluate yourself.

Quote 3 Translation:
Sometimes the most important thing in a whole day is the rest we take between two deep breaths.


FFUV (approved):
lgtm!

PR #5137, opened 2024-02-26
impr(languages): replace 2 fake code_vim commands with real ones (penguin-teal)

penguin-teal:
### Description

Replace 2 fake commands in the `code_vim` words with real ones.


Miodec:
Ugh, thanks

PR #5139, opened 2024-02-26
impr(quotes): Added Python quote (OwenJRJones)

OwenJRJones:
### Description

Added Disktra's algorithm implementation in Python.

Source: https://www.geeksforgeeks.org/python-program-for-dijkstras-shortest-path-algorithm-greedy-algo-7/


CedricBaaklini:
That is perhaps one of the most creative quotes that I've ever seen here. Oh my god........

Miodec (approved):

PR #5141, opened 2024-02-26
impr(quotes): add php quotes (NipulM)

NipulM:
I've included a couple of sorting algorithms (Selection and Bubble sort) and a couple of basic php code snippets in this pull request.


Miodec:
I would start with a few more quotes to avoid repetition.

PR #5147, opened 2024-02-27
fix: remove entries with spaces from english_5k (lynn)

lynn:
### Description

All of these seem to be words covered by the rest of the file.

To my surprise, these entries don't seem to cause e.g. 50-word tests to be actually 51 words... but my rationale is I use monkeytype to practice stenography, and then it feels like cheating to be dealt a test with `have to` in it, which is one stroke, and so I'm really doing a 49-word test in a way. (Of course I can still get `have` and then `to` in a test if I'm lucky, but I figure there's no reason to skew the distribution towards it.)

### Checks

- [ ] Adding quotes?


Miodec:
Its not really an issue - these automatically get split, so the word count never goes wrong. I don't really consider it cheating either - in real typing you will sometimes run into these pairs of "often grouped" words, which will be easier to type due to increased muscle memory.

PR #3074, opened 2022-06-06
update security.txt expiry date

teh-maxh:
### Description

`security.txt` had an expiry date of 2022-06-03T21:00:00.000Z. This moves the expiry date one year later.

PR #5156, opened 2024-02-29
chore: expire security.txt 2024-06-03

teh-maxh:
Changing the expiry date for security.txt to something that isn't already passed.

Unfortunately, the RFC requires having an expiry date. Long-term, there should probably be a better solution than someone manually updating this (or not) every year, but as it's technically broken now I've stuck with the quickest fix again (cf #3074).


Miodec:
Thanks

PR #5157, opened 2024-02-29
chore: improve docker files for local development (fehmer)

fehmer:
### Description

fix docker compose files

- use `docker compose` instead of the outdated `docker-compose`
- fixed port mapping
- don't share `node_modules`  directories outside docker to prevent permission issues
- fix frontend not able to build if root directory `node_modules` missing


Miodec:
Thanks

PR #5162, opened 2024-02-29
refactor: modal system

Miodec:
### Description

Provide a class which handles all of the animation for modals. Should greatly reduce code duplication. Also uses native `dialog` element for proper focus trapping.

Still need to convert all the rest of the popups.


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/8103013347

PR #5164, opened 2024-03-01
fix(commandline): Exact matches are selected (penguin-teal)

penguin-teal:
Because of the RegEx, typing a command such as the language "code c" selects the item "code python" because the word "code" is being matched twice. This change makes it so that when an exact match of a `display` field is typed, it is selected (when I type "code c" I expect to be able to just hit enter but it selects "code python").


Miodec (approved):
Thanks. I do have a commandline rewrite planned, so this code will change - but Ill still merge it in for the meantime.

PR #5169, opened 2024-03-03
impr: show total completed test percentage and restarts/test on profile (fehmer)

fehmer:
### Description

Add mouseover with percentage of test completed and restarts per completed test to the overall typing stats on profile and account page.


Miodec (approved):
Thanks

PR #5201, opened 2024-03-08
feat: add klingon language (fehmer)

fehmer:
### Description

Add klingon language.


FFUV:
"the constructed language spoken by a fictional alien race called the Klingons, in the Star Trek universe."

Maybe you should add this to funbox instead it would make more sense

Miodec (changes requested):
Please order list and groups file roughly by the language popularity.

Miodec (approved):
Thanks

PR #5202, opened 2024-03-08
fix(language): remove capital letters in russian_10k.json (zaqzxcswsde)

Zaqzxcswsde:
While it's correct for the words "Москва" and "Россия" to be capitalized, they seem out of place being that way among ~10k lowercase words.


Miodec (approved):

Miodec:
Thanks

PR #5203, opened 2024-03-08
feat: add ook coding language (fehmer)

fehmer:
#### Introduction
Ook! is a programming language designed for orang-utans. Ook! is essentially isomorphic to the well-known esoteric language [BrainF***](http://www.muppetlabs.com/~breadbox/bf/), but has even fewer syntax elements.

#### Design Principles
- A programming language should be writable and readable by orang-utans.
- To this end, the syntax should be simple, easy to remember, and not mention the word "monkey".
- Bananas are good.

### DIFF
--- a/frontend/static/languages/_groups.json
+++ b/frontend/static/languages/_groups.json
@@ -561,7 +561,8 @@
       "code_gdscript",
       "code_gdscript_2",
       "code_assembly",
-      "code_v"
+      "code_v",
+      "code_ook"
     ]
   }
 ]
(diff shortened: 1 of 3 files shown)


FFUV:
this would  really have no use because theres no use for ook just like how Brainf*** has no use besides being like some aesthetic

FFUV:
and its also quite litteraly a joke language https://esolangs.org/wiki/Ook!

FFUV:
I have done some considering and realised I was incorrect although it has no actual use this is a **CUSTOMIZABLE** minimal typing site after all

Miodec (approved):
Ook.

PR #5227, opened 2024-03-15
fix: enter quick restart not working on test page (fehmer)

fehmer:
### Description

fix `enter` not restarting the test while on test page


Miodec (approved):
Oops, thanks.

PR #5231, opened 2024-03-15
feat(language): add Xhosa (de-longhi)

de-longhi:
### Description

After many mistakes, I added support for the Xhosa (isiXhosa) language.


Miodec:
Thanks

PR #5237, opened 2024-03-19
fix: quote search modal (fehmer)

fehmer:


Miodec:
Actual fix in this case is making the parameter optional, but thanks.

PR #5241, opened 2024-03-22
impr(quotes): add English quote (Oirehm)

oirehm:
### Description

Added quote #7039

And then I came to a realization that you and death will never intersect. As long as you're here, death is not. As long as death is there, you are not.
- Norm Macdonald

English


Miodec:
This PR is modifying the while quotes file (most likely changing indentation) instead of just adding one quote. Please open a new PR.

PR #5248, opened 2024-03-23
style: Change csv file name when saving data (vjgtigers)

vjgtigers:
### Description

when downloading history it saves as a csv file but the data is not separated by commas. while most csv viewers can still read it, it could cause a little bit of confusion. This pull request changes the file extension to dsv (delimiter separated values) and changed  "Export CSV" to "Export History".  However changing "item.charStats" and "item.tags" to be joined by a different delimiter and changing the main delimiter to commas would also work and could keep the csv extension.

### DIFF
--- a/frontend/src/html/pages/account.html
+++ b/frontend/src/html/pages/account.html
@@ -601,7 +601,7 @@
     <div class="group aboveHistory">
       <div class="button exportCSV">
         <i class="fas fa-file-csv"></i>
-        Export CSV
+        Export History
       </div>
     </div>
     <div id="ad-account-2-wrapper" class="ad advertisement ad-h">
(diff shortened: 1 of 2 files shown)


fehmer:
I'd prefer changing the delimiter.

Google docs and Excel do not support `|` as delimiter. LibreOffice does support `|`.  If we would use `;` as delimiter it would work in google docs, excel and libreOffice. 

LibreOffice and Google Docs don't support `dsv`.

vjgtigers:
What if we change the main delimiter to be a "," and change the delimiter for "item.charStats" and "item.tags" to be something like "|" or ";". This way we keep the csv file extension and should be readable by most csv readers.

Miodec:
Is this PR still needed if we change the delimiter to be a comma?

Miodec:
Closing as mentioned by the author in a different pr

PR #5269, opened 2024-04-01
impr: added percentage until next level in account page (vjgtigers)

vjgtigers:
### Description

added percentage to next level in the accounts page


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/8503174016

Miodec:
I think instead of that we could add a "xx %" when hovering over the actual bar, instead of the xp text.

PR #5298, opened 2024-04-10
impr(backend): report feedback to user about accepted and rejected reports (sanidhyas3s)

### LINKED ISSUE #3909: Send feedback to user about their report

sanidhyas3s:
### Description
Add two admin endpoints which handle accepted and rejected reports and send mail to user accordingly.

### Checks

- [ ] Adding quotes?

Closes #3909

### DIFF
--- a/backend/src/api/controllers/admin.ts
+++ b/backend/src/api/controllers/admin.ts
@@ -54,10 +54,20 @@ export async function handleReports(
           404
         );
       }
-      const mailBody =
-        `Your report for the ${report.type}, ${report.contentId} (for ${report.reason}) was ` +
-        (accept ? `approved.` : `denied.`) +
-        (reason !== undefined ? `\nReason: ${reason}` : "");
+
+      let mailBody = "";
+      if (accept) {
+        mailBody = `Your report regarding ${report.type} ${
+          report.contentId
+        } (${report.reason.toLowerCase()}) has been approved. Thank you.`;
+      } else {
+        mailBody = `Sorry, but your report regarding ${report.type} ${
+          report.contentId
+        } (${report.reason.toLowerCase()}) has been denied. ${
+          reason !== undefined ? `\nReason: ${reason}` : ""
+        }`;
+      }
+
       const mailSubject = accept ? "Report approved" : "Report denied";
       const mail = buildMonkeyMail({
         subject: mailSubject,


backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,98 @@
...
+import * as db from "../../init/db";
+import MonkeyError from "../../utils/error";
+import { buildMonkeyMail } from "../../utils/monkey-mail";
+import * as UserDAL from "../../dal/user";
 
 export async function test(): Promise<MonkeyResponse> {
   return new MonkeyResponse("OK");
 }
+
+const COLLECTION_NAME = "reports";
```
Miodec, line 12: I wouldnt store this like that since we will have more functions here which might use a different collection name.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,98 @@
...
+}
+
+export async function acceptReports(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { reports } = req.body;
+
+  try {
+    let acceptedReports = 0;
+    console.log("code added");
```
Miodec, line 30: console log

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,98 @@
...
+    for (const { reportId, verdict } of reports) {
+      if (verdict === true) {
```
Miodec, line 32: We dont need a verdict boolean if we have a separate accept and deny endpoints.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,98 @@
...
+
+export async function rejectReports(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { reports } = req.body;
+
+  try {
+    let rejectedReports = 0;
+    for (const { reportId, verdict } of reports) {
+      if (verdict === false) {
```
Miodec, line 70: Same here, we don't need a verdict.

Miodec (changes requested):

backend/src/api/controllers/admin.ts
sanidhyas3s: so we just send accepted reports to the accepted endpoint and rejected to the rejected. will make the changes accordingly.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+    throw new MonkeyError(404, "Report not found.");
+  }
+}
+
+export async function acceptReports(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { reports } = req.body;
+  try {
+    for (const { reportId } of reports) {
```
fehmer, line 28: While performance does not matter much on this endpoint i'd still reduce the amount of calls to the db and fetch all reports in one call.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+    return new MonkeyResponse(e.message, e.status);
+  }
+}
+
+export async function rejectReports(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { reports } = req.body;
+  try {
+    for (const { reportId, reason } of reports) {
```
fehmer, line 59: a lot of duplicated code between the two functions, i'd add a third method with parameters `accepted:boolean, reportIds: number[]`

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { reports } = req.body;
+  try {
+    for (const { reportId } of reports) {
+      const query = { _id: new ObjectId(reportId) };
+      const report = await db
+        .collection<MonkeyTypes.Report>(REPORTS_COLLECTION)
+        .findOne(query);
+      if (!report) {
```
fehmer, line 33: not sure about the error handling here. If the request has three Ids with the first and the last one existing and the second one non existing you'll get a 404 response but the first Id will be handled. The caller has no indication why the request failed (error message doesn't contain the reportId which is non existing).

We could either ignore missing reportIds or should fail the call without modifing any report if there is one non-existing reportId in the request.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+import * as db from "../../init/db";
+import MonkeyError from "../../utils/error";
+import { buildMonkeyMail } from "../../utils/monkey-mail";
+import * as UserDAL from "../../dal/user";
 
 export async function test(): Promise<MonkeyResponse> {
   return new MonkeyResponse("OK");
 }
+
+const REPORTS_COLLECTION = "reports";
```
fehmer, line 12: should we follow the structure of the other endpoints and move the db related code to the `dal` layer?

backend/src/api/controllers/admin.ts
sanidhyas3s: Yeah, I see the problem, will fix it. Making the entire request atomic in way, going through with all or failing all.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+async function removeReportFromDb(reportId: string): Promise<void> {
```
fehmer, line 13: should we accept an array of ids?

backend/src/api/controllers/admin.ts
sanidhyas3s: was thinking about that too, I think we can, I'll do it

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+  const query = { _id: new ObjectId(reportId) };
+  const result = await db
+    .collection<MonkeyTypes.Report>(REPORTS_COLLECTION)
+    .deleteOne(query);
+  if (result.deletedCount === 0) {
```
fehmer, line 18: we send out the email before calling this function so the user will get a successful email while the report was not deleted.

backend/src/api/controllers/admin.ts
```diff
@@ -1,5 +1,85 @@
...
+  }
+}
+
+export async function acceptReports(
+  req: MonkeyTypes.Request
+): Promise<MonkeyResponse> {
+  const { reports } = req.body;
+  try {
+    for (const { reportId } of reports) {
+      const query = { _id: new ObjectId(reportId) };
```
fehmer, line 29: do we query by `report.uid` or `report._id` @Miodec ?

backend/src/api/controllers/admin.ts
```diff
@@ -1,85 +1,73 @@
...
+    return new MonkeyResponse(
+      `Reports not found for some IDs`,
+      missingReportIds,
+      404
+    );
+  }
+
+  await ReportDAL.deleteReports(reportIds);
+
+  // sorting both the arrays ensures one to one mapping as no reports are missing
```
fehmer, line 45: unusual approach. What about storing the reportsFromDb in a map?
`const reportById = new Map( reportsFromDb.map((it) => [it.id, it]));`

and in the loop `reportById.get(report.id);`

sanidhyas3s: Yeah, that seems better. Will change it.

fehmer (dismissed):
nice work

sanidhyas3s:
@Miodec can you merge this or anything left to do?

backend/src/api/routes/admin.ts
```diff
@@ -48,4 +48,48 @@ router.post(
...
+  }),
+  checkIfUserIsAdmin(),
+  validateRequest({
+    body: {
+      reports: joi
+        .array()
+        .items(
+          joi.object({
+            reportId: joi.string().required(),
+            reason: joi.string().optional(),
```
Miodec, line 64: Move reason to the reject endpoint

sanidhyas3s: I have kept it this way to keep the accept and reject endpoints consistent and also enabling better reuse of code. Anyway the reason parameter is optional and hence the accept endpoint can be used completely without it.

Miodec:
Thanks

PR #5352, opened 2024-05-02
refactor: live stats

Miodec:
Update live stats, allowing to mix and match styles of timer/progress, live speed, live acc, live burst.

 - Replace show live wpm with live speed style
 - Replace show live acc with live acc style
 - Replace show live burst with live burst style
 - Update html structure
 - Update styles
 - Move timer opacity dom modifying code to test ui


backend/src/api/schemas/config-schema.ts
```diff
@@ -53,6 +52,9 @@ const CONFIG_SCHEMA = joi.object({
   confidenceMode: joi.string().valid("off", "on", "max"),
   indicateTypos: joi.string().valid("off", "below", "replace"),
   timerStyle: joi.string().valid("off", "bar", "text", "mini"),
+  liveSpeedStyle: joi.string().valid("off", "text", "mini"),
```
fehmer, line 55: don't we need to migrate existing configs?

PR #5360, opened 2024-05-04
feat(language): adds portuguese without accent marks or cedilla

arthurdepina:
### Description

Adds portuguese without accented characters (such as "ã", "ê", "í", "õ", "ü", etc.) or cedilla ("ç").


Miodec:
No need to add a separate word list, thats what lazy mode is for. If you enable it, it will remove those characters on the fly.

arthurdepina:
Hey! While lazy mode removes those characters, it keeps accented words, just misspelled, (instead of "questão" we get "questao").
In my opinion, a separate word list would be nice to practice without getting used to misspelling accented words.
Not trying to argue about it though, if you say that lazy mode is enough, than that's fine!

PR #5369, opened 2024-05-06
impr: Build docker images for linux/arm64 (fehmer)

fehmer:
Add `linux/arm64` platform to the docker build. 

Fixes #5367


Miodec:
Do i need to do anything special after merging? Or will it just build on release

fehmer:
> Do i need to do anything special after merging? Or will it just build on release

It will build for both platforms on next release.

PR #5390, opened 2024-05-13
feat: add hi viz theme (fehmer)

fehmer:


maxvoltar:
Totally up to you, but the official Hi! Viz text color is #777777. Horrible for legibility, but fits the theme ¯\_(ツ)_/¯

fehmer:
I'm using `#777777` as `main color`. You can get the `#777777` in the typing test by setting `colorful mode` to `on` and optionally `flip test colors` to `on`

image
resulting in
image

PR #5413, opened 2024-05-17
fix: maintenance banner is blocked by ublock (fehmer)

fehmer:
PSA are hidden when using uBlock Origin by the filter rule `monkeytype.com##.banner`. 
Fixed by using different classes for regular banners and PSAs.


Miodec:
Why keep `banner` at all then if its going to get blocked? I would just fully change everything to `psa`

Miodec (approved):
Thanks

PR #5433, opened 2024-05-23
fix: too many decimals showing on profile page for some users (fehmer)

fehmer:
For some users too many decimal places are shown on the profile page:
image

The `user.xp` value for this user is not a `Long` as expected. Result from the `GET profile` call:
```json
{
  xp: 14852507.822024472
}
```
While wo do round the XP value to a full number in the [UserDal.incrementXp](https://github.com/monkeytypegame/monkeytype/blob/master/backend/src/dal/user.ts#L604) method this will not fix existing non-long values to be rounded in the database.

This PR fixes the `Numbers.abbreviateNumber` method not handling the `decimalPoints` parameter on values `<1000` but not fixing the inconsistencies in the database.

### DIFF
--- a/frontend/__tests__/utils/numbers.spec.ts
+++ b/frontend/__tests__/utils/numbers.spec.ts
@@ -1,0 +1,49 @@
+import * as Numbers from "../../src/ts/utils/numbers";
+
+describe("numbers", () => {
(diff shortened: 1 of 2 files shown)


Miodec (approved):
Thank you

PR #5444, opened 2024-05-25
docs: utilize blockquote color highlights (unrealapex)

unrealapex:
Use Github Flavored Markdown's blockquote color highlights to make notes and warnings more clear.

### DIFF
--- a/CONTRIBUTING_ADVANCED.md
+++ b/CONTRIBUTING_ADVANCED.md
@@ -25,7 +25,9 @@ This contribution guide is for cases in which you need to test the functionality
 
 ### Git
 
-**IMPORTANT: If you are on Windows, run `git config --global core.autocrlf false` before cloning this repo to prevent CRLF errors.**
+
+> [!WARNING]
+> If you are on Windows, run `git config --global core.autocrlf false` before cloning this repo to prevent CRLF errors.** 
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING_ADVANCED.md
```diff
@@ -25,7 +25,9 @@ This contribution guide is for cases in which you need to test the functionality
 
 ### Git
 
-**IMPORTANT: If you are on Windows, run `git config --global core.autocrlf false` before cloning this repo to prevent CRLF errors.**
+
+> [!WARNING]
+> If you are on Windows, run `git config --global core.autocrlf false` before cloning this repo to prevent CRLF errors.**
```
Warrobot10, line 30: **

Warrobot10, line 30:
```suggestion
> **If you are on Windows, run `git config --global core.autocrlf false` before cloning this repo to prevent CRLF errors.** 
```

SELF_HOSTING.md
```diff
@@ -196,4 +196,6 @@ Example content from `backend-configuration.json`:
 If you have the `curl` and `jq` installed you can also run `curl -wO- http://localhost:5005/configuration | jq ".data" > backend-configuration.json` to update the configuration file.
```
Warrobot10, line 196:
```suggestion

If you have the `curl` and `jq` installed you can also run `curl -wO- http://localhost:5005/configuration | jq ".data" > backend-configuration.json` to update the configuration file.```
While we are here, we could fix this typo aswell.

Warrobot10: We can fix this typo while we are here.

unrealapex: Are there any other grammatical changes that should be made in the documentation? If so, that could be a separate PR.

Warrobot10: Looking at the self hosting doc, there seems to be some sporadic use of lowercase at the start of some sections, particularly towards the start of the doc. I will not make the pr now, because I know that merging this will cause merge conflicts, so I’ll wait until this merges. Thanks!

Warrobot10 (approved):

PR #5447, opened 2024-05-26
fix(crt): shrinking xp bar and hints not glowing (NadAlaba)

NadAlaba:
### Description
1. Adding `filter` css property to the non-positioned `.levelAndBar` made it the containing block for the absolute positioned child `.xpBar`, so when crt is enabled the `width: 100%` of `.xpBar` is getting calculated relative to `.levelAndBar` instead of `nav .textButton.account`, which causes the shrinking effect of `.xpBar`.
**Before:**
xpBarBefore
**After:**
xpBarAfter
2. Hints of (indicate typos = below) had the css property `text-shadow: none`, which I am not sure where it is used, but it is disabling the glowing effect of crt mode.
The safest way to fix that is to override this property with a more specific one in `crt.css`.
**Before:**
hintsBefore
**After:**
hintsAfter


Miodec:
Thanks

PR #5448, opened 2024-05-26
fix(xp): hide decimal point if displayed xp is < 1000 (NadAlaba)

NadAlaba:
### Description
- Currently a displayed level xp of `10` is shown as `10.0`.
This PR hides the decimal point if xp is < 1000, but values larger than 1k still show 1 decimal after abbreviating the number (i.e, 4000 appears as `4.0k`.
**Before:**
xpDecimalsBefore
**After:**
xpDecimalsAfter


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/9245240285

monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/9245296982

frontend/__tests__/utils/numbers.spec.ts
```diff
@@ -22,8 +22,8 @@ describe("numbers", () => {
     });
 
     it("should round to two decimals", () => {
-      expect(Numbers.abbreviateNumber(1, 2)).toEqual("1.00");
-      expect(Numbers.abbreviateNumber(1.5, 2)).toEqual("1.50");
+      expect(Numbers.abbreviateNumber(1, 2)).toEqual("1");
+      expect(Numbers.abbreviateNumber(1.5, 2)).toEqual("1.5");
```
Miodec, line 26: I kinda think this should remain `1.50`

Miodec:
Looking at this overall, I think instead of modifying the function behavior for anything less than 1000 (making it inconsistent), we should instead call a different function (or just manually round the number when its less than 1000). I would prefer consistency in that "if i call a function and ask for 2 decimal points, i always get 2 decimal points"

Miodec:
Thanks

PR #5453, opened 2024-05-28
Fixes #5440

### LINKED ISSUE #5440: Switch to Zod for schema validation, pass types to the endpoint handlers.
Currently, we use JOI for endpoint schema validation which does not support Typescript. Because of this, the controllers don't know whats inside query or body. We should switch to ZOD and then provide the types to the async handlers.

ManasKhudale:
Fixes #5440


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/9276283697

PR #5460, opened 2024-05-30
feat: detect custom background image not available (fehmer)

### LINKED ISSUE #5459: Try to detect if the custom background image link is expired
If it is clear the setting and notify the user.

fehmer:
Fixes #5459


Miodec (approved):
Thanks

PR #5466, opened 2024-05-31
fix: fix backspace to previous word in stop on letter (NadAlaba)

NadAlaba:
### Description

- In `stop on error > letter`, and after having typed an incorrect first letter of some word, then backspacing to the previous word (which is possible either in `Freedom mode` or if the previous word was incorrect), the backspaced incorrect letter (or word in `Highlight mode > word`) will stay highlighted as an error.

https://github.com/monkeytypegame/monkeytype/assets/37968805/ee8c9156-8696-4eb8-a57c-133518887ac5

- This PR updates the word element before backspacing when (`stop on error = letter` && incorrect first letter typed).


frontend/src/ts/controllers/input-controller.ts
```diff
@@ -135,19 +134,27 @@ function backspaceToPrevious(): void {
...
     setWordsInput(" " + TestInput.input.current + " ");
+    void TestUI.updateWordElement();
   }
-  TestWords.words.decreaseCurrentIndex();
-  TestUI.setCurrentWordElementIndex(TestUI.currentWordElementIndex - 1);
-  TestUI.updateActiveElement(true);
-  Funbox.toggleScript(TestWords.words.getCurrent());
-  void TestUI.updateWordElement();
 
   if (Config.mode === "zen") {
```
Miodec, line 159: Why is this reordered? What is this fixing?

Miodec: I feel like not calling `updateWordElement` in all cases could introduce some issues in the future? I would just call it everytime I think.

NadAlaba: Done! Assuming we're talking about the after backspacing call only.

Miodec:
Thanks

PR #5468, opened 2024-06-01
fix: test activity graph bugs (fehmer)

fehmer:
- fix graph showing 13 months instead of 12
- graph not showing values older than 12month to fill the first week
image


Miodec:
Thanks

PR #5480, opened 2024-06-09
impr(layout): add alt layer character to azerty (theiereman)

theiereman:
### Description

Some accented characters require the use of dead keys ("^" ; "¨" on AZERTY keyboards for example) and I noticed the layout wasn't highlighting the dead key when their use was necessary.

- added highlight of the required dead key press for the characters having "^" ; "¨" and "~" accents
- if accented character is already present on the layout it will highlight that one and not the dead key
- added two-step highlight once a dead key was pressed to highlight the next character required to finish the accented character (exemple : "ê" will highlight "^" key and once pressed will highlight "e" key)
- if no key was mapped on layout to create the required accent, behavior will be the same as before = no highlight

Exemple for accented character already present on layout (BEPO) : 
firefox_5xenPxc9Cv

Exemple for 2-steps accented character highlight (AZERTY) : 
firefox_uiB7P97Ua0

Known issue : 
- I found no way to differentiate the dead key presses while still having a layout agnostic behavior so any dead key press will trigger the "2-step highlight" even if the required dead key character is wrong (exemple for "ê" : if you press "~", "e" will still get highlighted even if the finished character will generate an error


frontend/src/ts/elements/keymap.ts
```diff
@@ -22,6 +22,12 @@ const stenoKeys: MonkeyTypes.Layout = {
   },
 };
 
+const accentMap = {
+  "^": ["ê", "â", "î", "ô", "û"],
+  "¨": ["ë", "ä", "ï", "ö", "ü", "ÿ"],
+  "~": ["ã", "õ", "ñ"],
+};
+
```
theiereman, line 30: didn't really know where to put this

frontend/static/layouts/_list.json
```diff
@@ -419,10 +419,10 @@
     }
   },
   "azerty": {
-    "keymapShowTopRow": false,
+    "keymapShowTopRow": true,
```
theiereman, line 422: top row is required to show accented characters like "é" and "à"

frontend/src/ts/utils/strings.ts
```diff
@@ -149,3 +149,23 @@ export function cleanTypographySymbols(textToClean: string): string {
...
+export function unaccentedCharacter(character: string): string {
+  const charMap = {
+    e: ["ê", "ë"],
+    a: ["â", "ã", "ä"],
+    o: ["ô", "ö", "õ"],
+    i: ["î", "ï", "ĩ"],
+    u: ["û", "ü"],
+    n: ["ñ"],
+    y: ["ÿ"],
+  };
```
theiereman, line 162: same for that map, didn't really know where to put it

frontend/static/layouts/_list.json
```diff
@@ -419,10 +419,10 @@
...
     "type": "iso",
     "keys": {
-      "row1": ["`~", "&1", "é2", "\"3", "'4", "(5", "-6", "è7", "_8", "ç9", "à0", ")°", "=+"],
+      "row1": ["`", "&1", "é2~", "\"3", "'4", "(5", "-6", "è7", "_8", "ç9", "à0", ")°", "=+"],
```
Miodec, line 425: Whats the purpose of this?

frontend/src/ts/utils/strings.ts
Miodec: We already have `lazy-mode.ts` which includes a big list of accented and de-accented characters.

frontend/src/ts/elements/keymap.ts
```diff
@@ -32,15 +38,27 @@ function highlightKey(currentKey: string): void {
...
+          if (accentedChars.includes(currentKey)) {
+            highlightKey = `#keymap .keymapKey[data-key*="${accent}"]`;
+            break;
+          }
+        }
+      }
     }
 
-    // console.log("highlighting", highlightKey);
+    console.log("highlighting", highlightKey);
```
Miodec, line 61: Im not sure i like this approach, why not modify the function call instead of doing this hacky change?

frontend/src/ts/elements/keymap.ts
Miodec: This should probably be defined per language, somewhere in the constants directory.

Miodec (changes requested):

frontend/static/layouts/_list.json
theiereman: tilde is on Digit2 for AZERTY keyboards

image

Miodec: Im assuming the blue is the alt layer? Can you add the rest of the characters?

frontend/src/ts/elements/keymap.ts
theiereman: I agree that it is a hacky change but I'm not sure I understand what you mean, sorry. I thought this was the best location to implement that behavior because it was central

frontend/static/layouts/_list.json
```diff
@@ -422,8 +422,8 @@
     "keymapShowTopRow": true,
     "type": "iso",
     "keys": {
-      "row1": ["`", "&1", "é2~", "\"3", "'4", "(5", "-6", "è7", "_8", "ç9", "à0", ")°", "=+"],
-      "row2": ["aA", "zZ", "eE", "rR", "tT", "yY", "uU", "iI", "oO", "pP", "^¨", "$£"],
+      "row1": ["`", "&1", "é2~", "\"3#", "'4{", "(5[", "-6|", "è7`", "_8\\", "ç9", "à0@", ")°]", "=+}"],
```
theiereman, line 425: I didn't add the "^" on Digit9 because would conflict with "^¨" on `row2` and to be fair, no one uses that to write French. The only use I can imagine is math power symbol.

Exemple for conflict if added on Digit9 :
image

frontend/src/ts/elements/keymap.ts
Miodec: I mean instead of calling `highlightKey` with `á`, call it with `´` and then `a`.

Miodec:
Initial testing:

 - the highlighting behavior happens even when keymap mode is NOT set to `next`
 - if the compose character is the last one, keymap suggest space instead of `a`
image

theiereman:
@Miodec I can't even reproduce the bugs of your initial testing, it's so weird

Tbh I think that improvement is not really worth the effort and what I did feels overall really clunky
The fact that every special character is not shown on the layout (so can't really be highlighted) and that "dead key detection" is global to every dead key, makes it a very minor (if any) improvement. 

I'm ok with you closing that PR if you're ok with it.

PS : maybe you should just keep the modification I did on AZERTY layout

Miodec:
> @Miodec I can't even reproduce the bugs of your initial testing, it's so weird
> 
> Tbh I think that improvement is not really worth the effort and what I did feels overall really clunky The fact that every special character is not shown on the layout (so can't really be highlighted) and that "dead key detection" is global to every dead key, makes it a very minor (if any) improvement.
> 
> I'm ok with you closing that PR if you're ok with it.
> 
> PS : maybe you should just keep the modification I did on AZERTY layout

Yeah, keep the layout modifications and revert everything else (or open a new pr)

Miodec (approved):
Thanks

PR #5482, opened 2024-06-10
impr: added character-counter component to show how many characters are left in bio and keyboard fields on profile-edit page (theiereman)

theiereman:
### Description

I noticed a backend error when submitting a keyboard description that was too long (75 max), so I thought it would be nice to prevent backend errors and provide some kind of feedback about the number of characters left before submitting a value.

firefox_KG2Mu1W79i

### DIFF
--- a/frontend/src/ts/elements/character-counter.ts
+++ b/frontend/src/ts/elements/character-counter.ts
@@ -38,9 +38,9 @@ export class CharacterCounter {
 
     const remainingPercentage = (remaining / this.maxLength) * 100;
     if (remainingPercentage < 10) {
-      this.counterElement.addClass("red");
+      this.counterElement.addClass("warning");
     } else {
-      this.counterElement.removeClass("red");
+      this.counterElement.removeClass("warning");
     }
   }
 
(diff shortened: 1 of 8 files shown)


Miodec:
We already have a character counter on the report modal - can you make it look the same? 
image

You could update it to `x/(character limit)` too on the report and profile update screens

frontend/src/ts/elements/character-counter.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  private updateCounter(): void {
+    const maxLength = this.maxLength;
+    const currentLength = (this.textareaElement.val() as string).length;
+    const remaining = maxLength - currentLength;
+    this.counterElement.text(`${remaining.toString()} characters remaining`);
+
+    const remainingPercentage = (remaining / this.maxLength) * 100;
+    this.counterElement.css("color", remainingPercentage < 10 ? "red" : "");
```
Miodec, line 40: This should use theme error color instead of red.

Miodec:
Alternatively you could update the report modal to use your new system.

theiereman:
I changed every character counter I found (user report ; quote report ; quote submit) to the new system and adjusted the style of the component to fit the old one (except the "x/max length" part).

Miodec (approved):
Thanks

PR #5487, opened 2024-06-11
impr(custom theme rename modal): automatically select input (@fehmer)

fehmer:
Select the name text input on the `Update custom theme` modal so the user can directly type in the new name without deleting the old one


fehmer:
Make this option available on simple modals: https://github.com/monkeytypegame/monkeytype/blob/120b660efc0f6e22f133c96fe2a8975caf17aafd/frontend/src/ts/utils/animated-modal.ts#L176

Miodec (approved):
Thanks

PR #5488, opened 2024-06-11
impr: add button to remove custom theme background (@fehmer)

fehmer:
Add a button to remove the theme background instead of having the user to remove the text and hit save.


Miodec:
Thanks

PR #5496, opened 2024-06-16
fix: typed letters not visible in tts funbox if highlight=off (@fehmer)

fehmer:
I am not sure this is on purpose or not. If you choose highlight=off in tts funbox you cannot see what you typed.


Miodec:
Thanks

PR #5396, opened 2024-05-15
impr(dev): add endpoint to create test user/data (fehmer)

fehmer:
Allow easy creation of test users and test results on local development environment.

Adds a simple UI only available on dev to create test results and optionally also create test users.
Test users will use  `<username>@example.com` as email and the password is always `password`. You can choose the timeframe in which results are created and  `min/max` amount of results per day. The result count, test type, accuracy, wpm etc will be randomized.

PR #5499, opened 2024-06-17
feat: allow removal of password authentication (@fehmer)

fehmer:
Allow users to remove the password/email authentication if they have the account linked with google or github.


frontend/src/ts/modals/simple-modals.ts
```diff
@@ -673,6 +675,58 @@ list.removeGithubAuth = new SimpleModal({
   },
 });
 
+list.removePasswordAuth = new SimpleModal({
+  id: "removePaswordAuth",
+  type: "text",
+  title: "Remove Password authentication",
+  inputs: [
+    {
+      type: "checkbox",
```
fehmer, line 684: currently checkboxed are checked by default, the refactoring in https://github.com/monkeytypegame/monkeytype/pull/5396 will fix that

PR #5506, opened 2024-06-18
impr(quotes): add code_php quotes (aescanes)

aescanes:
### Description

Add code_php quotes. 

I used ChatGPT to generate examples of PHP 8.3 code. I created 50 quotes with text between 200 and 500 characters using "\n" as an escape character for a new line and "\t" as an escape character for a tab.

### DIFF
--- a/frontend/static/quotes/code_php.json
+++ b/frontend/static/quotes/code_php.json
@@ -1,0 +1,311 @@
+{
+  "language": "code_php",
+  "groups": [
+    [0, 100],
+    [101, 300],
+    [301, 600],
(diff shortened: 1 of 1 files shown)


Miodec:
Most of these duplicate the class names or have 2 classes inside, not separated by a new line
image

This syntax just seems incorrect

image


This code just doesnt make sense 

image

This just has comments in the function body

image


Im not sure if chat gpt is a good source.

aescanes:
Thank you for your time @Miodec. I'll change the source and create this again.

PR #5509, opened 2024-06-19
impr(quotes): add code_php quotes (aescanes)

aescanes:
### Description

Add code_php quotes.

I used ChatGPT and other sources to generate examples of PHP 8.3 code. I created 20 quotes with text between 150 and 650 characters using "\n" as an escape character for a new line and "\t" as an escape character for a tab.

@Miodec, this time I've checked more carefully the code from different sources. I hope everything looks fine now. Thanks


FFUV:
I would recommend not to use ChatGPT for programming quotes because of difficulties in the past.

Miodec:
Thanks

PR #5515, opened 2024-06-20
fix(zipf): improve approximation of zipf distribution (@NadAlaba)

NadAlaba:
### Description

Previous zipf index calculation was based on an approximation of the harmonic number that approaches the actual harmonic number near infinity, but does a very bad job at approximating the harmonic number of small values.

This caused the frequency of the first index to be 57% of the actual frequency, and the 2nd index became more frequent than the first one, which caused the word "be" to appear more than the word "the", when it should appear half as much as "the".

New approximation of the harmonic number is slightly better for large values and very much better for small values.

zipf


Miodec:
Thanks

DreymaR:
Nice! Obviously, I was not aware of this better approximation. Thanks!

PR #5517, opened 2024-06-22
feat(funbox): add pig latin funbox (spurkus)

Spurkus:
# Pig Latin Rules

**For words that begin with a vowel**
- Add yay to the end of the word

**For words that begin with a consonant**
- Move the first consonant to the end of the word
- Add ay to the end of the word

### Description

Added pig latin funbox.
Changed:
`backend\src\constants\funbox-list.ts`
`frontend\src\ts\test\funbox\funbox-list.ts`
`frontend\src\ts\test\funbox\funbox.ts`
`frontend\static\funbox\_list.json`

### DIFF
--- a/backend/src/constants/funbox-list.ts
+++ b/backend/src/constants/funbox-list.ts
@@ -344,6 +344,21 @@ const FunboxList: MonkeyTypes.FunboxMetadata[] = [
     frontendFunctions: ["alterText"],
     name: "ddoouubblleedd",
   },
+  {
+    name: "pig_latin",
+    difficultyLevel: 1,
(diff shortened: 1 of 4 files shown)


FFUV:
Thanks for the Pig Latin Funbox feature. However, I don't recommend merging it. Traditional Pig Latin has established rules beyond simple suffix additions, and creating a variant may be disrespectful. Additionally, there's already an existing Pig Latin language.

PIg Latin

Miodec:
Yeah im not sure whats the point of this when a pig latin language is already available?

PR #5519, opened 2024-06-23
impr(practice): random shuffle words in all practice modes (grzebiel)

grzebiel:
### Description

Repeating the same word multiple times in a row is far from realistic practice. This PR changes the order of words in practice (missed, slow, both) to be randomized.

### Checks

- [ ] Adding quotes?
    - Also, please add a screenshot of the theme, it would be extra awesome if you do so!


grzebiel:
> I would prefer using `CustomText.setMode("shuffle")` which will achieve the same result, just cleaner IMO.

I wasn't aware of it, updated. Thanks!

Miodec (approved):
Thanks

PR #5524, opened 2024-06-25
chore(husky): run `nvm use` only when needed in pre-push hook (@NadAlaba)

NadAlaba:
### Description

`nvm use` does not work on nvm-windows, and there is no need for it to run every time anyone pushes to any branch.
pre-push-hook


Miodec (approved):
Thanks

Issue #4857, opened 2023-12-11
HTML tags aren't entirely semantic


borisnezlobin:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

Yes

### Account name

randomletters

### Current Behavior

The HTML tags throughout the frontend are lacking in semantic structure (i.e., h1 for titles, h2 for subtitles, ul/ol for lists, etc)

### Expected Behavior

Currently, the layout looks something like:
``` 
 <div class="section">
    <div class="bigtitle">about</div>
    <h2>
      Monkeytype is a minimalistic and customizable typing test. It features
      many test modes, an account system to save your typing speed history, and
      user-configurable features such as themes, sounds, a smooth caret, and
      more. Monkeytype attempts to emulate the experience of natural keyboard
      typing during a typing test, by unobtrusively presenting the text prompts
      and displaying typed characters in-place, providing straightforward,
      real-time feedback on typos, speed, and accuracy.
      <br />
      <br />
      Test yourself in various modes, track your progress and improve your
      speed.
    </h2>
  </div>
```
Using semantics, it would look like:
```
  <section class="section">
    <h1 class="bigtitle">about</h1>
    <p>
      Monkeytype is a minimalistic and customizable typing test. It features
      many test modes, an account system to save your typing speed history, and
      user-configurable features such as themes, sounds, a smooth caret, and
      more. Monkeytype attempts to emulate the experience of natural keyboard
      typing during a typing test, by unobtrusively presenting the text prompts
      and displaying typed characters in-place, providing straightforward,
      real-time feedback on typos, speed, and accuracy.
      <br />
      <br />
      Test yourself in various modes, track your progress and improve your
      speed.
    </p>
  </section>
```
This would improve SEO and make things more clear for developers

### Steps To Reproduce

Go to any .html file in the frontend

### Environment

- OS: all
- Browser: all
- Browser Version: all

### Anything else?

https://www.w3schools.com/html/html5_semantic_elements.asp

Can I work on this? It's rather tedious, but worth doing

Miodec:
Known issue. Ive been slowly converting everything recently.

borisnezlobin:
if you want, I could help you out

[...]

PR #5528, opened 2024-06-25
impr(test config): allow keyboard navigation (@Truiteseche)

Truiteseche:
### Description
Improve keyboard navigation accessibility on the test-config section, improve semantic.

Maybe the fact that we were not able to focus textButtons with the keyboard was on purpose but I think it's more relevant to allow it, especially since in mobile layout it is focusable. I also sometimes accidentally log out because the log out button is just before the text field in the tabindex hierarchy, so it would resolve that small but annoying issue.

impr(test-config section): impossible to navigate with the keyboard in the test-config section (@Truiteseche)

Related issues (doesn't fix it at all):
#4857

### DIFF
--- a/frontend/src/html/pages/test.html
+++ b/frontend/src/html/pages/test.html
@@ -2,75 +2,75 @@
   <div id="testConfig" class="full-width">
     <div class="row">
       <div class="puncAndNum">
-        <div class="textButton punctuationMode">
+        <button class="textButton punctuationMode">
           <i class="fas fa-fw fa-at"></i>
           punctuation
-        </div>
-        <div class="textButton numbersMode">
+        </button>
+        <button class="textButton numbersMode">
           <i class="fas fa-fw fa-hashtag"></i>
           numbers
-        </div>
+        </button>
       </div>
 
       <div class="spacer leftSpacer"></div>
 
       <div class="mode">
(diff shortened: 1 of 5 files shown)


Miodec:
Focused elements should use the outline to indicate focus. I know it would require a lot more changes than this (you would need to remove variable padding on the button to avoid outline misalignment) - do you want me to do this or do you wanna give it a try?

Truiteseche:
Thank you for your feedback! Indeed I kicked the outline in favor of the same effect as on hover, because the outline was glitched, but yeah it would be better with the outline. I'd like to give it a try even though I don't fully understand what you mean removing **variable** padding** .

Miodec:
> Thank you for your feedback! Indeed I kicked the outline in favor of the same effect as on hover, because the outline was glitched, but yeah it would be better with the outline. I'd like to give it a try even though I don't fully understand what you mean removing **variable padding**.

By variable padding i mean this 
image
image

Truiteseche:
Thank you very much! I put the outline back in place, as you advised me I removed the variable padding (in favor of the variable margin) to avoid the outline misalignment, and I removed `overflow: hidden` on some containers to avoid the outline display problems On my side, it's ready to merge.

Miodec:
Nicely done, thanks.

PR #5533, opened 2024-06-26
fix(test): apply highlight mode in loading and test (mitjans)

### LINKED ISSUE #5507: Some highlight modes not working on initial page load

mitjans:
### Description

Apply highlight mode in both `"test"` and `"loading"` pages.
On initial page load, the current page is set to `"loading"`, so no highlight mode is applied.
When in `"loading"` page, the active element is yet to be defined, so we should check if we are actually in `"test"` page before calling `updateActiveElement();`

### Checks

- [ ] Adding quotes?

Closes #5507

### DIFF
--- a/frontend/src/ts/test/test-ui.ts
+++ b/frontend/src/ts/test/test-ui.ts
@@ -155,9 +155,12 @@ ConfigEvent.subscribe((eventKey, eventValue, nosave) => {
   if (eventKey === "theme") void applyBurstHeatmap();
 
   if (eventValue === undefined) return;
-  if (eventKey === "highlightMode" && ActivePage.get() === "test") {
+  if (
+    eventKey === "highlightMode" &&
+    (ActivePage.get() === "test" || ActivePage.get() === "loading")
+  ) {
     highlightMode(eventValue as SharedTypes.Config.HighlightMode);
-    updateActiveElement();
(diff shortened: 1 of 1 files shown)


fehmer:
Hi @mitjans , sorry I only found out that the bug I was working on was related to this. My fix #5537 fixes both issues.

mitjans:
> Hi @mitjans , sorry I only found out that the bug I was working on was related to this. My fix #5537 fixes both issues.

No problem @fehmer !

Do you want me to close this PR? I was wondering if we can continue with this one since this is my first contribution to Monkeytype 😄 

Also, I see you have removed the check `&& ActivePage.get() === "test"` entirely. I thought about it too, but Miodec added that check 2 weeks ago so I thought it was still relevant.

fehmer:
@mitjans  let's wait for mio to review.

I moved the `ActivePage.get() === "test"` condition to `updateActiveElement()`. This way the highlight class is getting set on `#words`  when the config changes/is applied but the active is only updated if we are on the test page.

mitjans:
I see, so the original check added by Miodec wanted to avoid the `activeElement` being undefined when loading.
In that case, it makes sense to only move the check. I'll update this PR as well to reflect this, thanks! :)

Miodec:
Thanks guys.

PR #5534, opened 2024-06-26
fix: lazy mode not working for vietnamese (@fehmer)

fehmer:
Add missing accents for vietnamese language.

Code used to extract accents: 
```js
const letters = new Set();
JSON.parse(readFileSync("./static/languages/vietnamese_5k.json"))
.words.map((it) => it.replace(/[a-zA-Z]/g, ""))
.filter((it) => it !== "")
.flatMap((it) => it.split(""))
.forEach((it) => letters.add(it));

const replacements = new Array(...letters).sort((a, b) =>
a.localeCompare(b)
);
console.log(replacements.length, JSON.stringify(replacements));
```


Miodec (approved):
Thanks

PR #5536, opened 2024-06-27
fix: fix mexican theme being unreadable (@fehmer)

fehmer:
There are some issues with the main and sub-alt color being the same. The PR is changing the sub-alt color to light blue. This changes the UI a lot (test settings, settings page, account page) but keeps the test colors the same, which I think is more important.

## Test config
- original 
image
- fixed 
image

## Account page
- original 
image
- fixed 
image

## Leaderboard
- original 
image
- fixed 
image


Miodec:
Thanks

PR #5549, opened 2024-07-01
chore: fix check annotations for dev routes (@fehmer)

fehmer:
Fix lint warnings during ci-be build:

- `backend/src/api/controllers/dev.ts:272`: 'it' is defined but never used. 
- `backend/src/api/routes/dev.ts:2`: 'authenticateRequest' is defined but never used.


Miodec (approved):
Thanks

PR #5550, opened 2024-07-01
fix: test history tags button flashing on load (@fehmer)

fehmer:
Tags button is flashing on load if the current active tags are matching the tags from the result:
image

Introduced with e29d280.


Miodec (approved):

Miodec:
Thanks

PR #5552, opened 2024-07-02
fix: mexican theme sub alt color (@fehmer)

fehmer:
Yesterday I learned that the sub-alt color should be a slightly brighter or slightly darker color of the background.


Miodec (approved):
Thanks

PR #5558, opened 2024-07-04
fix: optimize / fix concurrency in UserDal (@fehmer)

fehmer:
Fix possible concurrency problems by using atomic database calls instead of read-modify-write where possible:

- `addResultFilterPreset` do max filters check on the database
- `removeResultFilterPreset` do filter exists check on the database
- `addTag` do max tags check on the database
-  `editTag`, `removeTag`, `removeTagPb`  do tag exists check on the database
-  `updateLbMemory` set only a single field, not the whole `user.lbMemory`
- `incrementBananas` use condition to find `pb>=maxPb*0.75` and increment on the database
-  `addTheme` do max custom themes check on the database
- `removeTheme`, `editTheme` do theme exists check on the database  
- `addFavoriteQuote` move max favorite quotes check to database
- `removeFavoriteQuote` remove check for existing quote,  no error was thrown anyways

### other fixes
-  `isDiscordIdAvailable`  only fetch  `_id` to improve performance
- `getStats` fix method signature `
- `updateInbox` refactor to not have function as a string


monkeytypegeorge:
Continuous integration check(s) failed. Please review the failing check's logs and make the necessary changes. https://github.com/monkeytypegame/monkeytype/actions/runs/9814296390

Miodec (approved):
Thanks

PR #5559, opened 2024-07-04
fix: cleanup tmp while updating users inbox (@fehmer)

fehmer:
fix mongo update pipeline leaving behind tmp data


Miodec (approved):
Thanks

PR #5560, opened 2024-07-04
fix: updateInbox claiming all rewards (@fehmer)

fehmer:
- fix non-read rewards being claimed
- claim rewards on deletion
- fix claiming badges can result in duplicate badges in users inventory


Miodec (approved):
Thanks

PR #5562, opened 2024-07-04
fix: incorrect word highlight in some themes (@fehmer)

### LINKED ISSUE #5561: Highlight not working in rainbow trail theme
> Sorry to insist on this, but it looks like the theme "rainbow trail" has the same problem where the word disappears before being typed. I feel like adding a theme check is not enough because themes will depend on the "correct" HTML class. 

https://github.com/monkeytypegame/monkeytype/assets/7190600/dee3e3db-d545-483f-a77a-af87280f5664

> I can try to think of a solution that doesn't imply creating a new setting, but I'm not sure I understand the expected behavior on each of the highlighting modes.

_Originally posted by @mitjans in https://github.com/monkeytypegame/monkeytype/issues/5557#issuecomment-2209248083_

fehmer:
Fix word highlight not working in these themes:
- rainbow trail
- aurora
- fire
- grape
- rgb
- trance

Fixes #5561

### DIFF
--- a/frontend/static/themes/aurora.css
+++ b/frontend/static/themes/aurora.css
@@ -70,10 +70,21 @@ a:hover,
   animation: aurora 5s linear infinite;
 }
 
-#words.flipped .word letter.correct {
+#words.flipped .word:has(~ .active) letter,
+#words.flipped.highlight-next-word .word.active,
+#words.flipped.highlight-next-word .word.active + .word,
+#words.flipped.highlight-next-two-words .word.active,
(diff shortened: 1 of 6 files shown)


mitjans:
Thanks @fehmer! This fixes the problem for themes with special effects in correctly typed words.

Shouldn't the other themes have the same behavior? For example, I would expect correctly typed words to remain black for theme `9009` with the highlight mode "next word".

fehmer:
fixed by #5568.

PR #5595, opened 2024-07-09
fix: preview fonts not working correctly in the commandline 

Miodec:
Also fixes Comic Sans not working in the settings page.


fehmer (approved):
lgtm

PR #5597, opened 2024-07-09
fix: league of legends language being right to left (@fehmer)

fehmer:


Whydidthechickencrosstheroad:
It feels like it is not done yet

Whydidthechickencrosstheroad:
It's already done

PR #5606, opened 2024-07-11
fix: zen mode not working (@fehmer)

fehmer:
- fixes test input not visible. caused by a00a8bb94c2360e60647473cbb311e49c6ff69d7 
- fixes newline not working in zen mode, caused by aac44aead339cfd69c36e5a75a2e3945a44dd3d5

### DIFF
--- a/frontend/src/ts/test/caret.ts
+++ b/frontend/src/ts/test/caret.ts
@@ -76,7 +76,11 @@ export async function updatePosition(noAnim = false): Promise<void> {
     Config.caretStyle
   );
 
-  const wordLen = TestWords.words.getCurrent().length;
+  const wordLen = (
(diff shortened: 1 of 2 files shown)


frontend/src/ts/test/caret.ts
```diff
@@ -116,19 +119,19 @@ export async function updatePosition(noAnim = false): Promise<void> {
   );
 
   const letterPosTop =
-    currentLetter?.offsetTop ||
-    previousLetter?.offsetTop ||
+    currentLetter?.offsetTop ??
```
fehmer, line 122: offsetTop can be zero. using `||` will evaluate `0` as `false`

fehmer:
duplicate

PR #5607, opened 2024-07-11
fix: xp bar not updating when claiming rewards (@fehmer)

### LINKED ISSUE #5564: Claiming rewards on the account page does not update XP bar
## steps to reproduce
- have unread mails with rewards
- go to account page
- open inbox and claim all
- close notification panel

## Problem
xp bar in profile is not getting updated

fehmer:
fixes #5564


Miodec (approved):
Thanks

PR #5626, opened 2024-07-18
refactor: restructure project to a true monorepo

Miodec:
### Description

 - Rework project structure to follow a monorepo layout (- apps folder, files will be moved in a separate PR to avoid an even bigger diff)
   - use npm workspaces 
   - Add Turborepo to handle automatically running tasks on dependent packages
   - Update tsconfigs to support importing from local packages
     - update imports to add `.js` at the end
   - Remake all build, dev, test scripts to use turbo tasks
 - Move shared types to a package
   - due to this we are also slowly moving away from global namespaces in favor of importing types when needed
 - Move tsconfigs to a package
 - Move eslint config to a package
 - Update Node to 18.20.4 LTS
 - Change ts-node-dev to tsx for backend dev
 - ...


fehmer:
turrrbo!

docker/frontend/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
 COPY docker/frontend/firebase-config-live.ts frontend/src/ts/constants/firebase-config-live.ts
+COPY docker/frontend/.env frontend/.env
 
 #BUILD
 RUN npm ci
-RUN cd frontend && npm ci
-RUN cd frontend && npx vite build
-
+##TODO: remove ||true
+RUN npm run build || true
```
fehmer, line 22: remove after slim-select is fixed

fehmer (approved):

PR #5637, opened 2024-07-21
feat(practice): add option to practice missed biwords & convert practice popup to a modal (@sanidhyas3s)

### LINKED ISSUE #5316: Add option to practice missed biwords
meaning pairs of words, the word before the failed one and the failed one

sanidhyas3s:
### Description
Added an option to practice the missed word (along with the word before it). 
Converted the practice popup to a modal.
Tested, working.

### Checks

Closes #5316


frontend/src/html/popups.html
```diff
@@ -281,7 +281,8 @@
     </div>
     <div class="button missed" tabindex="1">Practice missed</div>
     <div class="button slow" tabindex="1">Practice slow</div>
-    <div class="button both" tabindex="1">Practice both</div>
+    <div class="button both" tabindex="1">Practice slow & missed</div>
+    <div class="button biwords" tabindex="1">Practice missed biwords</div>
   </div>
 </div>
```
Miodec, line 288: I just realised this way we cant practice missed biwords and slow words. Maybe we could remake this into a form? You select what you want to practice and click "practice". Checkboxes would include "missed words", "missed biwords", "slow words" (width the first 2 options being exclusive). That way the user can select any combination.

Miodec: Or maybe something similar to custom mode 
image

where missed will be a group of buttons ("missed", "missed biwords") and slow will be an "on" or "off"

frontend/src/ts/test/practise-words.ts
```diff
@@ -24,7 +24,11 @@ export const before: Before = {
   customText: null,
 };
 
-export function init(missed: boolean, slow: boolean): boolean {
+export function init(
+  missed: boolean,
+  slow: boolean,
+  biwords: boolean = false
+): boolean {
   if (Config.mode === "zen") return false;
```
Miodec, line 32: I think instead of adding a separate parameter i would change the type of missed to `"single" | "biword"`.

Miodec (changes requested):

frontend/src/html/popups.html
sanidhyas3s: > Or maybe something similar to custom mode 
> image
> 
> where missed will be a group of buttons ("missed", "missed biwords") and slow will be an "on" or "off"

This on the other hand seems counter intuitive and too much of a hassle.

Miodec: Just make sure to disable the "missed" checkbox if you enable "biword" checkbox and vice versa.

frontend/src/ts/test/practise-words.ts
sanidhyas3s: If we do it the "word before missed" way, then being separated would be better.

PR #5643, opened 2024-07-23
fix: bug in hexToRgb function (brenob6)

brenob6:
### Description

Added some test to the hexToRgb function. I found some issues and fixes it. 
To be able to test the hexToRgb function I exported it.

### DIFF
--- a/frontend/__tests__/utils/colors.spec.ts
+++ b/frontend/__tests__/utils/colors.spec.ts
@@ -1,0 +1,18 @@
+import { hexToRgb } from "../../src/ts/utils/colors";
+
+describe("colors.ts", () => {
+  describe("hexToRgb", () => {
+    it("Invalid hex values", () => {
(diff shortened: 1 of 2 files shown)


frontend/__tests__/utils/colors.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe("colors.ts", () => {
+  describe("hexToRgb", () => {
+    it("Invalid hex values", () => {
+      expect(hexToRgb("ffff")).toEqual(undefined);
+      expect(hexToRgb("fff0000")).toEqual(undefined);
+      expect(hexToRgb("#ff")).toEqual(undefined);
+    });
+    it("Valid hex value", () => {
+      expect(hexToRgb("#ff0000")).toEqual({
```
fehmer, line 11: suggest to use another value for testing like `#123456` to verify r,g,b is calculated correctly

Miodec:
Thanks

PR #5657, opened 2024-07-26
ci: optimize prime-cache checkout (@fehmer)

fehmer:
Only checkout the package-lock.json initially because it is used to fetch the cache.
If the cache is missing checkout all files and run the install to fill the cache.

Reduces the time of prime-cache in case the cache is up to date from ~10 seconds down to one second.


.github/workflows/monkey-ci.yml
```diff
@@ -68,7 +68,11 @@ jobs:
     needs: [pre-ci]
     if: needs.pre-ci.outputs.should-build-be == 'true' || needs.pre-ci.outputs.should-build-fe == 'true' || needs.pre-ci.outputs.should-build-pkg == 'true' || needs.pre-ci.outputs.assets-json == 'true'
     steps:
-      - uses: actions/checkout@v4
+      - name: checkout package-lock
```
Miodec, line 71: Capital C please ;D

PR #5658, opened 2024-07-26
impr: add Occitan lengadocian dialect (Mejans)

Mejans:
### Description

This PR adds lengadocian dialect, the central one and more used online, with 2 lists
It improve the first PR I made by allowing people the type words they actually use and exists in their dialect.


Miodec:
Whats the difference between the normal lists and lengadocian?

Mejans:
Sorry, I think I didn't generate my files like I wanted to =/
I'll update these when I get some time

PR #5659, opened 2024-07-27
feat: add keymap size (@butterflycup)

butterflycup:
The keymap cannot be resized, it can sometimes feel pretty small, especially when learning a new layout. This adds a setting called "Keymap Size" in which you can change the size of the keymap (1 being default, 0.5 minimum and 3.5 as maximum as anything higher wouldn't fit anyways)


frontend/.vscode/settings.json
Miodec, line 1: Remove this file please.

frontend/src/html/pages/settings.html
```diff
@@ -1051,6 +1051,21 @@
     </div>
     <div class="sectionSpacer"></div>
   </div>
+  <div class="section" data-config-name="keymapSize">
+    <div class="groupTitle">
+      <i class="fas fa-keyboard"></i>
+      <span>keymap size</span>
+    </div>
+    <div class="text">Change the scale of the keymap.</div>
```
Miodec, line 1059: scale > size

frontend/src/ts/config.ts
```diff
@@ -1583,6 +1583,31 @@ export function setKeymapShowTopRow(
...
 }
 
+export function setKeymapSize(keymapSize: number, nosave?: boolean): boolean {
+  if (!isConfigValueValid("keymap size", keymapSize, ["number"])) {
+    return false;
+  }
+
+  if (keymapSize < 0) {
+    keymapSize = 1;
+  }
```
Miodec, line 1593: This allows the size to be 0, which could be confusing if the keymap is enabled but still invisible. I would limit this to 0.5

frontend/src/html/pages/settings.html
```diff
@@ -1051,6 +1051,21 @@
...
+    <div class="inputs">
+      <div class="inputAndButton">
+        <input type="number" placeholder="keymap size" class="input" min="0" />
+        <button class="save no-auto-handle">
```
Miodec, line 1063: Update min property to 0.5 and add max with 3.5

Miodec (changes requested):

Miodec:
By scale > size i meant to change the text "scale" to "size", not changing everything to scale :D Sorry for the ambigous comment.

Miodec:
image

butterflycup:
My bad, though in the end I guess scale makes more sense than size anyway, which one do you prefer ?

Miodec:
Other settings use size - lets stick with that.

Miodec (changes requested):
Please hide the settings group if keymap is not enabled (just like the other settings which alter the keymap)

Also, what do you think about changing this to a slider (just like the custom background filter settings?)

image

image

butterflycup:
Keymap setting groups now hides, I prefer number input instead of slider. I'd go for slider if the changes were visible in real time (like for the custom background), but since the keymap is not visible in the settings page (and there's no slider in command line) I prefer number input, though I think both works. Maybe it'd be good to ask other people's opinion?

Miodec:
> Keymap setting groups now hides, I prefer number input instead of slider. I'd go for slider if the changes were visible in real time (like for the custom background), but since the keymap is not visible in the settings page (and there's no slider in command line) I prefer number input, though I think both works. Maybe it'd be good to ask other people's opinion?

Well, there is no slider for any of the settings in the commandline.

The slider makes more sense in this case because the value of this setting is limited between two values, and the range is quite small. @fehmer also liked the slider a bit more.

butterflycup:
I am way out of my comfort zone and I'm guessing this is why there is conflict with my fork (I couldn't figure out why), I'd appreciate the help a lot

fehmer:
> I am way out of my comfort zone and I'm guessing this is why there is conflict with my fork (I couldn't figure out why), I'd appreciate the help a lot

hi @butterflycup , i'm done with syncing your branch with monkeytype master and adapted to the new approach. You can take it from here.

Miodec:
> I think waiting for a slider implementation would unnecessarily postpone this feature of keymap resizing.

Converting to a slider would take like half an hour max.

Miodec:
Ill go ahead and convert to a slider

butterflycup:
Thank you all, I did the best I could, and while Miodec is right that it would take "half an hour max", I have to admit I am a noob and it wouldn't be as easy for me. Thank you for taking over

Miodec:
> Thank you all, I did the best I could, and while Miodec is right that it would take "half an hour max", I have to admit I am a noob and it wouldn't be as easy for me. Thank you for taking over

No worries, it was a big ask to be fair.

PR #5664, opened 2024-07-27
feat: Basic accessibility for users with screen readers (lauritzt)

lauritzt:
### Description

This pull request adds prompts readable by screen readers, informing users using a screen reader (for example blind users) which word to type and which key to press. It also features slight improvements to make the site easier to navigate for screen readers.

> [!NOTE]  
> This will almost certainly require more testing and more work to be "good". Depending on maintainer preference, it may be preferred to wait until it is more complete, though that will obviously be a larger PR to review. I have only done basic testing, though it doesn't seem to break any current functionality.

It does not (yet) contain:

- [ ] Preventing the screen reader from announcing "[word] deleted" when Monkeytype clears the word input box
- [ ] An accessible (-y navigable) results page
- [ ] An accessible (-y navigable) settings page
    - [ ] Information about whether a button is enabled or disabled, such as the mode buttons
- [ ] Prompt customizability
- [ ] Correct language tag for the word to type
- [ ] An intuitive way for the user to get familiar with the application
    - It seems that the user is simply dumped into the word input box without explanation to the user.
 - [ ] Better notification integration [(resource)](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-2/)
 - [ ] Assistance for finding the keys (for example "above left pinky" for Q on QWERTY)

And is probably missing other things I'm forgetting or haven't thought of.

### Checks

    - I couldn't find any.


frontend/src/styles/core.scss
```diff
@@ -418,3 +418,15 @@ key {
...
+.navigationSkipButton {
+  clip: rect(0 0 0 0);
+  width: 1px;
+  height: 1px;
+  margin: -1px;
+  padding: 0;
+  border: 0;
+  overflow: hidden;
+  position: absolute;
+}
```
Miodec, line 432: Where is this used?

Miodec: Nevermind i found it.

Miodec:
Are you planning to split this up into smaller PRs or one big one?

lauritzt:
> Are you planning to split this up into smaller PRs or one big one?

I think I'll just leave this PR as it is now.

The reason is that after working on this PR, I do kind of feel like any more significant effort from my side would be better spent on a separate app specifically for visually impaired users, with focus on learning rather than testing speed. A number of the further improvements I'd like to make seem outside of the scope of Monkeytype or might reduce functionality for sighted users without additional effort (an example requiring a button press to move focus to the word input field).

So I'm not sure how much additional work I'll do on this, at least for now. I do still believe that the current state of the PR is a significant improvement for users with screen readers using Monkeytype, though, so I still hope it is possible to get merged.

Miodec:
> > Are you planning to split this up into smaller PRs or one big one?
> 
> I think I'll just leave this PR as it is now.
> 
> The reason is that after working on this PR, I do kind of feel like any more significant effort from my side would be better spent on a separate app specifically for visually impaired users, with focus on learning rather than testing speed. A number of the further improvements I'd like to make seem outside of the scope of Monkeytype or might reduce functionality for sighted users without additional effort (an example requiring a button press to move focus to the word input field).
> 
> So I'm not sure how much additional work I'll do on this, at least for now. I do still believe that the current state of the PR is a significant improvement for users with screen readers using Monkeytype, though, so I still hope it is possible to get merged.

In that case im not really sure if doing this improved accessibility for just one page makes sense. Because as soon as you leave the test page things will be back to "not really that great", which will point out the issue even more.

PR #5682, opened 2024-07-30
fix(funbox): sometimes generating unrealistic ip addresses

### LINKED ISSUE #5622: IPv6 shortens into invalid addresses

- [X] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

The IPv6 shortening used when punctuation is turned on for IPv6 funbox mode produces invalid addresses.

https://github.com/monkeytypegame/monkeytype/blob/8f4d291fcfb47a187b696d1393853c14807050ee/frontend/src/ts/test/funbox/funbox.ts#L483-L490

Some real and pathalogical cases:
- `b70b:ad23:3d4b:23a9:8000:0000:0000:0000` -> `b70b:ad23:3d4b:23a9:8000:::`
- `0000:0000:0000:0000:0001:0000:0000:0000` -> `::1:::`
- `0000:0000:0000:0001:0002:0000:0000:0000` -> `::1:2:::`

### Expected Behavior

These cases should shorten as follows:
- `b70b:ad23:3d4b:23a9:8000:0000:0000:0000` -> `b70b:ad23:3d4b:23a9:8000::` (two `:`, not three)
- `0000:0000:0000:0001:0000:0000:0000:0000` -> `0:0:0:1::` (only longest series of zero groups gets shortened)
- `0000:0000:0000:0001:0002:0000:0000:0000` -> `::1:2:0:0:0` (if there's a tie, the leftmost is shortened)

### Steps To Reproduce

1. Turn on IPv6 funbox mode.
2. Enable punctuation.

### Environment

_No response_

### Anything else?

IETF has recommenations in [RFC5952](https://datatracker.ietf.org/doc/html/rfc5952) (also described on [Wikipedia](https://en.wikipedia.org/wiki/IPv6_address#Representation)).

I can work on a fix and test cases. The easiest solution might be to pull in an npm package for IP parsing but I'm not sure on what the feeling is around extra dependencies.

rizwanmustafa:
### Description

- [x] Fix IPv6 Compression algorithm
- [x] Write unit tests for random generation of valid IPv4 addresses
- [x] Write unit tests for random generation of valid IPv6 addresses

Closes #5622


rizwanmustafa:
A sample run on NodeJS gives us the following:
image

fehmer:
hi @rizwanmustafa , could you add your examples as unit tests?

fehmer (dismissed):

frontend/src/ts/test/funbox/funbox.ts
```diff
@@ -474,20 +474,18 @@ FunboxList.setFunboxFunctions("IPv4", {
...
     // Compress
     if (w.includes(":")) {
-      w = w
-        .replace(/\b(?:0+:){2,}/, "::")
-        .split(":")
-        .map((a) => a.replace(/\b0+/g, ""))
-        .join(":");
+      // console.log("Original IP: " + w);
+      w = IPAddresses.compressIpv6(w);
+      // console.log("Compressed IP: " + w);
```
Miodec, line 488: Remove logs

frontend/__tests__/utils/ip-addresses.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+      ];
+
+      for (let i = 0; i < rawIps.length; i++) {
+        expect(IpAddresses.compressIpv6(rawIps[i] as string)).toEqual(
+          compressedIps[i]
+        );
+      }
+    });
+  });
+});
```
Miodec, line 34: Should we have some sort of v4 tests? Would there be anything to test?

Miodec (changes requested):

frontend/__tests__/utils/ip-addresses.spec.ts
rizwanmustafa: I believe there could be tests for IPv4. Maybe a check that the numbers don't exceed 255. Example `256.255.255.255` is an invalid Ipv4 address

frontend/src/ts/test/funbox/funbox.ts
rizwanmustafa: ok

frontend/__tests__/utils/ip-addresses.spec.ts
rizwanmustafa: Done

Miodec (approved):
https://media1.tenor.com/m/cOsiTargIp8AAAAC/hes-back.gif

PR #5688, opened 2024-07-30
chore: upgrade node to 20.16.0

Miodec:
Noooooooooode


.github/workflows/monkey-ci.yml
```diff
@@ -68,24 +68,26 @@ jobs:
...
     if: needs.pre-ci.outputs.should-build-be == 'true' || needs.pre-ci.outputs.should-build-fe == 'true' || needs.pre-ci.outputs.should-build-pkg == 'true' || needs.pre-ci.outputs.assets-json == 'true'
     steps:
+
       - name: Checkout package-lock
         uses: actions/checkout@v4
         with:
           sparse-checkout: |
             package-lock.json
 
+      - name: Set up Node.js
```
fehmer, line 78: can be moved down after the fill checkout, we only need this in case of cache miss

PR #5690, opened 2024-07-30
chore: switch from npm to pnpm

Miodec:
pnpm goes brrrrrrr


fehmer:
as a reminder, docker builds  (for docker images + run fe/be in docker) need update as well

PR #5702, opened 2024-07-31
impr(settings): replace sound volume with a slider (@amarnathsama)

### LINKED ISSUE #4764: Replace sound volume settings with a slider

amarnathsama:
### Description
Replaced the volume settings with a slider for better user experience.

### Checks

Closes #4764

Looks like this
image


Miodec:
Thank you for the PR. However, some time recently we added a `range` mode to the `SettingsGroup` class, which means you dont need to create a slider handling file anymore. You should be able to just use `SettingsGroup` like this:

image

(and then ensure the html structure is correct).

Conversion of the config value to a float is good though.

Please update the PR.

amarnathsama:
Hi @Miodec, since changes are being handled by the SettingsGroup class, it shows the actually float value(instead of it being multiplied by 100), is this fine?
image

Miodec:
> Hi @Miodec, since changes are being handled by the SettingsGroup class, it shows the actually float value(instead of it being multiplied by 100), is this fine? image

Yeah

Miodec (approved):
Thanks

PR #5713, opened 2024-08-01
impr: use tsrest for admin endpoint (@fehmer)

fehmer:


backend/src/api/controllers/user.ts
```diff
@@ -160,7 +159,13 @@ export async function sendForgotPasswordEmail(
   req: MonkeyTypes.Request
 ): Promise<MonkeyResponse> {
   const { email } = req.body;
+  await doSendForgotPasswordEmail(email);
+  return new MonkeyResponse(
+    "Password reset request received. If the email is valid, you will receive an email shortly."
+  );
+}
 
+export async function doSendForgotPasswordEmail(email: string): Promise<void> {
```
Miodec, line 168: why the "do" at the beginning?

backend/src/api/controllers/admin.ts
```diff
@@ -1,31 +1,72 @@
...
+  const discordId = user.discordId;
+  const discordIdIsValid = discordId !== undefined && discordId !== "";
+
+  await UserDAL.setBanned(uid, !user.banned);
+  if (discordIdIsValid) await GeorgeQueue.userBanned(discordId, !user.banned);
+
+  return new MonkeyResponse2(`Ban toggled`, {
+    banned: !user.banned,
+  });
 }
```
Miodec, line 41: missing `addImportantLog` call

backend/src/api/controllers/admin.ts
```diff
@@ -75,14 +111,20 @@ export async function handleReports(
...
+
+export async function sendForgotPasswordEmail(
+  req: MonkeyTypes.Request2<undefined, SendForgotPasswordEmailRequest>
+): Promise<MonkeyResponse2> {
+  const { email } = req.body;
+  await authSendForgotPasswordEmail(email);
+  return new MonkeyResponse2(
+    "Password reset request received. If the email is valid, you will receive an email shortly.",
+    null
+  );
```
Miodec, line 129: Since these are the admin endpoints, we can say "password reset email sent", since we know the email is valid.

backend/src/utils/auth.ts
```diff
@@ -82,3 +85,24 @@ export async function revokeTokensByUid(uid: string): Promise<void> {
...
+      .auth()
+      .generatePasswordResetLink(email, {
+        url: isDevEnvironment()
+          ? "http://localhost:3000"
+          : "https://monkeytype.com",
+      });
+
+    await emailQueue.sendForgotPasswordEmail(email, name, link);
+  } catch {}
+}
```
Miodec, line 108: I think catching the error and not doing anything with it could be dangerous? It might lead to us not knowing about a 500 error.

I think we need to check if the error is a "email/user not found" and only ignore it then.

packages/contracts/src/admin.ts
```diff
@@ -0,0 +1,117 @@
...
+        200: MonkeyResponseSchema,
+      },
+    },
+  },
+  {
+    pathPrefix: "/admin",
+    strictStatusCodes: true,
+    metadata: {
+      openApiTags: "admin",
+      authenticationOptions: { noCache: true }, //TODO test
```
Miodec, line 112: Tested?

fehmer: yes

Miodec (approved):

PR #5715, opened 2024-08-02
impr(pace caret): add tags pb mode (@NadAlaba)

### LINKED ISSUE #5458: Pace caret mode: tag pb
take the highest pb from active tags and set the pace caret speed to that speed

NadAlaba:
### Description

- Add a pace caret mode "tags pb" which takes the highest wpm of results that have **_all_** current active tags.

### Discussion

- **[EDIT]**  
**_all_** mode was removed.

- When there are no active tags pace caret wpm gets set to 0: Other approaches could be to throw an error (which is imo an overkill)

Closes #5458


frontend/src/ts/db.ts
```diff
@@ -580,6 +580,73 @@ export async function getUserDailyBest<M extends Mode>(
...
+  if (!snapshot || !snapshot.results) return 0;
+
+  const activeTagIds: string[] = [];
+  snapshot.tags?.forEach((tag) => {
+    if (tag.active === true) activeTagIds.push(tag._id);
+  });
+
+  let highestWpm = 0;
+
+  if (anyTag && activeTagIds.length > 0) {
```
fehmer, line 605: could be simplified to sth like this (untested)
```ts
snapshot.tags?
  .filter(it=>it.active)
  .map(it=> getLocalTagPB(it._id, ...))
  .reduce((acc,cur)=> Math.max(acc,cur),0);
 ```

Miodec:
Thanks

PR #5716, opened 2024-08-02
impr: use tsrest for public endpoints (@fehmer)

fehmer:


packages/contracts/src/schemas/util.ts
```diff
@@ -1,8 +1,13 @@
...
-  return typeof val === "string" ? /^\d+$/.test(val) : false;
-});
+export const StringNumberSchema = z
+
+  .custom<`${number}`>((val) => {
+    if (typeof val === "number") val = val.toString();
+    return typeof val === "string" ? /^\d+$/.test(val) : false;
+  }, "Needs to be a number represented as a string.")
+  .transform(String)
+  .describe('number wrapped as string, e.g. "10"');
```
fehmer, line 10: update error message, can be either a number or a string

PR #5726, opened 2024-08-04
fix(caret): jumping in word tape mode and other issues (@NadAlaba)

NadAlaba:
### Description

1. In word tape mode, caret jumps forward if first letter of a word was typed before the tape finished scrolling back (In the gif below, the letter "c" in "come" was typed before the word finished scrolling back to center, so the caret jumped to the letter "e" before even the "o" was typed).  
This happens because the caret was referencing letters' positions while they were changing.  
tape
3. When a full width caret is at a zero-width character (e.g, a diacritic), that has a preceding letter that is also zero-width, the caret's width gets set to the last letter's width (which might be wider than it should be like in the gif below).
width
4. In blind mode or hide extra letters in RTL languages, caret of extra letters is getting positioned at the last letter instead of at the space after last letter.  
extra
5. In non-monospaced fonts, the caret at the space between words may be narrower or wider than the actual space (because it gets its width from the width of last letter).  
Now, space width is getting calculated, so that the caret don't depend on the letters' width.  
space


Miodec (approved):
Thanks

PR #5736, opened 2024-08-06
ci: reduce checkout for prettier (@fehmer)

fehmer:
testing, don't merge


fehmer:
sparse checkout in this case doesn't speed up the job
image

PR #5737, opened 2024-08-06
ci: skip cache download on prime-cache job (@fehmer)

fehmer:
Don't download the cache on prime-cache. We only need the  `cache-hit` flag.

Reduces the runtime by six seconds.


Miodec (approved):
Nice

PR #5743, opened 2024-08-08
refactor: move release code to a package (@miodec)

Miodec:
Moves all release related things to @monkeytype/release


package.json
```diff
@@ -29,11 +29,11 @@
     "start-fe": "turbo run start --filter @monkeytype/frontend",
     "docker": "cd backend && docker compose up",
     "audit-fe": "cd frontend && npm run audit",
-    "release": "node ./bin/release.mjs",
-    "release-fe": "node ./bin/release.mjs --fe",
-    "release-be": "node ./bin/release.mjs --be",
-    "release-no-deploy": "node ./bin/release.mjs --no-deploy",
-    "release-dry": "node ./bin/release.mjs --dry",
+    "release": "monkeytype-release",
```
fehmer, line 32: don't we need `build` first?

packages/release/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@monkeytype/release",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "nodemon --watch src --exec 'npm run release-dry'",
+    "dev-changelog": "nodemon ./src/buildChangelog.js",
+    "release": "node ./src/index.js",
```
fehmer, line 8: unused? you call them from the root level?

package.json
Miodec: handled by the release script (all though i just realised the path will be wrong)

packages/release/package.json
Miodec: removing

fehmer (approved):
👍

PR #5746, opened 2024-08-08
chore: add test for recordClientVersion (@fehmer)

fehmer:
- add tests to verify recordClientVersion is called
- remove tailing slash for `/presets` endpoints


Miodec (approved):
Thanks

PR #5750, opened 2024-08-09
refactor: implement recordClientVersion as middleware (@fehmer)

fehmer:
- move recordClientVersion from `index` to `psa` route
- move docs/v2 CSP header from `index` to `docs` route


Miodec (approved):
Thanks

Issue #5756, opened 2024-08-09
English_1k is actually 999 words


bskdany:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

Yes

### Account name

bskdany

### Expected Behavior

It's 1000 words and not 999

### Steps To Reproduce

get the length of the words array from here https://github.com/monkeytypegame/monkeytype/blob/master/frontend/static/languages/english_1k.

code example: https://www.mycompiler.io/view/0kUNNQfFl1j

### Environment

- OS:
- Browser:
- Browser Version:

HalfToothed:
hey if you are not working on this, i would like to fix it

avick-saha:
I've already resolved the issue.

HalfToothed:
ok great.

PR #5757, opened 2024-08-10
fix(language): add one thousandth word to english 1k (@avick-saha)

avick-saha:
### Description
English_1k is actually 999 words. So, I've added one more English word "universe".

Resolves #
Issue #5756


Miodec:
Thanks

PR #5758, opened 2024-08-10
impr: use ts-rest for results endpoint (@fehmer)

fehmer:


backend/redocly.yaml
```diff
@@ -5,11 +5,11 @@ apis:
   internal@v2:
     root: dist/static/api/openapi.json
   public-filter:
-    root: dist/static/api/openapi-tagged.json
+    root: dist/static/api/openapi.json
     decorators:
       filter-in:
         property: x-public
-        value: yes
+        value: [yes, true]
```
fehmer, line 12: revert

frontend/src/ts/db.ts
```diff
@@ -311,13 +311,6 @@ export async function getUserResults(offset?: number): Promise<boolean> {
     }
     if (result.afkDuration === undefined) result.afkDuration = 0;
     if (result.tags === undefined) result.tags = [];
-
-    if (
```
fehmer, line 315: code moved to backend (`controllers/result.ts convertResult` ) and legacy properties `correctChars` and `incorrectChars` removed from the rest api

Issue #5692, opened 2024-07-30
Save custom text button not enabled


brenob6:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

Yes

### Account name

brenob6

### Current Behavior

The save button on the custom text dialog does not enable when the long text checkbox is marked before filling in the input field

### Expected Behavior

The button should be enabled.

### Steps To Reproduce

In homepage click in custom, change, save, mark the checkbox and then enter a name.

### Environment

- OS: Arch
- Browser: Firefox
- Browser Version: 128.0

brenob6:
I can work on this.

rizwanmustafa:
I have changed the title for better understanding. It was confusing before

Uneeqbox:
It seems that the "Save custom text" button is not enabled. Please check if all required fields are filled out correctly and if there are any specific conditions that need to be met for the button to be activated. You can visit [this link](https://uneeqbox.com/all-product/) for more details on our products and services.

Sumedhbhat:
I am able to replicate this bug. It occurs when a user clears the input and enters a value then the button stays disabled. 
If it hasnt been addressed yet can I pick this up?

PR #5759, opened 2024-08-10
fix(save custom text modal): save button not being enabled (@brenob6)

brenob6:
### Description

Solved bug #5692


Miodec:
Thanks

PR #5762, opened 2024-08-11
chore: add test case for apekey permission bug (@fehmer)

fehmer:
cover fix in 77c9cc423f288aec08d729a56c41edceef16ff7f


Miodec (approved):
Thanks

PR #5763, opened 2024-08-11
impr: add local storage with schema class (@miodec)

Miodec:
Add more typesafety when using local storage.

!nuf


packages/contracts/src/schemas/users.ts
```diff
@@ -1 +1,66 @@
...
+    off: z.boolean(),
+  }),
+  date: z.object({
+    last_day: z.boolean(),
+    last_week: z.boolean(),
+    last_month: z.boolean(),
+    last_3months: z.boolean(),
+    all: z.boolean(),
+  }),
+  tags: z.record(z.boolean()),
```
fehmer, line 57: is this the same as ` z.record(z.string(), z.boolean())`

Miodec: it didnt complain, but ill fix it

fehmer (approved):

PR #5788, opened 2024-08-14
fix: result filters getting reset on page load (@fehmer)

fehmer:
Result settings getting reset to defaults on page load

Steps to reproduce:
- go to account page
- set filter to last week
- reload the page


Miodec (approved):
Thanks

PR #5816, opened 2024-08-22
docs: replace npm commands with pnpm (@anthonypz)

anthonypz:
Since Monkeytype uses pnpm, this change replaces all of the `npm run` commands with `pnpm run` for consistency and to avoid any unexpected dependency resolution issues.

### Checks

- [x] Adding quotes?
  - [x] Make sure to include translations for the quotes in the description (or another comment) so we can verify their content.
- [x] Adding a language or a theme?
  - [x] If is a language, did you edit `_list.json`, `_groups.json` and add `languages.json`?
  - [x] If is a theme, did you add the theme.css?

Closes #

### DIFF
--- a/docs/CONTRIBUTING_ADVANCED.md
+++ b/docs/CONTRIBUTING_ADVANCED.md
@@ -129,20 +129,20 @@ Run `pnpm i` in the project root to install all dependencies.
 Manual:
 
 ```
-npm run dev
+pnpm run dev
 ```
 
 ### Backend only
 
 | Manual                      | Docker                         |
 | --------------------------- | ------------------------------ |
-| `npm run dev-be`            | `cd backend && npm run docker` |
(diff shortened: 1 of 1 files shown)


Miodec:
I mean, npm is bundled with node and its quicker to type. So far ive had 0 issues when running commands with npm.

anthonypz:
Ah, that's a very good point on it being quicker to type.

It looks like there is only one small difference between `npm run` and `pnpm run`, according to the [pnpm docs](https://pnpm.io/8.x/cli/run#differences-with-npm-run):

> By default, pnpm doesn't run arbitrary pre and post hooks for user-defined scripts (such as prestart). 

Monkeytype doesn't use pre or post scripts so you're correct that `npm run` should be safe.

anthonypz:
PR changes not necessary.

Issue #3907, opened 2023-01-11
Tape mode + Smooth line scroll + enter


KorolOrol:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

Yes

### Account name

KorolOrol

### Account config

[...]

PR #5824, opened 2024-08-26
fix(zen mode): issues when changing tape mode and other caret issues (@NadAlaba)

NadAlaba:
### Description

1. Changing tape mode in zen mode removes a `.word` element which makes the word undefined. To solve this, the code that removes `.word` was moved to where that extra element was created (in `showWords()`) so that only what was created will be removed.
tapeChange

2. After applying previous fix, full-width caret gets shrunk to zero width when changing tape mode at the beginning of a new word.
This was happening due to concurrent calling of `Caret.updatePosition()` causing one call to [remove the underscore element](https://github.com/monkeytypegame/monkeytype/blob/30d440a70e56abfdfb889e62646cb5cc04ad08aa/frontend/src/ts/test/caret.ts#L249), while the other call is still using it.
These 2 calls are happening when hiding the commandline in [`animated-modal.ts`](https://github.com/monkeytypegame/monkeytype/blob/30d440a70e56abfdfb889e62646cb5cc04ad08aa/frontend/src/ts/utils/animated-modal.ts#L384-L388) through `this.wrapperEl.close()` and `await options?.afterAnimation?.(this.modalEl)` which are both triggering `#wordsInput` focus which is calling `Caret.updatePosition()`.
The chosen solution was to check if `#wordsInput` is focused before triggering a focus in the second call.
Other solution might be to debounce `Caret.updatePosition()`?
caretWidth

3. Because of the letter '_' that is [added](https://github.com/monkeytypegame/monkeytype/blob/30d440a70e56abfdfb889e62646cb5cc04ad08aa/frontend/src/ts/test/caret.ts#L140-L143) to the beginning of a word in zen mode, the caret is getting positioned on the 2nd letter in RTL languages.
Also, since `lastWordLetter` is always undefined (because wordLen is always 0) `offsetHeight` and `offsetTop` are getting the default values regardless of font characteristics.
caretRTL

4. In zen mode + word tape mode + blind mode, caret stays positioned at the first letter (because `wordLen` is always zero in zen mode).
blindTapeWord

### Known remaining issues

- `[ENTER]` causes problems in tape mode (in zen mode and other modes). A solution is planned on a future PR solving #3907.


NadAlaba:
All issues are still there after merging master

PR #5825, opened 2024-08-26
fix(release script): not updating root `package.json` version + not supporting some shells (@NadAlaba)

NadAlaba:
### Description

1. When updating the version, the script is getting current version from root `package.json`, but updating `release/package.json`, so root `package.json` never gets updated and minor will always be zero.
Chosen solution was to get version from root, but update both.
Other possible solution is to not use version in `release/package.json` at all (i.e, get and set version in root only).
**[EDIT]**
Solution changed to removing version from all `package.json` other than root one, and only get and update it in root.

2. Change single quotes in `nodemon --exec '...'` to double quotes.
releaseScript

3. Changes in `buildChangelog.getLog()` are because some shells don't support evaluating subexpressions in `$(...)`.

4. Change root `package.json` version to current version.

5. In commit titles, trailing parenthesis that are not usernames or PR numbers are being considered PRs. So now usernames get matched when they start with `@` and PRs when they start with `#`.
log

### DIFF
--- a/packages/release/src/buildChangelog.js
+++ b/packages/release/src/buildChangelog.js
@@ -251,19 +251,11 @@ function convertStringToLog(logString) {
 
     //split message using regex based on fix(language): spelling mistakes in Nepali wordlist and quotes (sapradhan) (#4528)
     //scope is optional, username is optional, pr number is optional
-    const [_, type, scope, message, message2, message3] = title.split(
-      /^(\w+)(?:\(([^)]+)\))?:\s+(.+?)\s*(?:\(([^)]+)\))?(?:\s+\(([^)]+)\))?(?:\s+\(([^)]+)\))?$/
+    const [_, type, scope, message, username, pr] = title.split(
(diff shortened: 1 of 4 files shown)


Miodec:
> ### Description
> 1. When updating the version, the script is getting current version from root `package.json`, but updating `release/package.json`, so root `package.json` never gets updated and minor will always be zero.
>    Chosen solution was to get version from root, but update both.
>    Other possible solution is to not use version in `release/package.json` at all (i.e, get and set version in root only).

I would change it to only update the root package.json and remove the version numbes from others.

packages/release/src/buildChangelog.js
```diff
@@ -276,7 +270,7 @@ function convertStringToLog(logString) {
         scope,
         message,
         usernames: usernames || [],
-        prs,
+        prs: prs || [],
         body: body || "",
       });
```
Miodec, line 275: Have you tested this change?

NadAlaba: Yeah it's tested.
It's not really needed (as far as my tests went), but I added it just for consistency with usernames.
If you can tell me what errors were you avoiding on 0a23897ae, maybe we can remove it from both usernames and prs.

Miodec (approved):
Thanks

PR #5826, opened 2024-08-26
fix(presets): Bug fix for auto-switch theme checkbox values (@amarnathsama)

amarnathsama:
### Description

Fixes bug where applying a preset wouldn't update the auto switch theme light and dark theme checkbox values (pictured below).
image

### Checks


frontend/src/ts/config.ts
Miodec, line 1: I would prefer to not add more DOM modifying code in the config module. Please use the config event instead and add this code in settings.ts

Miodec (changes requested):

frontend/src/ts/config.ts
amarnathsama: I have made the changes, please let me know if they are good or not.

frontend/src/ts/pages/settings.ts
```diff
@@ -1326,6 +1326,18 @@ ConfigEvent.subscribe((eventKey) => {
   }
 });
 
+ConfigEvent.subscribe((eventKey, eventValue) => {
```
Miodec, line 1329: Please move this inside the subscribe function thats just above.

Miodec (changes requested):

frontend/src/ts/pages/settings.ts
amarnathsama: done

Miodec (approved):
Thanks

PR #5834, opened 2024-08-30
refactor: determine active word based on logical value, not DOM class (@NadAlaba)

NadAlaba:
### Description

A continuation of what was done in c3274f0


fehmer (changes requested):
ignore me, testing labels

Miodec:
Thanks

PR #5842, opened 2024-09-01
chore(PR labeler): fix workflow having read access only, when triggered from forks (@NadAlaba)

NadAlaba:
### Description

- Previous automatic PR labeler doesn't work when triggered from forks because workflows triggered from forks by the events `pull_request` and `pull_request_review` have read access only, and can't write to PRs (change labels).

- This is an attempt to solve this issue by splitting the workflow in 2:
One with read access only, that gets triggered by those events, and when completed another workflow would run with write access that would perform the change.

- I added everything I could think of to the list of conditions to update the labels (because removing conditions later is easier than going back to the docs and adding more conditions). You may need to go over this list and remove what you deem unnecessary.


.github/workflows/update-labels.yml
```diff
@@ -0,0 +1,92 @@
...
+          (
+            github.event_name == 'issue_comment' &&
+            contains(fromJSON('["created", "edited"]'), github.event.action) &&
+            github.event.comment.user.id == github.event.issue.user.id
+          )
+        run: echo "UPDATE=-1" >> $GITHUB_ENV
+
+      - name: Save result in a JSON file
+        env:
+          RESULT_JSON: ${{ format('{{"waiting_for_review"{0} {1}, "waiting_for_update"{0} {2}, "pr_url"{0} "{3}"}}', ':', env.REVIEW, env.UPDATE, env.PR_URL) }}
```
fehmer, line 85: let's only include the pr number here

.github/workflows/write-labels.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Populate variables from JSON values
+        run: |
+          echo "WAITING_FOR_REVIEW=${{ fromJSON(env.JSON_TEXT).waiting_for_review }}" >> $GITHUB_ENV
+          echo "WAITING_FOR_UPDATE=${{ fromJSON(env.JSON_TEXT).waiting_for_update }}" >> $GITHUB_ENV
+          echo "PR_URL=${{ fromJSON(env.JSON_TEXT).pr_url }}" >> $GITHUB_ENV
+
+      - name: Add 'waiting for review' label
+        if: env.WAITING_FOR_REVIEW == 1
+        run: gh pr edit ${{ env.PR_URL }} --add-label "waiting for review"
```
fehmer, line 36: I'm not 100% sure this can be used for code injection. To be safe lets use the github actions you used previously and pass in the pr number only.

.github/workflows/write-labels.yml
```diff
@@ -0,0 +1,48 @@
+name: Write label on PR/issue
+
+permissions:
+  pull-requests: write
+  issues: write
```
fehmer, line 5: why do we need write access on issues?

.github/workflows/update-labels.yml
```diff
@@ -0,0 +1,92 @@
...
+        run: echo $RESULT_JSON > result.json
+
+      - name: Upload the JSON file
+        uses: actions/upload-artifact@v4
+        with:
+          name: result_json
+          path: ./result.json
```
fehmer, line 92: can we use a better name? e.g. `write-labels.json`

fehmer (changes requested):

.github/workflows/write-labels.yml
NadAlaba: the labeler removes the 'waiting for update' label on comments by the author on issues too.
So you can add the label manually and it'll get removed automatically.

.github/workflows/update-labels.yml
NadAlaba: sure

.github/workflows/write-labels.yml
NadAlaba: I made the change although I don't fully get the risk

fehmer (approved):
lgtm

Miodec:
Thanks

PR #5855, opened 2024-09-06
feat(layout): Re-add the rain layout (@ibrab000)

ibrab000:
### Description

Added back the rain layout. I saw that forusu is also adding a layout so I put her layout first and then added mine, so probably accept hers first.


Miodec:
Please resolve conflict and fix CI issues.

PR #5856, opened 2024-09-06
fix(caret): getting last letter top in long multiline words (@NadAlaba)

NadAlaba:
### Description

- When getting to a long multi-line word, the caret jumps to the last line of that word instead of the line of the first letter.
long-word-caret


Miodec (approved):
Thanks

PR #5857, opened 2024-09-06
fix: multi-line words causing issues in ui scrolling and tape mode (@NadAlaba)

NadAlaba:
### Description
1. When text is a single long multi-line word that takes up all of the screen, `#wordsInput` position will be below the visible screen which requires scrolling down.
    - This PR tweaks `#wordsInput` position calculation by considering a `word` height as the height of a single line of a word.
    - For visual illustration of this bug see this [video on discord server](https://discord.com/channels/713194177403420752/713195089303830548/1281633634650165278)
2. When a long multi-line word is used in tape mode, this word will wrap to multiple lines causing inability to type the letters which are below the first line.
    - The chosen solution was to use `white-space: nowrap` on `.tape .word` elements.
    - For visual illustration of this bug see [this](https://discord.com/channels/713194177403420752/713195115275091969/1281640882902401128), and [this](https://discord.com/channels/713194177403420752/713195089303830548/1281634954203369555) videos on discord server.

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -263,6 +263,7 @@
 
   &.tape .word {
     margin: 0.25em 0.6em 0.75em 0;
+    white-space: nowrap;
   }
 
   /* a little hack for right-to-left languages */
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -530,8 +530,10 @@
 }
 
 #wordsInput {
-  font-size: 1rem;
-  opacity: 0;
+  width: 1ch;
+  font-size: 1em;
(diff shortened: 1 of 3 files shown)


frontend/src/styles/test.scss
```diff
@@ -529,8 +530,10 @@
 }
 
 #wordsInput {
-  font-size: 1rem;
-  opacity: 0;
+  width: 1ch;
+  font-size: 1em;
+  height: 1em;
+  // opacity: 0;
```
NadAlaba, line 536: don't merge without me

Miodec (approved):
Thanks

PR #5863, opened 2024-09-10
impr: Migrate user deletion to asynchronous job (@Bruception)

Bruception:
### Checks

- [x] Adding quotes?
  - [x] Make sure to include translations for the quotes in the description (or another comment) so we can verify their content.
- [x] Adding a language or a theme?
  - [x] If is a language, did you edit `_list.json`, `_groups.json` and add `languages.json`?
  - [x] If is a theme, did you add the theme.css?

Closes #

### DIFF
--- a/backend/src/api/controllers/user.ts
+++ b/backend/src/api/controllers/user.ts
@@ -38,7 +38,7 @@ import {
   TestActivity,
   UserProfileDetails,
 } from "@monkeytype/contracts/schemas/users";
-import { addImportantLog, addLog, deleteUserLogs } from "../../dal/logs";
+import { addImportantLog, addLog } from "../../dal/logs";
 import { sendForgotPasswordEmail as authSendForgotPasswordEmail } from "../../utils/auth";
 import {
   AddCustomThemeRequest,
@@ -231,39 +231,7 @@ export async function deleteUser(
 ): Promise<MonkeyResponse2> {
   const { uid } = req.ctx.decodedToken;
 
-  const userInfo = await UserDAL.getPartialUser(uid, "delete user", [
-    "banned",
-    "name",
-    "email",
-    "discordId",
-  ]);
-
-  if (userInfo.banned === true) {
-    await BlocklistDal.add(userInfo);
-  }
-
-  //cleanup database
-  await Promise.all([
-    UserDAL.deleteUser(uid),
-    deleteUserLogs(uid),
-    deleteAllApeKeys(uid),
-    deleteAllPresets(uid),
-    deleteConfig(uid),
-    deleteAllResults(uid),
-    purgeUserFromDailyLeaderboards(
-      uid,
-      req.ctx.configuration.dailyLeaderboards
-    ),
-  ]);
-
-  //delete user from
-  await AuthUtil.deleteUser(uid);
-
-  void addImportantLog(
-    "user_deleted",
-    `${userInfo.email} ${userInfo.name}`,
-    uid
-  );
+  await UserDAL.softDeleteUser(uid);
 
   return new MonkeyResponse2("User deleted", null);
 }
--- a/backend/src/jobs/delete-users.ts
+++ b/backend/src/jobs/delete-users.ts
@@ -1,0 +1,67 @@
+import { CronJob } from "cron";
+
+import { deleteAllApeKeys } from "../dal/ape-keys";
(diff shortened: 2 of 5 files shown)


Bruception:
The main benefit of doing this async is that it provides stronger consistency guarantees. Consider the case where the deletion is interrupted by a server restart or network interruption. This results in an inconsistent state of partial deletion.

Probably needs additional consideration in the frontend to prevent soft deleted users from accessing their account.

backend/src/api/controllers/user.ts
```diff
@@ -231,39 +231,7 @@ export async function deleteUser(
...
-
-  //delete user from
-  await AuthUtil.deleteUser(uid);
-
-  void addImportantLog(
-    "user_deleted",
-    `${userInfo.email} ${userInfo.name}`,
-    uid
-  );
+  await UserDAL.softDeleteUser(uid);
```
fehmer, line 234: some actions need to be done immediately:
- remove or deactivate users ape keys
- remove users token from auth cache

backend/src/jobs/delete-users.ts
```diff
@@ -0,0 +1,65 @@
...
+import { addImportantLog, deleteUserLogs } from "../dal/logs";
+import { deleteAllPresets } from "../dal/preset";
+import { deleteAll as deleteAllResults } from "../dal/result";
+import * as UserDAL from "../dal/user";
+import { getCachedConfiguration } from "../init/configuration";
+import * as AuthUtil from "../utils/auth";
+import { purgeUserFromDailyLeaderboards } from "../utils/daily-leaderboards";
+import { mapLimit } from "../utils/misc";
+
+const CRON_SCHEDULE = "*/10 * * * *"; // every 10 minutes
```
fehmer, line 15: can two cron jobs run at the same time, if one takes longer than 10 minutes to complete? If yes, should we prevent this?

fehmer:
Hi @Bruception ,

> Probably needs additional consideration in the frontend to prevent soft deleted users from accessing their account.

I think we should'n rely on the frontend to handle this but instead handle a soft-deleted user like a unknown/deleted user on the backend. For this we need to adjust the UserDal `getUser` and `getUserPartial` to throw on soft-deleted users.

backend/src/jobs/delete-users.ts
```diff
@@ -0,0 +1,65 @@
...
+    "user_deleted",
+    `${userInfo.email} ${userInfo.name}`,
+    uid
+  );
+
+  await UserDAL.deleteUser(uid);
+}
+
+async function deleteUsers(): Promise<void> {
+  const softDeletedUsers = await UserDAL.getSoftDeletedUsers(DELETE_BATCH_SIZE);
```
fehmer, line 56: shouldn't this be a loop? fetch x users and delete them until no sof-deleted users are left?

backend/src/utils/misc.ts
```diff
@@ -343,3 +343,35 @@ export function replaceObjectIds<T extends { _id: ObjectId }>(
   if (data === undefined) return data;
   return data.map((it) => replaceObjectId(it));
 }
+
+type MapLimitIteratee<T, V> = (element: T, index: number) => Promise<V>;
+
+export async function mapLimit<T, V>(
```
fehmer, line 349: can we improve the naming and add some jsdoc?

e.g. 
```
processPooled(input:T[], concurrency: number, fn:MapLimitIteratee)
```

fehmer: just in case we don't want to implement this ourselves:
https://github.com/sindresorhus/p-limit

backend/src/jobs/delete-users.ts
Bruception: Yes, but given that deletion is idempotent, there should be no issue. Though reviewing this code again there are some areas of improvement I've identified.

backend/src/jobs/delete-users.ts
Bruception: The intention was for this job to lazily delete users every 10 minutes. But yeah we can do it all in one run.

backend/src/utils/misc.ts
fehmer:
```ts
mapConcurrently<T,V>(
  input: T[],
  fn:({element:T, index:number}))=>Promise<V>,
  options?: {maxConcurrency?:number}
)=> Promise<V[]>
```

PR #5866, opened 2024-09-10
feat: reduce motion if requested (@fehmer)

fehmer:
Disable unnecessary animations if the user prefers reduced motion.


Miodec (changes requested):
Animated modals are still animated (alerts panel, forgot password popup and so on...)

Miodec (approved):
Thanks

PR #5872, opened 2024-09-11
refactor: enable no-unsafe-argument rule (@miodec)

Miodec:
More protection against `any`


backend/src/api/controllers/webhooks.ts
```diff
@@ -8,7 +8,7 @@ export async function githubRelease(
 
   if (action === "published") {
     const releaseId = req.body.release.id;
-    await GeorgeQueue.sendReleaseAnnouncement(releaseId);
+    await GeorgeQueue.sendReleaseAnnouncement(releaseId as string);
```
fehmer, line 11: not needed aftter the ts-rest  migration

PR #5876, opened 2024-09-11
fix(quote search): search service not building in code_javascript quotes (@NadAlaba)

NadAlaba:
### Description

`stemmedToken in normalizedTokenToOriginal` always returns `true` when `stemmedToken` is `"constructor"`.
This is not the case for `Object.hasOwn(normalizedTokenToOriginal, stemmedToken)`.


Miodec (approved):
Thanks

PR #5877, opened 2024-09-11
fix(custom font): allow underscore in custom-font's name (@NadAlaba)

NadAlaba:


fehmer:
@NadAlaba could you find out if there is any reason not to allow spaces in config values?

If there is a reason i'd rather use an existing encoding like [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI).

NadAlaba:
> @NadAlaba could you find out if there is any reason not to allow spaces in config values?
> 
> If there is a reason i'd rather use an existing encoding like [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI).

wouldn't it need db migration? is it worth it?

PR #5894, opened 2024-09-17
refactor: move common util functions to a package (@miodec)

Miodec:
Also moves esbuild to a package.


packages/util/src/numbers.ts
```diff
@@ -0,0 +1,126 @@
...
+      ? (nums[mid] as number)
+      : ((nums[mid - 1] as number) + (nums[mid] as number)) / 2;
+  } catch (e) {
+    return 0;
+  }
+}
+
+/**
+ * Calculates the consistency of variance (CoV) of an array of numbers.
+ * @param cov The CoV of the input array (standard deviation / mean).
```
NadAlaba, line 71:
```suggestion
* Calculates consistency by mapping COV from [0, +infinity) to [100, 0).
* The mapping function is a version of the sigmoid function tanh(x) that is closer to the identity function tanh(arctanh(x)) in [0, 1).
 * @param cov The coefficient of variation of an array of numbers (standard deviation / mean).
```

backend/tsconfig.json
```diff
@@ -11,8 +11,7 @@
   "ts-node": {
     "files": true
   },
-  "files": ["./src/types/types.d.ts"],
-  "include": ["./src/**/*"],
+  "include": ["src", "scripts"],
```
fehmer, line 14: not sure if this makes a difference or not.

scripts is not part of the deployed application. Should it be part of the typescript config? We have a tsconfig in https://github.com/monkeytypegame/monkeytype/blob/master/backend/scripts/tsconfig.json same as we handle it for `__tests__`.

fehmer (dismissed):

fehmer (approved):

PR #5895, opened 2024-09-17
impr: show xp gain details as list (@fehmer)

fehmer:
image
image


frontend/src/ts/modals/dev-options.ts
fehmer, line 1: changes only for easy testing

mmonetha:
I don't know if this is the right place to comment on the feature. But I just wanted to praise everyone that has been involved in developing the feature. I didn't know I needed this, but I saw it and immediately felt super happy with the way all the details are displayed. 

God bless all of you 🙏

PR #5900, opened 2024-09-18
fix: random theme is applied over custom theme (@fehmer)

fehmer:
When having a custom theme applied and random theme set both get applied.

Custom theme:
image

Custom theme mixed with chaos theory
image

Custom theme mixed with terrazzo
image


Miodec (approved):
Thanks

PR #5904, opened 2024-09-20
impr(quotes): add german quotes (@rojadesign)

rojadesign:
### Description
- added 5 quotes by Fyodor Mikhailovich Dostoevsky:

1. 	> ‘There is only love in the world because people believe in immortality. With the cessation of faith, love would be destroyed.’

2. 	> ‘The mirror of nature is the clearest mirror! You have to look at it, enjoy it.’

3. 	> ‘If you want to get to know a person properly and find out something about his innermost nature, don't bother to analyse how he speaks, is silent, cries or is moved by lofty thoughts. All you have to do is watch him laugh. If he has a good laugh, he is a good person.’

4. 	> ‘A single word, spoken with conviction in full sincerity and without wavering while standing face to face, says far more than a few dozen sheets of written paper.’

5. 	> ‘There are faces that bring something new every time they appear, something you haven't noticed about them before, even if you've met them a hundred times.’

### Checks

- [X] Adding quotes?
  - [X] Make sure to include translations for the quotes in the description (or another comment) so we can verify their content.


Miodec:
Thanks

PR #5912, opened 2024-09-23
chore: updated scatman quote (@DanielGrenehed)

DanielGrenehed:
### Description
Updated frontend/static/quotes/english.json scatman quote (id: 3328) to include "...check out my message...", instead of "...check my message...", making sure the quote corresponds correctly to the lines of the song it is quoting.


Miodec:
Thanks

PR #5914, opened 2024-09-23
fix: command palette not closing with reduced motion (@fehmer)

fehmer:
- commandline was flickering with reduced motion setting the animation time to zero


fehmer:
mio found a better fix

PR #5918, opened 2024-09-25
impr(code go): add more keywords to Golang (@rammba)

rammba:
I've added some keywords to Golang mostly based on the following article: https://go.dev/tour/basics/11


Miodec:
Thanks

PR #5920, opened 2024-09-26
refactor: split updateWordsHeight() and move some style changes to .scss files (@NadAlaba)

NadAlaba:
### Description

1. Split the tasks of `updateWordsHeight()` into `updateWordsWrapperHeight()` and `updateWordsMargin()` and the remaining tasks are done in `test.scss` (`width` and `overflow`) since they depend on the `.tape` class only.
2. `#words` height is no longer changed since it does not have visible effects.
3. Previously a temporary `<div.word>word height</div>` was added in `showWords()` in zen mode to help in calculating word height in `updateWordsHeight()`, then removed after `updateWordsHeight()` finishes. Now a dummy '_' letter is added to the active word instead, and the addition and removal is done in `updateWordsWrapperHeight()`, because it fixes a bug when turning off tape mode on an empty word in a new line in zen mode, `#wordsWrapper` height would become smaller than it should be, and the new lines would not be visilbe in tape=off.
tapeOff
4. Previously, `#words` margin-left was unset and `scrollTape()` was called when config.tapeMode was changed. This is not needed since these will be called in `updateWordsMargin()` which is called in `updateWordWrapperClasses()` which is called when config.tapeMode is changed.


frontend/src/ts/test/test-ui.ts
```diff
@@ -148,7 +148,8 @@ ConfigEvent.subscribe((eventKey, eventValue, nosave) => {
   }
   if (eventKey === "fontSize" && !nosave) {
     OutOfFocus.hide();
-    updateWordsHeight(true);
+    updateWordsWrapperHeight(true);
+    updateWordsMargin();
     void updateWordsInputPosition(true);
```
Miodec, line 153: Instead of these, maybe we should just use `updateWordWrapperClassses`?

frontend/src/ts/test/test-ui.ts
```diff
@@ -487,101 +476,85 @@ export async function updateWordsInputPosition(initial = false): Promise<void> {
...
     setTimeout(() => {
-      $("#words").css("height", finalWordsHeight + "px");
-      $("#wordsWrapper").css("height", finalWrapperHeight + "px");
-      $(".outOfFocusWarning").css(
-        "margin-top",
-        finalWrapperHeight / 2 - convertRemToPixels(1) / 2 + "px"
-      );
-    }, 0);
+      $("#words").css("margin-left", "unset");
+    }, 125);
```
Miodec, line 557: We should either remove this delay or reduce it to 0 because it creates a visual bug where the words remain offset to the right when switching tape mode off:

https://github.com/user-attachments/assets/2ff1d497-5740-40e7-9c09-2acce610cde0

Miodec (changes requested):
Biggest issue so far is that this PR makes zen mode very jumpy:


https://github.com/user-attachments/assets/252d332d-1dcc-47ff-a5d4-63b207743ea7

NadAlaba:
> 7 month pr :X thanks and sorry

tbh most of the delay was mine, but it's good to see it done

PR #5924, opened 2024-09-28
feat(lang): add Bengali language (@SubhodeepRoy2)

SubhodeepRoy2:
Added Bengali as a language using the English alphabet (transliteration). Although Bengali was already included, it was in the Bengali script. This contribution adds Bengali words written in the English alphabet.
[feat(languages): add bengali (http://github.com/SubhodeepRoy2) ]


rammba:
Does it make sense to put it in the same language group as already existing language? I think we have similar case for Serbian, where we use both cyrillic and latin versions.

SubhodeepRoy2:
That can also be done. or it could be added as a separate language variant. Both works, what's your suggestion?

rammba:
I would say it's better to group them, since they are related. I see that also sanskrit language is in the [same group](https://github.com/monkeytypegame/monkeytype/blob/fa58e090db77fb7dfb98d3db6315bf94f4e8e7b9/frontend/static/languages/_groups.json#L421C20-L421C28) as sanskrit_roman which is similar case as yours, right?

SubhodeepRoy2:
hey @rammba can you help me with the errors that I'm getting here?
image

rammba:
I think you should rename title of this PR to something like: `feat(bengali): add Bengali language (@SubhodeepRoy2)`.

frontend/static/languages/_groups.json
```diff
@@ -609,5 +609,9 @@
       "code_cobol",
       "code_common_lisp"
     ]
+  },
+  {
+    "name": "bengali",
+    "languages": ["bengali"]
   }
 ]
```
Miodec, line 617: This file (and the _list file) is meant to be ordered roughly by popularity. Please update the position of bengali.

Miodec (changes requested):

PR #5925, opened 2024-09-28
impr(serbian): add more attributes to Serbian files (@rammba)

rammba:
I've specified `rightToLeft`, `ligatures` and `bcp47` attributes in all Serbian files just to make them more understandable.
I suppose that `rightToLeft` and `ligatures` are false by default. Please let me know if I'm wrong.


Miodec:
Yeah those values are false by default but its fine.

PR #5926, opened 2024-09-28
impr(dockerfile): add support for Dockerfile language (@rammba)

rammba:
I've added support for Dockerfile "language" even if it's not the language actually. All words added are listed in the [official docs](https://docs.docker.com/reference/dockerfile/#overview). I've put it to the other group of languages.

P.S. File name is `docker_file.json`, because when I name it `dockerfile.json`, my VS Code detects it as an actual dockerfile. Let me know if name should be changed.


Miodec:
Thanks

PR #5930, opened 2024-09-28
impr: monkey now types with the same hand as the user (@ShaneBerhoff)

ShaneBerhoff:
### Description

The typing monkey from the Toggle Monkey commandline feature would originally just cycle through left and right hand presses. Now it creates a handmap from the user's layout and gets the specific key down and up events. This way the hand the monkey types with can approximately follow the hand the user types with (spacebar is set to both hands). It also gets layout updates so it will still work even if the user changes their layout.

### Checks

- [ ] Adding quotes?


Miodec (commented):
This seems quite simmilar to opposite shift tracking. I think the two features could be combined somehow to have one central place for this.

ShaneBerhoff:
I agree. There are two main key conversions happening in both places: converting key based on layout and converting from key to keyboard side. I think this could be combined into a key-converter.ts util file which can then centralize any sort of key conversions for this case and in the future. How does this sound?

I am going to start working on this.

ShaneBerhoff:
I like that idea. I think it would optimize and centralize a lot of what I have been seeing in the codebase with the way layouts are used. It would also make it easier for people to build out layout based or conversion based features in the future. 

I also just found a bug with my original feature implementation based on the way I did it. Currently for users with alternate keyboard layouts it only works if they have the layout emulator set to their keyboard layout.

I am going fix the bug while combining the shared parts of the features and look into doing what you recommended with layouts overall because I really like that idea.

Miodec:
> I also just found a bug with my original feature implementation based on the way I did it. Currently for users with alternate keyboard layouts it only works if they have the layout emulator set to their keyboard layout.

For that just make sure you base the monkey hands on the key locations, not key characters. For example the key in position of the left index finger will always be `KeyF`, no matter what layout youre using. 

(unless you are remapping the keys at the keyboard level with something like QMK, then the physical locations will change, but at that point we can't handle that because we don't know what the taget locations are, in fact layout emulator breaks if you remap at the keyboard level, so don't worry about that)

Miodec:
For B, Y and Space:

Monkey hands: either alternate or choose randomly. 
Opposite shift: ignore those keys (any shift can shift those characters, just like the current implementation).

Miodec:
Thanks

PR #4497, opened 2023-08-01
added middlemak and middlemak-nh

krig:
### Description

Adds the Middlemak and Middlemak-NH keyboard layouts:

Middlemak: https://github.com/KeyboardLayout2/Middlemak
Middlemak-NH: https://github.com/KeyboardLayout2/Middlemak-NH

Closes #

PR #5932, opened 2024-09-30
fix(layout): typo in middlemak and middlemak-nh (@toast-riot)

toast-riot:
### Description

Fixes the `.>` key instead being a `,>` on the middlemak and middlemak-nh keyboard layouts.

Issue introduced in #4497


Miodec:
Thanks

PR #5957, opened 2024-10-14
feat(layout): add haruka (@inscapist)

inscapist:
### Description

Adds haruka layout. 

Featuring same hand j/k (on the right because combinations like lsft+lctl+j/k are common)

- [cyanophage playground](https://cyanophage.github.io/playground.html?lan=english&layout=quopzvfdlm-ieanbgstrc%27%2C%2F.h%3Bjykxw%5C%5E&mode=ansi)


Miodec:
Thanks

PR #5959, opened 2024-10-14
fix: allow only one funbox changing capitalisation (@fehmer)

fehmer:
Fixes funboxes like  `capitals` and `instant messaging` not working at the same time.


Miodec:
Thanks

PR #5960, opened 2024-10-14
feat: show tape mode used on result screen (@fehmer)

fehmer:
Show the tape mode used in the test type on the result screen.

This was requested on discord. I am not sure about this. On the one hand we do include the funboxes in the test mode. Funboxes like `plus three` or `crt` only change the appeareance and are shown. Tape mode could've been implemented as a funbox and would show up on the result screen.
On the other hand this is a regular setting and other settings like `highlight` work similar and are not shown on the result page.

### DIFF
--- a/frontend/src/ts/test/result.ts
+++ b/frontend/src/ts/test/result.ts
@@ -158,7 +158,7 @@ async function updateGraph(): Promise<void> {
         borderRadius: 3,
         position: "start",
         display: true,
-        content: `${content}`,
+        content: content,
       },
     });
   }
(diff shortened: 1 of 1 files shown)


Miodec:
The result page only includes things that change the behavior of the test (you could argue that plus n funboxes make the test harder) - it doesnt include anything that is just changing the visuals of the test like highlight mode, smooth caret or font family/size, so i think tape mode shouldnt be included (crt is a unique case as it was made into a funbox to work with every theme). Thank you though.

PR #5976, opened 2024-10-21
impr(quotes): Add Transdermal Celebration by Ween to english.json (@CoolerFred)

CoolerFred:
Add the song Transdermal Celebration by the band Ween as a quote. Specifically part from the first verse that goes "Transdermal celebration, caused a slight mutation in the rift. It toppled down a nation, and left the people running for the hills."


Miodec:
Thanks

PR #5991, opened 2024-10-28
feat(input): add 'code unindent on backspace' for code languages (@notTamion)

notTamion:
When using a code language that has tab characters as indentation and you backspace into the indentation, the entire indentation will now be deleted and you will be set back to the end of the previous line. This is how all Jetbrains ides and also a lot of other ides do it. We could add a config option for this if someone wants one (i wasn't really able to come up with a good name for one).


Miodec:
Not sure i like this change, I've never used a code editor that does what you described.

notTamion:
Hmm i mostly use Jetbrains Ides (i assume you use vsc then?) I can add a setting to change this then

Miodec:
I would change the setting to "code unindent on backspace" and have the settings be on or off (so, a boolean). The setting description could also use some work so that its easier to understand. The default value should be off.

Miodec:
Doesnt work in this case, the tab character is deleted but input doesn't go back to the previous line, caret doesn't move.

image

notTamion:
fixed

notTamion:
Woops. just noticed i didn‘t pull your changes to the message. could you push that commit again or tell me what the new message was so i can change it

Miodec:
Thats why you generally wanna avoid force pushing ;)

Miodec (approved):

Miodec:
Thanks

notTamion:
> Thats why you generally wanna avoid force pushing ;)

yeah i usually prefer rebasing over merging but i sometimes forget to pull the feature branch

PR #5993, opened 2024-10-28
fix: remove space at beginning of custom text lines

notTamion:
when creating a multi line text in the custom text editor and then reopening the editor all lines after the first will have a space at the start, this is because .join adds a space regardless of whether there is a newline character


Miodec:
Thanks

PR #6006, opened 2024-11-04
impr(read ahead): show words after correcting typo with backspace (@notTamion)

notTamion:
### Description
Makes it easier to recover when not correctly remembering a word

### DIFF
--- a/backend/src/constants/funbox-list.ts
+++ b/backend/src/constants/funbox-list.ts
@@ -205,7 +205,7 @@ const FunboxList: FunboxMetadata[] = [
     frontendForcedConfig: {
       highlightMode: ["letter", "off"],
     },
-    frontendFunctions: ["applyCSS", "rememberSettings"],
+    frontendFunctions: ["applyCSS", "rememberSettings", "handleKeydown"],
     name: "read_ahead_easy",
   },
   {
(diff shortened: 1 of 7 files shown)


Miodec:
The issue i see with this is that if you're typing fast, you will get some very annoying flashing of the words if you make mistakes but keep typing. Also, the words shouldnt be shown if blind mode is enabled.

So yeah, im not sure about it, the flashing could be very annoying.

notTamion:
Yeah i guess this depends on your typing style. I  personally prefer to go back and fix my errors. Will come back to this when im home

notTamion:
Now it only shows the words if you backspace

Miodec:
I wonder if it would be better to hide the words again on any keypress rather than only on space. What do you think

notTamion:
i feel like hiding them on any key press leaves a bit too little time. considering when you error it will already show you the correct letter in red (and then most people will just do a backspace + that letter in one motion which leaves practically no time for reading anything)

Miodec:
The one issue i found is that you can space to the next word, press backspace (deleting nothing) and then the words become visible. Thats why i was thinking about the 'hiding on every keypress' thing.

notTamion:
fixed

Miodec:
Thanks

PR #6017, opened 2024-11-07
impr: allow multiple funboxes with css (@notTamion)

notTamion:
### Description
Allows enabling multiple funboxes that have a stylesheet. Which funboxes should be deemed compatible with each other will need some discussion


Miodec:
You will need to go through the list of funboxes and check if they alter the same element. For example, now, you can enable mirror and upside down at the same time, but because they edit the same dom element, over overrides the other.

notTamion:
there might be more combinations that i am not sure whether we should allow, for example: crt and spaceballs (having both enabled with still tilt the screen but no longer display the background image from spaceballs)

Miodec:
Any combinations which change the behavior like that should be avoided.

notTamion:
Should have added validation for all incompatible funboxes now. ready for review

Miodec:
Funbox refactor is now on master. Have a look and let me know if you need help.

notTamion:
should be ported over properly i didn't really extensively test it as i am a bit handicapped right now cause my gpu decided to die yesterday so i can only use 1 monitor for the moment.

fehmer (dismissed):
lgtm

Miodec (dismissed):
Thanks, sorry it took a while to merge this.

notTamion:
Thanks for taking the time to review and add finishing touches.

PR #6027, opened 2024-11-10
impr(nvm): use faster fnm instead of nvm (@maxhu08)

maxhu08:
### Description

Change `docs/CONTRIBUTING_ADVANCED.md` to recommend using fnm instead of nvm since it's faster.


Miodec:
I never heard of fnm before and since switching node versions often, I would rather stick with a well known tool. Sorry.

PR #6031, opened 2024-11-11
impr(quotes): add English quote (@marcogrivol)

MarcoGrivol:
Added a Walt Whitman quote. 
For reference: [Poetry Foundation](https://www.poetryfoundation.org/poems/45479/when-i-heard-the-learnd-astronomer).


Miodec:
This whole quote just seems to be writting in broken english.

PR #6033, opened 2024-11-12
fix(language): removed non-french from the french language files (@theiereman)

theiereman:
### Description

I was doing some typing tests and I noticed the "music" word in the french languages pack. I checked in a dictionnary and that word doesn't exist so I removed it from every package that had that word in it. The only correct existing word is "musique".


Miodec:
Thanks

PR #6035, opened 2024-11-13
fix(language): remove catalan opening punctuation (@Perseus333)

Perseus333:
### Description

Previously, Spanish exclusive punctuation was included in the Catalan language. In Catalan "¿" and "¡" don't exist. I just made it so that it adds the "¿" and "¡" if the current language is set to spanish.

### Checks

- [ ] Adding quotes?
- [ x Make sure the PR title follows the Conventional Commits standard. (https://www.conventionalcommits.org for more info)

Closes #


Miodec (approved):
Thanks

frontend/src/ts/test/words-generator.ts
```diff
@@ -57,10 +57,14 @@ export async function punctuateWord(
     if (currentLanguage === "spanish" || currentLanguage === "catalan") {
```
Miodec, line 57: Wait, so, if you check if the language is spanish in both cases inside this if, why not just remove this or condition?

Perseus333: How would you remove the or condition? You still need the "!" and "?" in Catalan. How would you optimize this? If you removed the or, then you would need to have a separate if statement with duplicate logic. 

Or maybe there is something evident that I'm not seeing (very possible). Let me know.

Miodec: You can just remove the `currentLanguage === "catalan"` check on that line and one line `74`, the rest is already handled.

Perseus333: Now it should be fixed. Thanks for the help, @Miodec!

Miodec (approved):
Thanks

PR #6036, opened 2024-11-13
Fix Quick restart key tip

mrtianjin829:
### Description
People will get confused if they see "off" as the quick restart key bind, so if the quick restart keybind is not bound, instead it will show `(No key binded)` instead.

### DIFF
--- a/frontend/src/ts/ui.ts
+++ b/frontend/src/ts/ui.ts
@@ -47,7 +47,7 @@ function updateKeytips(): void {
 
   const commandKey = Config.quickRestart === "esc" ? "tab" : "esc";
   $("footer .keyTips").html(`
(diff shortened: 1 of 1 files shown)


Miodec:
Looks like an oversight by the last editor of this code. It should say "tab + enter" instead of "off" or "no key bound". Could you modify your PR?

mrtianjin829:
> Looks like an oversight by the last editor of this code. It should say "tab + enter" instead of "off" or "no key bound". Could you modify your PR?

is it okay now?

Miodec:
Made some minor fixes and reran prettier. Thanks for your PR.

mrtianjin829:
Welcome 👍

PR #6096, opened 2024-12-06
fix(layout): changes in the header, remove unnecessary attributes (@Mineclinee)

Mineclinee:
### Checks

- [ ] Adding quotes?

Closes #
- Changing the nesting of the div tag in the header from div to span
image
image
- Removing an unnecessary empty tag in the header (added a related css rule to change the layout breakage)
- Removing unnecessary attributes in the textarea tag that are an error and are complained about by w3c validator

### DIFF
--- a/frontend/src/404.html
+++ b/frontend/src/404.html
@@ -53,7 +53,7 @@
             </svg>
           </div>
           <h1 class="text">
-            <div class="top">monkey see</div>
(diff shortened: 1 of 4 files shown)


Miodec:
Not sure about this, seems like changes for the sake of making a change, instead of fixing an actual issue or making an improvement of some sort.

PR #6100, opened 2024-12-08
impr: center mini stats when using tape mode (@notTamion)

notTamion:
### Description
old:
image

new:
image

### DIFF
--- a/frontend/src/ts/config.ts
+++ b/frontend/src/ts/config.ts
@@ -907,8 +907,11 @@ export function setTapeMode(
     return false;
   }
 
-  if (mode !== "off" && config.showAllLines) {
+  if (mode !== "off") {
     setShowAllLines(false, true);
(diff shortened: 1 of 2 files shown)


Miodec:
This will overlap the test modes notice when focus mode is disabled.

notTamion:
focus mode? how do i enable that? i can't find it in the settings.

Miodec:
Focus mode just refers to the ui hiding itself when you start typing.

notTamion:
yeah i thought that might be it but i don't know what its actually called.

Miodec (changes requested):
Instead of modifying the dom from the config file, please move this change into `test-ui.ts` - at the end of the file there is a `ConfigEvent` subscriber - make the css change there.

Miodec (approved):
Thanks

PR #6102, opened 2024-12-09
feat(themes): added witchthorne theme (NotMageLock)

NotMageLock:
added new theme, witchthorne
IMG_5666
IMG_5664


fehmer:
Very similar to `joker`

PR #6104, opened 2024-12-10
fix: some funboxes not working if reduced motion is preferred (@fehmer)

fehmer:


packages/funbox/src/list.ts
```diff
@@ -37,15 +37,15 @@ const list: Record<FunboxName, FunboxMetadata> = {
     description: "I think I'm gonna be sick.",
     canGetPb: true,
     difficultyLevel: 2,
-    properties: ["hasCssFile"],
+    properties: ["hasCssFile", "forceAnimation"],
```
Miodec, line 40: Rename to `ignoreReducedMotion`?

Miodec (approved):
Thanks

PR #6105, opened 2024-12-10
fix: remove tailing space and non-break space from various languages (@fehmer)

fehmer:
Reported by dongseonga on discord.


Miodec (commented):
Thanks

PR #6106, opened 2024-12-10
Update spanish.json

Libz24:
impr(quotes): add spanish quotes (@Libz24 )

### Description

More philosophy and poetry quotes for spanish

### Checks

- [ ] Adding quotes?

Quotes added

People of different trades or classes, who have nothing in common, living near each other by chance, are combined into an artificial group that has to be represented by a delegate.

The more love you give, the better you are (this one is not 100% correct, in spanish saying "Mejor estás" can be interpreted as something like saying "the better you feel and the better you see yourself")

The poor had to stand and the rich had boxes. But there was something that united the banker and the poor: the hatred of beauty.

Nobody understood that phrase and it causes me some difficulty explaining what Teresa meant when she compared a nudist beach to the Russian invasion.

A look from the sewer can be a vision of the world, rebellion consists of looking at a rose until your eyes are pulverized.  (this one is poetry so the meaning and the depth cannot be 100% correctly translated)

 @Libz24

Closes #

 There´s not that many quotes for spanish. An hour of playing is enough to start getting the same ones over and over

### DIFF
--- a/frontend/static/quotes/spanish.json
+++ b/frontend/static/quotes/spanish.json
@@ -1140,6 +1140,41 @@
       "source": "Enrique de Mesa - La alegre carreta",
       "length": 81,
       "id": 198
-    }
+    },
+    {
+    "text": "Personas de distintos oficios o clases, que no tienen nada en común, viviendo cerca de sí por casualidad, son combinadas en un grupo artificial qe tiene que ser representado por un delegado",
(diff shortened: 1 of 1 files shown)


Miodec:
I only really like the second quote, please remove the rest.

Miodec:
Closing due to inactivity

Issue #5881, opened 2024-09-13
Replace JSON.parse with  parseJsonWithSchema


fehmer:
Replace `JSON.parse` with `Misc.parseJsonWithSchema` for type safety.

Example:
https://github.com/monkeytypegame/monkeytype/blob/2a0a5b408cc38a678328c2ff44e281416ebb9c4d/frontend/src/ts/utils/url-handler.ts#L68-L92

Miodec:
I think before we do this we should extract common util functions (like this one) to a package.

rajneesh-18:
It should be go on the discussion.

dev-mohit06:
Hi @fehmer, Could you assign me that issue.

fehmer:
@dev-mohit06  I'd say just work on it and create a PR

dev-mohit06:
Hey @fehmer, I’d prefer to avoid duplication of effort, so could you assign this issue to me if possible?

dev-mohit06:
@fehmer - Could you please indicate which files in the frontend need JSON.parse replacements? This will help ensure I cover all necessary changes.

fehmer:
https://github.com/search?q=repo%3Amonkeytypegame%2Fmonkeytype+JSON.parse+language%3ATypeScript+path%3A%2F%5Efrontend%5C%2Fsrc%5C%2Fts%5C%2F%2F&type=code

NadAlaba:
Isn't this done in #6207?

Miodec:
> Isn't this done in [#6207](https://github.com/monkeytypegame/monkeytype/pull/6207)?

Backend is not using `parseJsonWithSchema` yet

dev-mohit06:
@Miodec I will be happy to do that. I will start now onwards.

byseif21:
> > Isn't this done in [#6207](https://github.com/monkeytypegame/monkeytype/pull/6207)?
> 
> Backend is not using `parseJsonWithSchema` yet


been a while, should I give it a try?

dev-mohit06:
@byseif21 broo I already finished 😅

byseif21:
> [@byseif21](https://github.com/byseif21) broo I already finished 😅

[...]

PR #6109, opened 2024-12-12
refactor: move parseJsonWithSchema to utils package (@fehmer)

fehmer:
- extract parseJsonWithSchema for #5881
- add one test in `funbox` package because eslint was complaining that there are no tests


packages/funbox/__test__/utils.spec.ts
fehmer, line 1: After adding the `json.spec.ts` file in `util` package eslint was complaining that the `packages/funbox/__test__/tsconfig.json` cannot find any files to include.

Instead of supressing this I added a test.

packages/util/src/json.ts
```diff
@@ -0,0 +1,27 @@
...
+  } catch (error) {
+    if (error instanceof ZodError) {
+      throw new Error(error.issues.map(prettyErrorMessage).join("\n"));
+    } else {
+      throw error;
+    }
+  }
+}
+
+function prettyErrorMessage(issue: ZodIssue | undefined): string {
```
fehmer, line 23: previously only the `error.errors` was used to compile the error message. It only contained the validation errors like `Expected boolean, received string` but not the path so it was unclear what exactly is wrong in the json. 

Adding the full path like we do in the error handler on the backend api.

thejus-r (approved):

Miodec (approved):
Thanks

PR #6133, opened 2024-12-24
feat(result page): open daily leaderboard when clicking the daily leaderboard rank (@theiereman)

theiereman:
### Description

Clicking on the rank (after getting a better rank in the daily leaderboard) will automatically open the leaderboard modal on the corresponding language screen

gifgit

Honestly, I mostly added this for me because I like to go to the daily leaderboard after I did a good run and it was annoying me to have to open it + change the language to french + change to daily.


Miodec:
Thanks

PR #6155, opened 2025-01-06
fix: funbox plus x not working with custom text using pipe delimiter (@fehmer)

### LINKED ISSUE #6151: Plus three issues with pipe delimiter

fehmer:
Fixes #6151

### DIFF
--- a/packages/funbox/package.json
+++ b/packages/funbox/package.json
@@ -9,7 +9,6 @@
     "lint": "eslint \"./**/*.ts\""
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


packages/funbox/package.json
```diff
@@ -20,6 +19,9 @@
     "typescript": "5.5.4",
     "vitest": "2.0.5"
   },
+  "dependencies": {
+    "@monkeytype/util": "workspace:*"
```
fehmer, line 23: we need utils during runtime

Miodec (approved):
Thanks

PR #6179, opened 2025-01-13
fix(funbox): punctuation disabled when simon_says is enabled (@ippatev)

### LINKED ISSUE #6152: simon says doesn't show shifts

- [X] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

gtbot2007

### Account config

_No response_

### Current Behavior

When using simon says, shifts never show on the keyboard

### Expected Behavior

It should show shifts

### Steps To Reproduce

1. turn on simon says
2. choose a code language
3. when it asks to type < or > it shows as , and . which is the unshifted version

### Environment

OS: Windows 10
Browser: Google Chrome

### Anything else?

_No response_

ippatev:
### Description

Disabled the punctuation button and punctuation itself when using simon says funbox

Closes #6152

### DIFF
--- a/frontend/src/ts/test/funbox/funbox.ts
+++ b/frontend/src/ts/test/funbox/funbox.ts
@@ -57,6 +57,13 @@ export function toggleFunbox(funbox: "none" | FunboxName): boolean {
     get(funbox as FunboxName).functions?.applyGlobalCSS?.();
   }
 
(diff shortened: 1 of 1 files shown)


Miodec:
Sorry - this isnt really a fix for this issue - i would ideally like to modify the keymap so it can show shifts if needed. Thank you though.

PR #6206, opened 2025-01-23
refactor: use vendor css from npm package (@fehmer)

fehmer:
This PR removes the remaining "vendor" css files copied into `static/css` and pick them from npm packages instead.

For the slimselect the css should match the version of the js. Since we use the slimselect `js` file from the npm package we should use the `css` file from the package as well.

For the balloon.css the file was copied into the project. Adding it as a dependency makes sure we know the version of it and we don't modify it in the project leading to problems when we try to update it later.

Similar to the vendor `js` files (js dependencies from packages) extract the vendor `css` files into a separate `css` file as well.


Miodec (approved):
Thanks

PR #6207, opened 2025-01-23
refactor: Replace JSON.parse with parseJsonWithSchema (@dev-mohit06)

dev-mohit06:
## Description
Replaces raw JSON parsing with schema-based validation across frontend TypeScript files to improve type safety and error handling.

### Scope of Changes
- Updated JSON parsing in:
  - `account.ts`
  - `import-export-settings.ts`
  - `analytics-controller.ts`
  - `local-storage-with-schema.ts`
  - `url-handler.ts`
  - `commandline/lists.ts`
  - `test/wikipedia.ts`

- Added schema in `test/custom-text.ts`:
  ```typescript
  export const customTextDataSchema = z.object({
    text: z.array(z.string()),
    mode: CustomTextModeSchema,
    limit: z.object({ 
      value: z.number(), 
      mode: CustomTextLimitModeSchema 
    }),
    pipeDelimiter: z.boolean(),
  });
  ```

### Benefits
- Enhanced runtime type safety
- More robust error handling
- Consistent JSON parsing approach

### Checks
- [x] Follows Conventional Commits
- [x] Includes GitHub username
- [ ] Adding quotes? (N/A)
- [ ] Adding language/theme? (N/A)

PR #6229, opened 2025-02-03
fix: fix error handling in parseWithSchema (@fehmer)

fehmer:
`instanceof ZodError` is not working if the code is packaged as a module for unknown reason. Found while adding tests for the url-handler in #6207.


fehmer:
Depends on #6207 to be merged first.

Miodec (approved):
Thanks

PR #6230, opened 2025-02-04
fix(english contractions): Remove objectionable words (@AnnaBuchholz)

AnnaBuchholz:
### Description

Removed two objectionable words from english_contractions word list.

### Checks

- [x] Make sure the PR title follows the Conventional Commits standard.(https://www.conventionalcommits.org for more info)

Closes #


AnnaBuchholz:
I clearly don't have the best Git proficiency so I don't know why it has a bunch of my old merges and whatnot from other PRs on this PR. I'd be curious to hear why if anyone knows.

AnnaBuchholz:
Also, is there any particular standard for which words are added to this wordlist? It only has ~200 words and a lot of them have apostrophes (for self-explanatory reasons) and thus don't show up if punctuation is set to false. Also, because it's such a small word list, a good portion of the words have multiple apostrophes (such as fo'c'sle and the quite outrageous y'all'd'n't've). I personally think such words have an outsized impact on this word list and perhaps we could add other contractions to counteract their effect.

Miodec:
I think the two words youre removing in this PR are fine to keep. As for the punctuation issues, I think ill add a special case to the word generation so that these are not removed.

As for the weird words like `fo'c'sle`, we should probably remove them and add some more common ones like you said.

Ill close this PR for now since the words are fine. Feel free to open a new one if you wish to replace the weird words.

PR #6260, opened 2025-02-12
feat(layout): colemak_wide keyboard layout (@papersacculos)

papersaccul:
It seems to be a pretty popular mod for colemak without additional mods, which is easy to switch to from vanilla colemak, but for some reason nobody added it ;/

https://colemakmods.github.io/ergonomic-mods/wide.html


Miodec (approved):
Thanks

PR #6268, opened 2025-02-16
feat: Self hosting offline with account support using firebase emulator (@ergoxi)

ergoxi:
### Description

Hi, just thought I would circulate this for anyone searching for a way to run locally with account support without having to set up a real firebase project. If this has a chance of getting merged I can clean it up, update the docs, and surround the emulator code snippets in conditionals. 

### Checks

- [ ] Adding quotes?
- [ ] Make sure to include your GitHub username prefixed with @ inside parentheses at the end of the PR title.

Closes #

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -10,13 +10,11 @@ MONKEYTYPE_FRONTENDURL=http://localhost:8080
 #url of the backend server, this must be accessible by your clients browser
 MONKEYTYPE_BACKENDURL=http://localhost:5005
 
-# MODE=dev
-
 # below config is only needed, if you need user accounts
 # firebase config
 FIREBASE_APIKEY=AAAAAAAA
 FIREBASE_AUTHDOMAIN=monkeytype-00000.firebaseapp.com
-FIREBASE_PROJECTID=monkeytype
+FIREBASE_PROJECTID=local-project
 FIREBASE_STORAGEBUCKET=monkeytype-00000.appspot.com
 FIREBASE_MESSAGINGSENDERID=90000000000
 FIREBASE_APPID=1:90000000000:web:000000000000


fehmer:
Hi @ergoxi ,

this would be an interesting feature. 

From the current state of the PR it is not easy to judge how many changes would be needed and how to configure it. Can you please create a PR with the actual changes including the updated [documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/SELF_HOSTING.md)?

Some things to keep in mind:
- We don't expect users to clone the monkeytype respository for self-hosting. Maybe you can use an existing image for the firebase-emulator docker container instead? This would eliminate the `Dockerfile` and `firebase.json` files.
- If possible define every variable only once in the `.env` file.
- It should work on the network as well as on localhost.
- it stil has to work with hosted firebase, obviously

fehmer:
@ergoxi are you in the discord by chance? We could discuss this there faster.

docker/frontend/updateConfig.sh
```diff
@@ -7,7 +7,7 @@ sed -i "s/###FIREBASE_PROJECTID###/${FIREBASE_PROJECTID}/g" js/firebase.*.js
 sed -i "s/###FIREBASE_STORAGEBUCKET###/${FIREBASE_STORAGEBUCKET}/g" js/firebase.*.js
 sed -i "s/###FIREBASE_MESSAGINGSENDERID###/${FIREBASE_MESSAGINGSENDERID}/g" js/firebase.*.js
 sed -i "s/###FIREBASE_APPID###/${FIREBASE_APPID}/g" js/firebase.*.js
-
+sed -i "s/###FIREBASE_AUTH_EMULATOR_HOST_FRONTEND###/${FIREBASE_AUTH_EMULATOR_HOST_FRONTEND}/g" js/firebase.*.js
```
fehmer, line 10: ```sed -i "s/###FIREBASE_AUTH_EMULATOR_HOST_FRONTEND###/${FIREBASE_AUTH_EMULATOR_HOST_FRONTEND//\//\\/}/g" js/firebase.*.js```

fehmer:
I am not sure we should offer this as an officially supported feature for self hosting

1. The firebase emulator exports the users and their credentials in plain text which is not good
2. The website shows a big warning on the bottom of the page 
image
3. There seems to be no authentication on the firebase auth emulator endpoints. So if you exposing them on your self hosted system everybody can read and delete your user.

PR #6272, opened 2025-02-16
refactor: use validation on email update modal (@fehmer)

fehmer:
alternative approach for 535700f0c5861171bb4bad5538c8f699f154ca3f

### DIFF
--- a/frontend/src/ts/modals/simple-modals.ts
+++ b/frontend/src/ts/modals/simple-modals.ts
@@ -35,7 +35,7 @@ import {
 } from "../utils/simple-modal";
 import { ShowOptions } from "../utils/animated-modal";
 import { GenerateDataRequest } from "@monkeytype/contracts/dev";
-import { UserNameSchema } from "@monkeytype/contracts/users";
+import { UserEmailSchema, UserNameSchema } from "@monkeytype/contracts/users";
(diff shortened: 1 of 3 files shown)


frontend/src/ts/modals/simple-modals.ts
```diff
@@ -230,11 +230,21 @@ list.updateEmail = new SimpleModal({
...
     {
       type: "text",
       placeholder: "Confirm new email",
       initVal: "",
+      validation: {
+        schema: UserEmailSchema,
+        isValid: async (confirm) =>
+          confirm ===
+            (document.getElementById("updateEmail_1") as HTMLInputElement)
+              .value || "not matching email",
```
Miodec, line 246: I would rename `confirm` to `currentValue`, and im thinking if we could pass `this` to the `isValid` function so that we can do `this.inputs[1] === currentValue`

Miodec (changes requested):

frontend/src/ts/modals/simple-modals.ts
fehmer: Check https://github.com/monkeytypegame/monkeytype/pull/6272/commits/49658af48b8d0c76a0d405806c53f78c299e83fb

Miodec:
Thanks

PR #6285, opened 2025-02-20
feat: leaderboard scroll to users position when using the user button (@fehmer)

fehmer:
- scroll to users position when using the user button
- add scroll to top button on leaderboard page


Miodec (approved):
Thanks

PR #6290, opened 2025-02-21
fix: remove user from XP leaderboard if opt-out, reset or deleted (@fehmer)

fehmer:
Ensure user is removed from the weekly XP leaderboards if they either `opt-out`, `reset` their account or `delete` their account.


Miodec:
Thanks

PR #6291, opened 2025-02-21
fix: leaderboard keeping pagination after switching language (@fehmer)

fehmer:
Fixes daily leaderboard staying on a selected page after switching languages. If the new language has less pages the user was stuck as none of the navigation buttons were active.

The easy fix is to always start at the first page after switching languages. We do the same when switching leaderboards (e.g. daily to xp weekly)


Miodec (approved):
Thanks

PR #6292, opened 2025-02-21
fix: daily leaderboard worker not giving XP to users (@fehmer)

fehmer:


backend/src/workers/later-worker.ts
```diff
@@ -18,7 +18,7 @@ import { WeeklyXpLeaderboard } from "../services/weekly-xp-leaderboard";
 import { MonkeyMail } from "@monkeytype/contracts/schemas/users";
 import { mapRange } from "@monkeytype/util/numbers";
 
-async function handleDailyLeaderboardResults(
+export async function handleDailyLeaderboardResults(
```
fehmer, line 21: todo revert

PR #6300, opened 2025-02-22
fix: Remove invalid words from Greek lang (@sledgehammer999)

sledgehammer999:
### Description

These words aren't valid greek words.


sledgehammer999:
An unfortunate sideeffect of this PR is that the wordlists will have a smaller size than their name indicates.

Miodec:
> An unfortunate sideeffect of this PR is that the wordlists will have a smaller size than their name indicates.

Its ok if its missing a few. 

You could also add a few more words if you want to?

sledgehammer999:
>You could also add a few more words if you want to?

Maybe in another PR?

Is there any rule on what goes into the 1k/5k/10k wordlists? Are they based on the frequency of usage of each word or something?

Miodec:
> > You could also add a few more words if you want to?
> 
> Maybe in another PR?
> 
> Is there any rule on what goes into the 1k/5k/10k wordlists? Are they based on the frequency of usage of each word or something?

Same PR is fine. Words should be based on frequency, so 1k would include around 1000 most common words in a given language.

sledgehammer999:
>Same PR is fine.

You misunderstood me. I meant if the new words can be expected in a new PR as to not stall this one.

>Words should be based on frequency, so 1k would include around 1000 most common words in a given language.

What does one discover common words in a given language? Is there any common practice?

Miodec:
> > Same PR is fine.
> 
> You misunderstood me. I meant if the new words can be expected in a new PR as to not stall this one.
> 

Ok. You can open a new PR if you wish. The number of words isnt that bad. 

> > Words should be based on frequency, so 1k would include around 1000 most common words in a given language.
> 
> How does one discover common words in a given language? Is there any common practice?

Pretty much just googling around.

PR #6302, opened 2025-02-23
fix: remove deprecated galite layout (@almk-dev)

almk-dev:
### Description
After feedback from the AKL community, I have reverted many of the changes made in Galite to the point where it is not meaningfully different from Gallium. As such, Galite is deprecated and no longer supported/recommended. Apologies for the inconvenience.


Miodec:
Thanks

PR #6314, opened 2025-02-26
fix: handle legacy test settings urls (@fehmer)

fehmer:
Some people have old links with test settings we should support.


Miodec:
Thanks

PR #6317, opened 2025-02-27
fix(docker): fix configuration endpoint, add configuration export (@fehmer)

fehmer:
- fixes /configure endpoint not working in docker image because of the missing `private` directory
- adds `export configuration` button to api server configuration


Miodec (approved):
Thanks

PR #6319, opened 2025-02-27
fix: Slow and Steady Challenge Authentication (@AnnaBuchholz)

AnnaBuchholz:
### Description

I changed the Slow and Steady authentication to check if "liveSpeedStyle" is "off" rather than checking the if the legacy "showLiveWpm" was "false". This allows people to actually get the Slow and Steady challenge role automatically instead of getting a failure error even when they had liveSpeedStyle off.

### Checks

Closes #


AnnaBuchholz:
I think I just needed to update the legacy naming to fix the challenge authentication, but if I missed other places I need to change, please let me know.

fehmer:
can you verify the fix is working locally?

AnnaBuchholz:
I just tried doing the Slow and Steady challenge on localhost, but at the end there were no notifications at all (of challenge success or failure). It seems like I might need to be logged in to test the challenge authentication but I'm not sure how to log in as a user to test the challenge on the test site. Any idea what I could do to test it?

AnnaBuchholz:
Alright, I posted some proof in the Discord. If there's anything else I need to do, please let me know.

Miodec (approved):
Thanks

PR #6321, opened 2025-02-27
fix(language): Remove nekudos from Yiddish file (@isaacweber613)

isaacweber613:


LazerDoubleU:
I agree that we don't need nekudos when typing in Hebrew most of time, maybe it should be a setting to turn it on or off.

Miodec:
Instead of removing these from the word list, you should edit the `frontend/src/ts/test/lazy-mode.ts` file and add a ll the accents and their non accented versions. Also, remove `noLazyMode` from the list file.

isaacweber613:
> Instead of removing these from the word list, you should edit the `frontend/src/ts/test/lazy-mode.ts` file and add a ll the accents and their non accented versions. Also, remove `noLazyMode` from the list file.

Hey, thanks for the feedback!
I get why other languages handle diacritics in lazy-mode.ts, but Yiddish is different. Yiddish is almost never written with nekudos in real life—they're mainly used for educational or religious texts, not everyday writing. Keeping them in the JSON and stripping them dynamically just adds extra processing for no reason.
The best solution is to remove the nekudos from the word list entirely, since that’s how Yiddish is actually written. This keeps things simple, accurate, and closer to what native speakers expect.
besides I'm not good at coding and don't know how to do this.

Miodec:
> > Instead of removing these from the word list, you should edit the `frontend/src/ts/test/lazy-mode.ts` file and add a ll the accents and their non accented versions. Also, remove `noLazyMode` from the list file.
> 
> Hey, thanks for the feedback! I get why other languages handle diacritics in lazy-mode.ts, but Yiddish is different. Yiddish is almost never written with nekudos in real life—they're mainly used for educational or religious texts, not everyday writing. Keeping them in the JSON and stripping them dynamically just adds extra processing for no reason. The best solution is to remove the nekudos from the word list entirely, since that’s how Yiddish is actually written. This keeps things simple, accurate, and closer to what native speakers expect. besides I'm not good at coding and don't know how to do this.

I would keep it in lazy mode to give the user an option. Extra processing needed is not really a consideration. If you prepare a list of character pairs, i can add it myself.

ember-ana:
> If you prepare a list of character pairs, i can add it myself.
```
אַ -> א
אָ -> א
בּ -> ב
בֿ -> ב
וּ -> ו
וֹ -> ו
יִ -> י
כּ -> כ
פּ -> פ
פֿ -> פ
שׂ -> ש
תּ -> ת

ײַ -> יי
ײ -> יי
ױ -> וי
װ -> וו
```

those should be all

Miodec:
> > If you prepare a list of character pairs, i can add it myself.
> 
> ```
> אַ -> א
> אָ -> א
> בּ -> ב
> בֿ -> ב
> וּ -> ו
> וֹ -> ו
> יִ -> י
> כּ -> כ
> פּ -> פ
> פֿ -> פ
> שׂ -> ש
> תּ -

image


Does this look correct?

ember-ana:
one thing i see there is that האַלדז became האאלדז with an extra alef?

Miodec:
> one thing i see there is that האַלדז became האאלדז with an extra alef?

This should be correct now right?

image

ember-ana:
lgtm 🐿️

PR #6328, opened 2025-03-01
impr(leaderboard): show typing speed in selected unit (@fehmer)

fehmer:


fehmer:
@Miodec should we update the page if the speedUnit is changed by using the command line?

Miodec (approved):
Thanks

PR #6334, opened 2025-03-05
fix(leaderboards): user button navigating to the wrong page (@fehmer)

fehmer:


frontend/src/ts/pages/leaderboards.ts
```diff
@@ -449,7 +449,8 @@ function updateJumpButtons(): void {
   }
 
   if (isAuthenticated() && state.userData) {
-    const userPage = Math.floor(state.userData.rank / state.pageSize);
+    //TODO: should we hide the user button? if we are on the same page the user can still scroll down to his entry
```
fehmer, line 452: @Miodec what do you think?

Miodec (approved):
Thanks

PR #6338, opened 2025-03-06
chore: Improve branch check in release script (@fehmer)

fehmer:


packages/release/src/index.js
```diff
@@ -227,6 +236,7 @@ const main = async () => {
   console.log("Starting release process...");
 
   checkBranchSync();
+  process.exit(1);
```
fehmer, line 239: remove after testing

Miodec (approved):
Thanks

PR #6341, opened 2025-03-06
impr(language): add more keywords and chars to code_pascal (@fau)

fau:
### Description

Added more keywords from Delphi and Free Pascal and special chars used in pascal code.


frontend/static/languages/code_pascal.json
```diff
@@ -2,34 +2,156 @@
...
+    "#",
+    "%",
+    "^",
+    "+",
+    "<",
+    "<=",
+    "<>",
+    "=",
+    ">",
+    ">=",
```
Miodec, line 35: Im kinda thinking these should be added in punctuation mode.

Miodec (changes requested):

frontend/static/languages/code_pascal.json
fau: I think that commonly used symbols and their combinations in code should be included in the dictionary. This will allow practice not only with keywords but also with these symbols and combinations. This is especially useful for testing the usability of layouts on non-standard keyboards, where symbols and numbers may be placed on a separate layer.

Miodec:
Thanks

PR #6342, opened 2025-03-06
impr(result page): clicking on daily leaderboard rank opens the leaderboards page the user is on (@fehmer)

fehmer:


fehmer:
replaced by #6380

PR #6343, opened 2025-03-06
build: fix openapi spec path for docs & stats endpoints on dev (@NadAlaba)

NadAlaba:
### Description

1. fix openapi specification path so that /docs endpoint would work on dev.

> Error: ENOENT: no such file or directory, stat 'backend/src/static/api/public.html' Stack: Error: ENOENT: no such file or directory

2. fix openapi specification path so that API Operation would work on /stats endpoint on dev.

> Cannot read openApi specification from backend/src/api/routes/../../static/api/openapi.json. Swagger stats will not fully work.
stats

3. remove unneeded `files` & `ts-node` options from `tsconfig.json` in various places.

> File 'backend/src/types/types.d.ts' not found.
  The file is in the program because:
    Part of 'files' list in tsconfig.json

4. add the missing `pnpm test` script to the funbox package.

### DIFF
--- a/backend/src/api/routes/swagger.ts
+++ b/backend/src/api/routes/swagger.ts
@@ -5,7 +5,7 @@ import { readFileSync } from "fs";
 import Logger from "../../utils/logger";
 
 function addSwaggerMiddlewares(app: Application): void {
-  const openApiSpec = __dirname + "/../../static/api/openapi.json";
+  const openApiSpec = __dirname + "/../../../dist/static/api/openapi.json";
   let spec = {};
   try {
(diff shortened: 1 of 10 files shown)


backend/src/api/routes/swagger.ts
```diff
@@ -5,7 +5,7 @@ import { readFileSync } from "fs";
 import Logger from "../../utils/logger";
 
 function addSwaggerMiddlewares(app: Application): void {
-  const openApiSpec = __dirname + "/../../static/api/openapi.json";
+  const openApiSpec = __dirname + "/../../../dist/static/api/openapi.json";
```
fehmer, line 8: is this working on prod/docker image?

PR #6355, opened 2025-03-09
fix: protect against stored xss in custom background url (@fehmer)

### LINKED ISSUE #6354: Stored Self XSS in theme background image

- [x] I have cleared my cache

Yes

No

Yes

Yes

### Account name

DmitryGashko

### Account config

_No response_

### Current Behavior

Payload example:

```
https://asdf.com/404.jpg?q="onerror="alert(`1`)"
```

Payload limitations: 
None - letters, number, symbols, unlimited length.

Severity: 
Low, but still with some social engineering it's possible to hide payload inside the url string and force some users to apply such url and then you can do basically anything you want (e.g. still user session).

For example I can create a website like "Awesome Monkeytype Themes" with backgrounds, and in some put the payload, then the user would click "Copy path", insert without checking, save, done.

(there's also, no need to update the page, it's works instantly on save, and on each page load after that)

Vulnerable code:

Image

(but i think you may want to escape the url before saving/using as well).

Screenshots:

Image

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- OS:
- Browser:
- Browser Version:

### Anything else?

_No response_

fehmer:
closes #6354

### DIFF
--- a/packages/contracts/__test__/schema/config.spec.ts
+++ b/packages/contracts/__test__/schema/config.spec.ts
@@ -54,7 +54,7 @@ describe("config schema", () => {
       },
       {
         name: "javascript url",
-        input: `javascript://alert('asdf')//https://example.com/img.jpg`,
+        input: `javascript:alert('asdf');//https://example.com/img.jpg`,
         expectedError: "Unsupported protocol.",
       },
       {
--- a/packages/contracts/__test__/schema/config.spec.ts
+++ b/packages/contracts/__test__/schema/config.spec.ts
@@ -1,0 +1,82 @@
+import { CustomBackgroundSchema } from "../../src/schemas/configs";
+
+describe("config schema", () => {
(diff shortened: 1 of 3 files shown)


packages/contracts/__test__/schema/config.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        expectedError: "May not contain quotes.",
+      },
+      {
+        name: "back tick",
+        input: `https://example.com/404.jpg?q=alert(\`1\`)`,
+        expectedError: "May not contain quotes.",
+      },
+      {
+        name: "javascript url",
+        input: `javascript://alert('asdf')//https://example.com/img.jpg`,
```
DimaGashko, line 57: Javascript schema is `javascript:`, not `javascript://`

Miodec (approved):
Thanks

PR #6357, opened 2025-03-10
impr(quotes): Adding a quote from the LEGO Movie

CedricBaaklini:
This is quote from one of my all time favorite movies.


fehmer (changes requested):
hi @CedricBaaklini , 

please remove the unrelated commits/changes from the PR, thanks.

PR #6363, opened 2025-03-12
impr: error handling on user deletion (@fehmer)

fehmer:


backend/src/api/controllers/user.ts
```diff
@@ -251,14 +251,22 @@ export async function sendForgotPasswordEmail(
...
+
+  try {
+    userInfo = await UserDAL.getPartialUser(uid, "delete user", [
+      "banned",
+      "name",
+      "email",
+      "discordId",
+    ]);
+  } catch (e) {
+    //userinfo was already deleted. We ignore this and still try to remove the  other data
```
Miodec, line 266: Should we check here if the error is 404, and only then continue - otherwise rethrow?

Miodec (approved):
Thanks

PR #6367, opened 2025-03-12
feat(font): add minecraft font (@maxhu08)

maxhu08:
### Description

Add the minecraft font to the list of available fonts

SS2025-03-12_17-07-43

SS2025-03-12_17-06-53

Closes #


Miodec:
As much as I like this font, its not free for commercial use - Microsoft could be angry if I add it.

Sorry, youll need to use the custom font feature if you want to use this font.

maxhu08:
Unfortunate.

Can you contact mojang/microsoft just to check though?

PR #6344, opened 2025-03-06
chore(workflows): update deprecated action and remove duplicate workflow message (@NadAlaba)

NadAlaba:
### Description

1. remove duplicate correct examples in semantic-pr-title error message, and make the examples more diverse (show how to credit multiple contributors).
patternMessage

2. make the scope (quotes, languages, ...) consistently plural across the repo docs (A subjective change I know, but in the changelog, the scope looks better when it is plural i.e. **quotes:** add various quotes).
3. update deprecated action `upload-artifact` from v3 to v4, because it was making the `ci-failure-comment` workflow fail.
> Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
4. use the same official action to download and upload artifacts across the repo's workflows.
5. add 'waiting for update' and remove 'waiting for review' labels when the Monkey CI workflow fails.
6. fix actions' major version only, to allow minor updates and patches.

PR #6368, opened 2025-03-13
chore(workflows): use specific version of add-remove-label-action (@NadAlaba)

NadAlaba:
revert fixing major version only of the `PauMAVA/add-remove-label-action` action from #6344 because this action does not have a `v1` tag


Miodec:
Thanks

PR #6375, opened 2025-03-14
feat(custom-keymap): Custom Keymap Style(@thewickest)

thewickest:
### Description

Captura de pantalla 2025-03-14 a las 23 23 43
Pasted Graphic

New Keymap Style inside the Keymap section. This serves as a custom keymap where you can put the JSON content generated in the Keyboard Layout Editor or pages that generates equivalent content.
Features:
1. If you include an object with an X in the array of keys, it will be treated as a blank key
2. If you include an object with an W and a size, the size of the key will increment. Possible sizes right now are (1, 1.25, 1.5, 1.75, 2.25, 2.50, 6.25)
3. If you want to maintain the dynamic features of the keymap style, you will have to fullfill the keymap with you prefered layout (qwerty, dvorak, etc).
4. For adding and space, just write "spc" inside the key.

### Checks

- [ ] Adding quotes?

feature(custom-keymap): Custom Keymap Style(@thewickest)


frontend/src/styles/keymap.scss
```diff
@@ -544,4 +544,40 @@
...
+      grid-template-columns: none;
+    }
+
+    .keymapSplitSpacer {
+      display: block;
+      width: 2rem;
+      height: 2rem;
+    }
+
+    .key125 {
```
fehmer, line 564: some sizes missing, e.g. 2, 2.75, 3, 6, 7. I'd suggest to just inline the width in the html.

frontend/src/ts/commandline/lists/keymap-style.ts
```diff
@@ -61,6 +61,14 @@ const subgroup: CommandsSubgroup = {
         UpdateConfig.setKeymapStyle("steno_matrix");
       },
     },
+    {
+      id: "setKeymapStyleCustom",
```
fehmer, line 65: Should we add an option to set the custom keymap in the command line as well. opening a modal with a textfield. (could use  simple-modal so you can verify the input before sending to the backend)

frontend/src/html/pages/settings.html
```diff
@@ -1033,6 +1033,20 @@
         <button data-config-value="split_matrix">split matrix</button>
         <button data-config-value="steno">steno</button>
         <button data-config-value="steno_matrix">steno matrix</button>
+        <button data-config-value="custom">custom</button>
+      </div>
+    </div>
+    <div class="section fullWidth" data-config-name="keymapCustom">
+      <div class="inputs">
+        <div class="textareaAndButton">
```
fehmer, line 1041: kle comes in two flavors, the "raw data" you can copy from the website and the json file. We should support both.

frontend/src/ts/utils/custom-keymap.ts
```diff
@@ -0,0 +1,98 @@
+import { Config } from "@monkeytype/contracts/schemas/configs";
+import { dataKeys as keyToDataObject } from "../constants/data-keys";
+
+export type KeyProperties = {
```
fehmer, line 4: duplicate types. We should use the one from the config schema

frontend/src/ts/utils/custom-keymap.ts
```diff
@@ -0,0 +1,98 @@
...
+  return (key && keyToDataObject[key]) ?? "";
+}
+
+function isKeyProperties(
+  element: KeyProperties | string
+): element is KeyProperties {
+  return typeof element === "object" && element != null;
+}
+
+export function stringToKeymap(keymap: string): KeymapCustom {
```
fehmer, line 24: why do we need to convert between `KeymapCustom` and  `String` ?

frontend/src/ts/utils/custom-keymap.ts
```diff
@@ -0,0 +1,98 @@
...
+  const keymapCopy = [...keymapStyle];
+
+  const keymapHtml = keymapCopy.map(
+    (row: (KeyProperties | string)[], index: number) => {
+      const rowCopy = [...row];
+      const rowHtml = rowCopy.map(
+        (element: KeyProperties | string, index: number) => {
+          let size = "",
+            keyHtml: string = "",
+            keyString: string =
```
fehmer, line 66: We should sanitize or escape the input to prevent xss.
```json
[["Esc\"<script>alert(`!`)</script>"]]
```

packages/contracts/src/schemas/configs.ts
```diff
@@ -77,9 +77,28 @@ export const KeymapStyleSchema = z.enum([
   "split_matrix",
   "steno",
   "steno_matrix",
+  "custom",
 ]);
 export type KeymapStyle = z.infer<typeof KeymapStyleSchema>;
 
+export const KeymapCustomSchema = z.array(
```
fehmer, line 84: not sure how to best implement this, but I guess we should have an upper limit of overall payload size here.

fehmer:
hi @thewickest ,

thank you for your contribution. The feature looks cool. I left some comments on the code.

packages/util/src/strings.ts
```diff
@@ -0,0 +1,12 @@
...
+  if (str === "") {
+    return "";
+  }
+
+  // TODO improve this sanitation
+  return str
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"'"/g, '"&apos;"')
+    .replace(/script/g, "");
```
fehmer, line 11: `script` should be fine. I think we should replace these characters with html entities

backtick  `&#96;` , single quote `'` `&apos;`, double quote  `"` `&quot;`, less than  `<` `&lt;`, greater than  `>`  `&gt;`, and backslash  `\`  `&bsol;`

frontend/src/ts/utils/custom-keymap.ts
```diff
@@ -0,0 +1,98 @@
...
+  return typeof element === "object" && element != null;
+}
+
+export function stringToKeymap(keymap: string): KeymapCustom {
+  try {
+    const processedKeymap = sanitizeString(keymap).replace(
+      /([{,])\s*(\w+)\s*:/g,
+      '$1"$2":'
+    );
+    const jsonKeymap: KeymapCustom = JSON.parse(
```
fehmer, line 25: Please use `parseWithSchema` from  `@monkeytype/util/json`. It will validate the string and strip out properties kle supports but we don't, like key background color,  homing, stepped etc.

fehmer:
Hi, i did some testing.

1. if you use the full size ansi example on kle
```[["Esc",{x:1},"F1","F2","F3","F4",{x:0.5},"F5","F6","F7","F8",{x:0.5},"F9","F10","F11","F12",{x:0.25},"PrtSc","Scroll Lock","Pause\nBreak"],[{y:0.5},"~\n`","!\n1","@\n2","#\n3","$\n4","%\n5","^\n6","&\n7","*\n8","(\n9",")\n0","_\n-","+\n=",{w:2},"Backspace",{x:0.25},"Insert","Home","PgUp",{x:0.25},"Num Lock","/","*","-"],[{w:1.5},"Tab","Q","W","E","R","T","Y","U","I","O","P","{\n[","}\n]",{w:1.5},"|\n\\",{x:0.25},"Delete","End","PgDn",{x:0.25},"7\nHome","8\n↑","9\nPgUp",{h:2},"+"],[{w:1.75},"Caps Lock","A","S","D","F","G","H","J","K","L",":\n;","\"\n'",{w:2.25},"Enter",{x:3.5},"4\n←","5","6\n→"],[{w:2.25},"Shift","Z","X","C","V","B","N","M","&lt;\n,","&gt;\n.","?\n/",{w:2.75},"Shift",{x:1.25},"↑",{x:1.25},"1\nEnd","2\n↓","3\nPgDn",{h:2},"Enter"],[{w:1.25},"Ctrl",{w:1.25},"Win",{w:1.25},"Alt",{w:6.25,a:7},"",{w:1.25,a:4},"Alt",{w:1.25},"Win",{w:1.25},"Menu",{w:1.25},"Ctrl",{x:0.25},"←","↓","→",{w:2,x:0.25},"0\nIns",".\nDel"]]```
the settings page is stuck loading with error message `ERR Syntax error, unrecognized expression: .pageSettings .section[data-config-name=&#039;keymapCustom&#039;]` thrown in [settings-groups:L130](https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/ts/elements/settings/settings-group.ts#L130). The layout is not displaying correctly.

2. Legends like `"!\n1"` don't work.
3. Long legends overflow the key, e.g.  `[["CapsLock"]]`

packages/util/src/strings.ts
thewickest: I also thought that in first place but I  will have to rethink it because the string for sure will contain double quotes, since is a json string, and I can't replace those quotes by html entities.

fehmer: don't you only apply the content of the key to this function, not the full json string? Escaping quotes with html entities should be fine.

frontend/src/html/pages/settings.html
thewickest: Do you mean to be able to upload the json file or just copy the content? It's true that the website has a slightly different format. Maybe we can allow both three.

Miodec:
I tried a simple 60% and its not aligned

image


Also JD40

image


Ergodox is cursed

image

thewickest:
Okey some updates about the feature.
After some tests I have realized that there were some features missing from the Raw syntax that we are using here:
**Improvements:**

- All basic features are working now. Width, height, spacing between keys (horizontal and vertical) and tilted keys. Other more visual features like stepped keys, ghosted keys, color, etc, have been ignored to maintain the same style in the whole application.
- You can included any character that can be shifted. Normal key letters, numbers, etc. They will be shifted using the English layout. So "," will be "<", "/" will be "?", etc. We should include more languages in the future as the keyboard layout option does. We probably can reuse everything that is already created in the layout setting.
- For long texts (more than 3 characters) that are not include in any main key (shift, enter, arrow, Delete, etc) we have to take into account that the user could write anything inside of it so it is up to the user to make the layout "pretty". It would be really difficult to make the texts bigger or smaller depending on the length of the text and also adapt it to the size of the key.

**Considerations**

- Because of everything that I mentioned before, things like more than one character per key is not working. We could make it work but that would break the style of the website I think.
- I have tested many different layouts but edge-cases are a thing and we could have some of them. With rotations and spacing between keys maybe. It shouldn't but it could happen at some point.
- And also, it wouldn't make much sense to use this feature with the layout setting. As I mentioned, we could adapt it to use it with a different language rather than a different layout. If the user wants another layout (dvorak for example) they can include it manually. That way he/she can include whatever custom layout they want.

**Future Improvements**

- As it was mentioned in [this thread](https://github.com/monkeytypegame/monkeytype/pull/6593) we could have something like a UI for creating a custom style inside the setting section, although I think it would have a lot of work and I don't know if it is really worthy for this feature.
- Include the remaining languages rather than the English layout. With languages I mean, the different types of shifted keys that are different depending on the language.

PR #6382, opened 2025-03-18
fix(scroll): hide scroll-to-top button on page navigation (@byseif21)

byseif21:
This fix hides the scroll-to-top button after page navigation, improving the user experience, especially on the test page due to its shorter length. The button no longer stays visible when it's not needed.
Screenshot 2025-03-18 122208
Screenshot 2025-03-18 122103


Miodec:
Thanks but this is already fixed in f84f8508ca658bb640d4f0df79b87ba4e6cfa6f9. I guess this is not yet deployed to the live page, sorry about that.

PR #6388, opened 2025-03-21
feat(filters): Add new tag filtering modes `AND`, `EXACT` & `NOT` (@TristanMarion)

### LINKED ISSUE #4239: Add tags filter logic mode: "or" and "and"

TristanMarion:
### Description

This PR implements issue #4239, which adds a new toggle feature to switch between `OR`, `AND` and `EXACT` logic modes for tag filtering in the account page :
- `OR`: any selected tag must match
- `AND`: all selected tags must match
- `EXACT`: exactly the selected tags must match

[Expected behaviors](https://discord.com/channels/713194177403420752/713196019206324306/1354422612754694226) :
- `AND` mode
  - `none` - Results without tags
  - tags - Results tags should include the selected tag/tags
  - `none` and tag - Nothing
  - `all (or no tag)` - Nothing
- `OR` mode
  - `none` - Results without tags
  - tags - Result tags should include any of the selected tags
  - `none` and tag - Result without tags plus result tags should include any of the selected tags
  - `all (or no tag)` - Everything
- `EXACT` mode
  - `none` - Results without tags
  - tags - Results tags should exactly match the selected tags
  - `none` and tag - Nothing
  - `all (or no tag)` - Nothing

### Checks

Closes #4239

Several checks in video :

## Dataset used

We will mainly have tag `1`, `2` and `3` for our tests, the list of test tags is : `1`, `2`, `3`, `1,2`, `1,3`, `2,3`, `1,2,3`

https://github.com/user-attachments/assets/d4afb820-e2bc-4d9b-b447-0b2cf69679a3

`OR` mode with tag `1` : we only have `1`, `1,2`, `1,3` & `1,2,3`

https://github.com/user-attachments/assets/f0502cc2-1205-482e-8687-0191c41dae9a

`OR` mode with tags `1` & `2` : we only have `1`, `2`, `1,2`, `2,3`, `1,3` & `1,2,3`

https://github.com/user-attachments/assets/300b251d-c41a-4d12-b138-88c60f8c2e86

`AND` mode with tags `1` & `2` : we only have `1,2` & `1,2,3`

https://github.com/user-attachments/assets/5b63ca2d-3433-4752-8d83-585d8081a9ad

`EXACT` mode with tags `1` & `2` : we only have `1,2`

https://github.com/user-attachments/assets/8f65fa8c-c7f0-401e-bd4d-0f9e9875ad59

`EXACT` mode with tag `1` : we only have `1`

https://github.com/user-attachments/assets/ab5781eb-9c0c-48ec-841f-51ab04475285

`AND` mode with tag `1` (same behavior as `OR` with tag `1`) : we only have `1`, `1,2`, `1,3` & `1,2,3`

https://github.com/user-attachments/assets/a2df3287-9470-4b0f-b754-89bebac736c1


frontend/src/ts/pages/account.ts
```diff
@@ -450,18 +450,51 @@ async function fillContent(): Promise<void> {
...
+          // Loop through all valid tags to find which ones are enabled in the filter
+          validTags.forEach((tagId) => {
+            if (tagId !== "none" && ResultFilters.getFilter("tags", tagId)) {
+              enabledTagIds.push(tagId);
+            }
+          });
+
+          if (enabledTagIds.length === 0) {
+            // No tag filters enabled, show everything
+            tagHide = false;
```
NadAlaba, line 480: Since "none" does not get pushed to `enabledTagIds`, does this mean when only "none" is enabled, all results (with & without tags) will be shown?

frontend/src/ts/pages/account.ts
```diff
@@ -450,18 +450,51 @@ async function fillContent(): Promise<void> {
...
           } else {
-            //tag not found in valid tags, meaning probably deleted
-            if (ResultFilters.getFilter("tags", "none")) tagHide = false;
+            // Check if result has ALL the enabled tag filters
+            const resultHasAllTags = enabledTagIds.every((tagId) =>
+              result.tags?.includes(tagId)
+            );
```
NadAlaba, line 485: In your test video, when only "pouet" tag was enabled in "AND" mode, results tagged with both "pouet" & "plop" were shown. If this behavior is intended, I don't agree with it, I think it should only show results tagged with "pouet", (I believe this is what mio is calling "exact" mode).

If this was not intended it could be fixed with adding a check if the 2 arrays have the same length to `resultHasAllTags`, e.g,
```ts
const resultHasAllTags = enabledTagIds.length === result.tags.length &&
enabledTagIds.every(tagId => result.tags.includes(tagId));
```

TristanMarion:
Thanks for the feedbacks 😊 indeed this was intended to have a non exclusive AND (to match at least the tags selected but not all the tags) but if you think that it should be an exact match I can perform the change 👍
Let me know what you prefer and I’ll update it 👌

NadAlaba:
> Thanks for the feedbacks 😊 indeed this was intended to have a non exclusive AND (to match at least the tags selected but not all the tags)

My bad, I see the use case now. However, I think an "exact" mode would be equally useful.
This is just my free opinion as a user, it is up to miodec if he prefers 3 modes or 2.

TristanMarion:
Pretty clear, I’ll handle the changes to match the expected behaviour. Makes sense to have the three modes 👌

frontend/src/ts/pages/account.ts
TristanMarion: Indeed, TBH I did not know how to handle this edge case, do you think we should display only the tests with no tags at all ?

frontend/src/ts/pages/account.ts
```diff
@@ -450,18 +450,82 @@ async function fillContent(): Promise<void> {
...
+
+            if (resultHasAllTags) {
+              tagHide = false;
+            } else if (
+              ResultFilters.getFilter("tags", "none") &&
+              result.tags.length === 0
+            ) {
+              // Special case: "none" tag is enabled and result has no tags
+              tagHide = false;
+            }
```
NadAlaba, line 495: this block will never be reached (it is in an else block that gets executed when `result.tags.length > 0`), I wonder why the linter didn't complain

frontend/src/ts/pages/account.ts
```diff
@@ -450,18 +450,82 @@ async function fillContent(): Promise<void> {
...
+
+            if (resultHasExactTags) {
+              tagHide = false;
+            } else if (
+              ResultFilters.getFilter("tags", "none") &&
+              result.tags.length === 0
+            ) {
+              // Special case: "none" tag is enabled and result has no tags
+              tagHide = false;
+            }
```
NadAlaba, line 526: this block will never be reached

frontend/src/ts/elements/account/result-filters.ts
```diff
@@ -33,6 +33,8 @@ export function mergeWithDefaultFilters(
         merged[groupKey] = id;
       } else if (groupKey === "name") {
         merged[groupKey] = filters[groupKey] ?? defaultResultFilters[groupKey];
+      } else if (groupKey === "tagsFilterMode") {
```
TristanMarion, line 36: I can merge this condition with the previous one as they contain the same code

frontend/src/ts/pages/account.ts
TristanMarion: Actually this is already done in the existing code and this block won't be reachable if the result has no tag so I will remove this part and ensure we have the correct behavior anyway

alarsyo:
Hey there, I was looking for such functionality and landed on this PR. What's the status here? Does this just need a rebase?

PR #4174, opened 2023-04-12
Added Marathi

akhilesh-w:
Created a new file "marathi.json" to add Quotes in "Marathi" language.

PR #6395, opened 2025-03-23
fix(language): fixed Marathi so that it shows up in the list of languages (@agokule)

agokule:
### Description

Right now, the Marathi language should exist on Monkeytype (there is a `marathi.json` file in both `frontend/static/quotes` and `frontend/static/languages`) but it doesn't. This is probably because it isn't in the `_list.json` and `_groups.json` file, so I added it to those files.

This is related to #4174.


agokule:
Ok, I did a Marathi typing test and the text looks broken... I probably should have enabled ligatures now that I think about it.

Should I open a new pull request, or because it's just one line of code, do you want to do that yourself?

PR #6397, opened 2025-03-23
fix(language): remove rustlang "box" keyword (@Huterenok)

Huterenok:
### Description

Removes Rust language "box" keyword, as it was removed a long time ago in rust-lang/rust#49733


TristanMarion:
Hello, I've noticed this keyword is also used in some quotes of `code_rust` language so you may want to remove it there too 😊

TristanMarion:
I did a Cmd + F in VSCode but I may be wrong as it's `Box` and not `box`, I don't really know Rust so it's maybe not relevant 😅 
Screenshot 2025-03-24 at 15 00 11

Screenshot 2025-03-24 at 15 01 09

Sorry if it's an error 😬

PR #6401, opened 2025-03-24
fix: local development using docker not working with pnpm (@fehmer)

fehmer:
Global install of pnpm is not working as user  `node`. As a workaround we add `pnpm` during the build phase of the image and run the entrypoint as user  `node`


Miodec:
Thanks

PR #6407, opened 2025-03-26
fix(language list): updated words for spanish 1k (@CarloskHard)

CarloskHard:
### Checks

- [ ] Adding quotes?
  --> I only updated words in the existing "spanish_1k.json"
- [ ] Make sure to include your GitHub username prefixed with @ inside parentheses at the end of the PR title.

Closes #

I updated the "Spanish _1k.json" list because it was based on an old study from 1975 to 2004 and there was too many political/legal words that are not used often today (not at least in the 1k most used words) and also updated some old words with more modern ones.
There was a lot of basic words missing: there were words like "gastropod" or "triphthong" but "hello" or "yes" weren't there.

After doing my research I found out there isn't any other modern study so I had to manually select the outdated words and manually change them for basic missing words, so this is not perfect but it's better than what there was before. Someone could still filter it a little bit more by changing rare words for more normal ones that better fit the "1k" list (Maybe I'll do it in the future).

Thanks and happy typing!


CarloskHard:
I fixed the tittle altough I have my doubts with the optional scope so let me now if it's okay like that.
Thanks!

Miodec (approved):
Thanks

PR #6409, opened 2025-03-27
impr(language): Update french words list to use 300 most used words (@TristanMarion)

TristanMarion:
### Description

I've updated the list to use the 300 most used words according to the official website of the Ministry of Education : https://eduscol.education.fr/186/liste-de-frequence-lexicale

### Checks


Miodec:
Thanks, but for consistency im keeping each base language 200 words in length. If the user wants more words they can switch to 1k or other versions.

TristanMarion:
Ok, I can reduce the number of words if you want, but this list would be more relevant as they are the most used words, compared to the current list 😊

Miodec:
I would be worried about modifying the base list too much, people could get really confused. It depends if you make a lot of changes or not.

PR #6428, opened 2025-04-06
impr: lazy load chartData on results (@fehmer)

fehmer:
Optimize results endpoint by removing heavy or unused data. 

We load the whole result chart data for up to 1000 results each time, but it is very unlikely the user will view the charts for all old results. By removing the size in my tests went down from 1152kb to 276kb.


frontend/src/ts/pages/account.ts
```diff
@@ -1152,13 +1151,46 @@ $(".pageAccount #accountHistoryChart").on("click", () => {
...
+    target.html('<i class="fas fa-fw fa-spin fa-circle-notch"></i>');
+
+    const response = await Ape.results.getById({
+      params: { resultId: result._id },
+    });
+
+    target.html('<i class="fas fa-chart-line"></i>');
+    target.removeClass("loading");
+
+    if (response.status === 200) {
```
fehmer, line 1180: Todo error handling

frontend/src/ts/pages/account.ts
```diff
@@ -1152,13 +1151,46 @@ $(".pageAccount #accountHistoryChart").on("click", () => {
...
+      chartData = response.body.data.chartData as ChartData;
```
fehmer, line 1181: Todo handle "toolong"

Miodec (approved):
Thanks

PR #6446, opened 2025-04-13
fix(theme): inconsistent/unreadable styling on settings page when using phantom theme (@fehmer)

fehmer:
1. Some buttons have the color animation set like funboxes on the settings page while others have not:
image

2. On the active theme button the effects overlap making it unreadable:
image

3. On the footer color effect was applied to the `textButton` (like github, discord links) but not `button` (like contract and support). This applies to all animated themes.


Miodec:
Thanks

PR #6454, opened 2025-04-16
feat(funbox): add polyglot (@fehmer)

fehmer:
Add polyglot funbox which let you practice on multiple languages at once in a single test.


frontend/src/ts/pages/settings.ts
```diff
@@ -477,19 +477,13 @@ async function fillSettingsPage(): Promise<void> {
...
-    for (const group of languageGroups) {
-      html += `<optgroup label="${group.name}">`;
-      for (const language of group.languages) {
-        const selected = language === Config.language ? "selected" : "";
-        const text = Strings.getLanguageDisplayString(language);
-        html += `<option value="${language}" ${selected}>${text}</option>`;
-      }
-      html += `</optgroup>`;
-    }
+    element.innerHTML = buildLanguageDropdown(
```
fehmer, line 481: maybe use setData instead of modify the html

packages/funbox/src/list.ts
```diff
@@ -437,6 +437,14 @@ const list: Record<FunboxName, FunboxMetadata> = {
     frontendFunctions: ["alterText"],
     name: "ALL_CAPS",
   },
+  polyglot: {
+    description: "Switch between languages in a single test.",
```
fehmer, line 441: "Use words from multiple languages in a single test"

Miodec:
Thanks

PR #6459, opened 2025-04-16
chore: throw error when response in fetchJson() is not json (@NadAlaba)

NadAlaba:
because vite dev server does not throw a 404 when fetching a non existing json document, but responds with an html document and a 200 status code


Miodec (approved):
Thanks

PR #6462, opened 2025-04-17
fix(settings): Update layoutfluid on config change (@fehmer)

fehmer:
When changing the layoutfluid settings via cmdline the settings page did not update.
This also happens on prod so it is not directly related to the change #6445


Miodec (approved):
Thanks

PR #6468, opened 2025-04-19
fix(settings): fix dropdown not rendered correctly (@fehmer)

fehmer:
Dropdowns for layoutfluid and polyglot were not rendered correctly after logout/login


Miodec (approved):
Thanks

PR #6481, opened 2025-04-22
refactor(backend): improve redis and json.parse type safety with zod (@byseif21)

### LINKED ISSUE #5881: Replace JSON.parse with  parseJsonWithSchema

byseif21:
### Description

refactored backend files to enhance type safety and reliability using Zod validation and Redis  instead of JSON.parse , I tried to avoid the files that isn't necessary tho so I hope I don't miss any or included unnecessary ones!! didn't fully test only verified code compilation and partial tests without Redis!!.

Should Close #5881 
Related to #6207


backend/src/api/routes/swagger.ts
```diff
@@ -1,14 +1,21 @@
...
 import { readFileSync } from "fs";
 import Logger from "../../utils/logger";
+import { z } from "zod";
 
 function addSwaggerMiddlewares(app: Application): void {
   const openApiSpec = __dirname + "/../../../dist/static/api/openapi.json";
   let spec = {};
   try {
-    spec = JSON.parse(readFileSync(openApiSpec, "utf8")) as string;
+    // a schema that allows any object structure for OpenAPI spec
```
fehmer, line 13: if we don't do any validation, why not keep `JSON.parse` ?

byseif21: Yah makes sense ! Just did it for consistency with other files ,  thought of better error handling, and future stricter validation if needed,  I guess I will revert it to `JSON.parse` for simplicity, strict schema isn't necessary here either ,  kinda trusted source. 
OR let me know if you’d like me to keep the Zod approach with a tighter schema instead.

byseif21:
Hi @fehmer , thank you for the feedback. Could you please let me know if there is anything else needs changing too?

backend/src/api/routes/swagger.ts
fehmer: i'd just keep JSON.parse. we just pass it down to swagger-stats so i think we dont need to validate it ourself. If swagger-stats is happy with the object, so are we.

byseif21: Got it, that makes sense! I’ll revert it to `JSON.parse`. Thanks for the clarity

backend/src/dal/new-quotes.ts
```diff
@@ -8,20 +8,26 @@ import MonkeyError from "../utils/error";
...
+
+const QuoteDataSchema = z.object({
+  language: z.string(),
+  quotes: z.array(JsonQuoteSchema),
+  groups: z.array(z.tuple([z.number(), z.number()])),
+});
+
+// derive TS types from Zod schemas
+type JsonQuote = z.infer<typeof JsonQuoteSchema>;
+// type QuoteData = z.infer<typeof QuoteDataSchema>;
```
Miodec, line 30: This comment seems a bit redundant - `z.infer` is self explanatory. I would also remove unused `QuoteData` type.

backend/src/init/firebase-admin.ts
```diff
@@ -24,11 +26,25 @@ export function init(): void {
...
+    });
+
+    const serviceAccount = parseJsonWithSchema(
       readFileSync(SERVICE_ACCOUNT_PATH, {
         encoding: "utf8",
         flag: "r",
-      })
+      }),
+      serviceAccountSchema
     ) as ServiceAccount;
```
Miodec, line 48: I don't think a schema for something we don't control is safe. If the structure of this ever changes (all though unlikely), the validation will fail.

I think its safe in this case to just assume the type exported by the package will be correct (`ServiceAccount` in this case)

What do you think @fehmer ?

backend/src/services/weekly-xp-leaderboard.ts
```diff
@@ -61,7 +93,8 @@ export class WeeklyXpLeaderboard {
   ): Promise<number> {
     const { entry, xpGained, timeTypedSeconds } = opts;
 
-    const connection = RedisClient.getConnection();
+    const connection =
+      RedisClient.getConnection() as RedisConnectionWithCustomMethods | null;
```
Miodec, line 97: This typecast is added every time we get the connection. Should we move this and the type definition to `backend/src/init/redis.ts`?

backend/src/services/weekly-xp-leaderboard.ts
```diff
@@ -213,11 +275,21 @@ export class WeeklyXpLeaderboard {
...
-    //TODO parse with zod?
-    const parsed = JSON.parse((result as string) ?? "null") as Omit<
-      XpLeaderboardEntry,
-      "rank" | "count" | "totalXp"
-    >;
+    // safely parse the result error handling
+    let parsed: Omit<XpLeaderboardEntry, "rank" | "totalXp">;
+    try {
+      parsed = parseJsonWithSchema(
+        (result as string) ?? "null",
```
Miodec, line 282: `as string` here is not needed right?

backend/src/services/weekly-xp-leaderboard.ts
```diff
@@ -261,11 +334,14 @@ export async function purgeUserFromXpLeaderboards(
...
+  // using type assertion for Redis custom command
+  await (
+    connection as unknown as {
+      purgeResults: (
+        keyCount: number,
+        uid: string,
+        namespace: string
+      ) => Promise<void>;
+    }
+  ).purgeResults(0, uid, weeklyXpLeaderboardLeaderboardNamespace);
```
Miodec, line 346: Why cast as `unknown` then object when you added the same thing in `RedisConnectionWithCustomMethods`?

backend/src/utils/daily-leaderboards.ts
```diff
@@ -2,19 +2,51 @@ import _, { omit } from "lodash";
...
+    maxRank: number,
+    withScores: string
+  ) => Promise<[string[], string[]]>;
+  purgeResults: (
+    keyCount: number,
+    uid: string,
+    namespace: string
+  ) => Promise<void>;
+};
+
```
Miodec, line 49: Same thing here, i think it should be moved to the redis file.

backend/src/init/firebase-admin.ts
Miodec: Outside the scope of this pr.

So for now, just revert back to `JSON.parse`

byseif21: Makes sense — I'll revert to JSON.parse and remove the schema validation for now.

backend/src/services/weekly-xp-leaderboard.ts
byseif21: sure , even better I will refactor this too and move the one in `daily-leaderboards.ts`

backend/src/services/weekly-xp-leaderboard.ts
```diff
@@ -87,27 +120,43 @@ export class WeeklyXpLeaderboard {
...
+      uid: z.string(),
+      name: z.string(),
+      discordId: z.string().optional(),
+      discordAvatar: z.string().optional(),
+      badgeId: z.number().int().optional(),
+      lastActivityTimestamp: z.number().int().nonnegative().optional(),
+      timeTypedSeconds: z.number().nonnegative().optional(),
+      isPremium: z.boolean().optional(),
+    });
+
```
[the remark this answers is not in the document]
byseif21: tbh I may overthink with this one ,I defined CurrentEntrySchema separately to match the exact Redis hash structure, excluding computed fields like `rank` and `totalXp` and made `lastActivityTimestamp` and `timeTypedSeconds` optional to handle potential edge cases . reusing `XpLeaderboardEntrySchema` makes sense , should I derive it with `.omit({ rank: true, totalXp: true })`, ? as this would keep things more consistent. If there’s any chance older entries might be missing fields, I can add .partial() too?.

backend/src/services/weekly-xp-leaderboard.ts
byseif21: oh my bad, i think it's unneeded yh

backend/src/services/weekly-xp-leaderboard.ts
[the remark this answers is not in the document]
Miodec: Yeah, just use `omit`. Older entries wont exist because data is thrown away after 2 weeks.

backend/src/services/weekly-xp-leaderboard.ts
byseif21: sorry too, the double cast is unneeded with `RedisConnectionWithCustomMethods`,I should have called `urgeResults` directly I guess. will fix

backend/src/services/weekly-xp-leaderboard.ts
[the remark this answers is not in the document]
byseif21: Okaaaay

backend/src/init/redis.ts
```diff
@@ -73,11 +109,11 @@ export function isConnected(): boolean {
...
-export function getConnection(): IORedis.Redis | undefined {
+export function getConnection(): RedisConnectionWithCustomMethods | null {
   const status = connection?.status;
   if (connection === undefined || status !== "ready") {
-    return undefined;
+    return null;
   }
 
-  return connection;
+  return connection as RedisConnectionWithCustomMethods;
```
Miodec, line 118: Any specific reason why the switch to `null`?

backend/src/utils/discord.ts
```diff
@@ -58,7 +59,7 @@ export async function iStateValidForUser(
   state: string,
   uid: string
 ): Promise<boolean> {
-  const connection = RedisClient.getConnection();
+  const connection = RedisClient.getConnection() as Redis | null;
   if (!connection) {
```
Miodec, line 63: Are these typecasts necessary?

backend/src/init/redis.ts
byseif21: not much really, type clarity!! just thought clearer way to say “no connection exists”. null is often used for intentional absence (like “no value here”), while undefined can mean “not set” or “uninitialized.” for a function returning a connection, null feels more deliberate?.

byseif21:
> @byseif21 modified the xp leaderboard code a bit, check my latest commit and let me know what you think

Very nice work, that's even cleaner! seems perfect to me.

Miodec:
Thanks!

byseif21:
> Thanks!

Glad to contribute on that and thanks for the help

PR #6492, opened 2025-04-26
refactor: add trycatch util (@miodec)

Miodec:
Adds trycatch util to cleanup try catch code.


packages/util/src/trycatch.ts
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    let data = await promiseOrFunction;
+    return { data, error: null };
+  } catch (error) {
+    return { data: null, error: error as E };
+  }
+}
+
+export function tryCatchSync<T, E = Error>(
+  functionOrValue: () => T
```
fehmer, line 27: always a function?!

fehmer (dismissed):

fehmer (approved):

PR #6546, opened 2025-05-09
fix: remove some misspellings in english_450k (@byseif21)

byseif21:
### Description

Remove & correct  some common misspelled words, mainly in english_450k


Miodec (approved):
Thanks

PR #6554, opened 2025-05-10
fix(settings): handle boolean values for indicateTypos (@fehmer)

fehmer:
Handle legacy values for indicateTypos.

Fixes FRONTEND-MJ


Miodec (approved):
Thanks

PR #6556, opened 2025-05-12
impr: add customLayoutfluid and customPolyglot to commandline (@fehmer)

fehmer:


fehmer:
will add too many results to the command line. Make the new lists only show if the user selects the command, like forcing wmanual" on them.

fehmer:
closed in favor of another aproach

PR #6559, opened 2025-05-13
impr(commandline): improve performance (@fehmer)

fehmer:


frontend/src/ts/commandline/commandline.ts
```diff
@@ -361,6 +363,10 @@ async function showCommands(): Promise<void> {
   }
 
   const list = (await getList()).filter((c) => c.found === true);
+  if (lastList && areSortedArraysEqual(list, lastList)) {
```
fehmer, line 366: this avoids buidling the html if the list is the same and also reduces the time for recalculate style and layout.

frontend/src/ts/commandline/commandline.ts
```diff
@@ -458,28 +464,8 @@ async function showCommands(): Promise<void> {
   if (firstActive !== null && !usingSingleList) {
     activeIndex = firstActive;
   }
-  element.innerHTML = html;
 
-  for (const command of element.querySelectorAll(".command")) {
```
fehmer, line 463: moving the mouse listeners to the container element

frontend/src/ts/commandline/commandline.ts
```diff
@@ -573,23 +559,20 @@ async function runActiveCommand(): Promise<void> {
...
-  } catch (e) {
-    if (e instanceof Error) {
-      console.log("could not scroll suggestions: " + e.message);
-    }
+
+  const active: HTMLElement | null = document.querySelector(
+    ".suggestions .command.active"
+  );
+
+  if (active === null || active.dataset["index"] === lastActiveIndex) {
```
fehmer, line 570: skip scrolling if we don't need to reduces the time for recalculate style and layout

PR #6560, opened 2025-05-13
fix: smooth transition for theme palette icon in chaos_theory theme (@byseif21)

byseif21:
### Description

fix the transition animation for the theme palette icon in the chaos_theory theme.

hovering over the current theme button would trigger a smooth 180° flip of the palette icon. However, when quickly moving the cursor away, the icon would abruptly snap back to its original position instead of transitioning smoothly.
### Before: 

https://github.com/user-attachments/assets/55b6683a-c83b-4f93-806b-bf289bf30c5f

-------------

### Now:

https://github.com/user-attachments/assets/59d08d45-ba2d-4921-9cc4-ee84455a0571

### DIFF
--- a/frontend/static/themes/chaos_theory.css
+++ b/frontend/static/themes/chaos_theory.css
@@ -94,6 +94,11 @@ header.focus #logo:after {
   direction: ltr;
 }
 
+footer .leftright .right .current-theme .fas.fa-fw.fa-palette {
+  transform: rotateY(0deg);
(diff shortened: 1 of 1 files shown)


byseif21:
It was working fine here (https://github.com/monkeytypegame/monkeytype/pull/1727#issuecomment-899875378)

Miodec (approved):
Thanks

PR #6581, opened 2025-05-20
fix: auto-refresh themes UI when navigating to settings page (@byseif21)

byseif21:
### Description
When (changing / adding or removing themes from favorite ) via command line or from the current theme button "in page e.g test page" and then
navigating to Settings > Themes, the UI does not immediately reflect the current theme changes . The user has to manually refresh the page to see these changes. 

**Now**
 UI automatically refreshes
to reflect the current theme selection and favorites without requiring
a manual page refresh.

- Added refreshThemeUI function in theme-picker.ts that updates all
  theme-related UI components
- Called this refresh function in the settings page's beforeShow lifecycle hook to ensure the theme UI is always up-to-date when the page is displayed.

**FOR TESTING BEFORE & AFTER:**
keep changing themes then navigate to the settings and check and keep doing those things again to verify that the active theme and favorites update.

### DIFF
--- a/frontend/src/ts/elements/settings/theme-picker.ts
+++ b/frontend/src/ts/elements/settings/theme-picker.ts
@@ -39,6 +39,12 @@ function updateActiveButton(): void {
     ?.classList.add("active");
 }
 
+export async function refreshThemeUI(): Promise<void> {
+  await refreshPresetButtons();
+  updateActiveButton();
+  await refreshCustomButtons();
+}
+
 function updateColors(
   colorPicker: JQuery,
   color: string,
--- a/frontend/src/ts/pages/settings.ts
+++ b/frontend/src/ts/pages/settings.ts
@@ -1364,6 +1364,7 @@ export const page = new Page({
     await UpdateConfig.loadPromise;
     await fillSettingsPage();
     await update();
+    await ThemePicker.refreshThemeUI();
   },
 });
 


Miodec:
This change will cause multiple calls to fill settings page on first load. This needs to be done another way.

byseif21:
> This change will cause multiple calls to `refreshCustomButtons` / `refreshPresetButtons` on first load. This needs to be done another way.

hey mio check this approach and tell me what do you think of it !

I’ve added a flag to track whether the UI has already been set up. Now refreshThemeUI() skips redundant work unless the state is explicitly reset, I kept thinking for other ways like an event-based approach or refactor the theme UI with lifecycle methods, but this is the one with minimal changes I feel, so let me know what do you feel about that and if you spot any edge cases I might’ve missed!

Miodec:
How about calling `fillPresetButtons` and `fillCustomButtons` in `Settings.fillSettingsPage`, then calling `updateActiveButton` in `Settings.update`?

byseif21:
> How about calling `fillPresetButtons` and `fillCustomButtons` in `Settings.fillSettingsPage`, then calling `updateActiveButton` in `Settings.update`?

I did it thinking it would be the best we could do here, but then I noticed that we can't make this work without not calling fillPresetButtons inside update() atp to keep favorites in sync, not just updateActiveButton. That means extra DOM work and redundancy. I have been thinking about that for a while so Idk what to do here actually! I'll rethink it again tom :( .

Miodec (approved):
Thanks

PR #3855, opened 2022-12-26
Adding new Ukrajins'ka latynka language

tumophiy:
### Description

Ukraine is fighting and trying to improve on other directions as well. One of those improvements that is being discussed right now is moving from cyrillic alphabet to latin. If you have time and interest I recommend you to watch [this video](https://www.youtube.com/watch?v=nHeE2x2UNw4), which has great details and explanations about why we are doing this.

More over many people already migrated to latin alphabet! A proof of that is [website](https://nachasi.com/ul/) fully in ukrains'ka latynka. I started using latin layout a couple days ago and sadly noticed that there is no training tool for it(

I think I'm not the only one who is searching for it, so your website will have a benefit after including it.

Thank's for ui and ux it's a joy to use! Hope to get a feedback and merge!

PR #6584, opened 2025-05-21
fix(language): update Ukrainian Latin to national standard (@paiv)

paiv:
### Description

In short, this PR is based on Ukraine's national standard [DSTU 9112:2021 (A)](https://en.wikipedia.org/wiki/DSTU_9112:2021), and replaces non-standard transliteration originally submitted.

- Updates #3855

### Context

Unfortunately, this topic has been a bit toxic in Ukraine, I am sorry to bring this on the developers of this popular tool. I am trying to stay neutral [¹](https://paiv.github.io/blog/2024/11/26/ukrainian-latin.html "The state of Ukrainian Latin"), with Ukraine's ultimate benefits in mind.

At the moment, there is no definite Ukrainian Latin script. The state standard (KMU 55) is a lossy transliteration, not useable for general writing, only applied to personal names and places.

Practitioners of Latin script for the Ukrainian language is still only a marginal group, without a unifying movement. Until 2021 basically everyone had their own transliteration method, derived from two dozen historical schemes.

In 2021 comes Ukraine's national standard DSTU 9112:2021, and objectively is good enough for general writing among alternatives. It does not prescribe transition from Cyrillic. Its future is in integration with European languages, gradually replacing legacy KMU 55.

Thus my argument for the practitioners of Ukrainian Latin script is to adopt DSTU standard, given its perspective and unifying power, and faze out non-standard schemes, of which #3855 is only one.

I hope @tymof1j as the original contributor could critically review these notes, with the two year perspective.

The script used for conversion from Cyrillic: https://gist.github.com/paiv/df2f38ed86a103471a49cfa8064d0d2e


tymof1j:
in 1 hour i'll get back to this

PR #6586, opened 2025-05-22
fix(commandline): improve caching to fix stale checkmark and UI state (@byseif21, @fehmer)

byseif21:
### Description

Previously, when selecting a language via the text button on the test page, the checkmark (fa-check) in the commandline language list wouldn't update until a page refresh. This was due to the commandline's caching mechanism not detecting changes triggered outside its own control.

Although it first appeared to be a language-specific issue, it was later identified that the caching logic was generally insufficient — it didn’t account for updates to the active command state or configuration flags like usingSingleList.

## Solution
Fixed the caching logic in the commandline module by tracking a more complete internal state. The system now correctly detects changes in the command list, active state, and configuration, and rebuilds the list UI when necessary.

## Technical Details
- The commandline uses a caching mechanism (`lastList`) to avoid rebuilding the HTML if the list hasn't changed done in #6559
- Replaced the old lastList cache with a new lastState object that stores: the list of commands, each with its isActive flag, the usingSingleList configuration flag
- Improved the cache comparison logic, uses areSortedArraysEqual to compare command lists including active state ,compares the usingSingleList flag
- Previously, this cache wasn't being cleared when the language changed through the text button
- Now we clear the cache on changes, forcing a rebuild of the list with the correct checkmark

## Performance Impact
- Minimal performance impact
- The list is only rebuilt when:
  1. The language actually changes
  2. The list content changes
  3. The input value changes
- The caching mechanism still prevents unnecessary rebuilds in all other cases

## Testing
- [x] Language selection through text button updates checkmark immediately
- [x] Language selection through commandline works as before
- [x] No unnecessary rebuilds when language hasn't changed
- [x] Checkmark appears next to correct language in all cases

### DIFF
--- a/frontend/src/ts/commandline/commandline.ts
+++ b/frontend/src/ts/commandline/commandline.ts
@@ -202,7 +202,6 @@ async function goBackOrHide(): Promise<void> {
     await filterSubgroup();
     await showCommands();
     await updateActiveCommand();
-    //lastSingleListModeInputValue = "";
     return;
   }
 
--- a/frontend/src/ts/commandline/commandline.ts
+++ b/frontend/src/ts/commandline/commandline.ts
@@ -39,7 +39,7 @@ let subgroupOverride: CommandsSubgroup | null = null;
 let isAnimating = false;
 let lastSingleListModeInputValue = "";
 
-type CommandWithActiveState = Omit<Command, "active"> & { active: boolean };
+type CommandWithActiveState = Omit<Command, "active"> & { isActive: boolean };
 
 let lastState:
   | {
@@ -202,7 +202,7 @@ async function goBackOrHide(): Promise<void> {
     await filterSubgroup();
     await showCommands();
     await updateActiveCommand();
-    lastSingleListModeInputValue = "";
+    //lastSingleListModeInputValue = "";
     return;
   }
 
@@ -363,33 +363,6 @@ async function getList(): Promise<Command[]> {
   return (await getSubgroup()).list;
 }
 
-function fillActiveState(command: Command): CommandWithActiveState {
-  let isActive = false;
-
-  if (command.active !== undefined) {
-    isActive = command.active();
-  } else if (command.configKey !== undefined) {
-    if (command.configValueMode === "include") {
-      isActive = (
-        Config[command.configKey as keyof typeof Config] as (
-          | string
-          | number
-          | boolean
-          | number[]
-          | undefined
-        )[]
-      ).includes(command.configValue);
-    } else {
-      isActive =
-        Config[command.configKey as keyof typeof Config] ===
-        command.configValue;
-    }
-  }
-
-  const { active: _originalActive, ...restOfCommand } = command;
-  return { ...restOfCommand, active: isActive };
-}
-
 async function showCommands(): Promise<void> {
   const element = document.querySelector("#commandLine .suggestions");
   if (element === null) {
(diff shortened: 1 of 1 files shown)


fehmer (changes requested):
Hi,

I think the caching is not correct for other cases as well, not just language. So listening on config changes for language is not sufficient. 

We could extend the list of commands with the active flag and compare the  `list`  and `lastList`  including the active flag.  in addition we should store the usingSingleList flag and compare this as well.

e.g rename lastList to lastState
```ts
let lastState: {
  list: (Command&{active:boolean})[],
  usingSingleList: boolean
}
```

and the cache check:
```ts
const list = (await getList()).filter((c) => c.found === true).map(fillActiveState);
if(lastState && 
  usingSingleList == lastState.usingSingleList &&
  areSortedArraysEqual(list, lastState.list) 
{ 
  return;
}

...
```

byseif21:
> Hi,
> 
> I think the caching is not correct for other cases as well, not just language. So listening on config changes for language is not sufficient.
> 
> We could extend the list of commands with the active flag and compare the `list` and `lastList` including the active flag. in addition we should store the usingSingleList flag and compare this as well.
> 
> e.g rename lastList to lastState
> 
> ```ts
> let lastState: {
>   list: (Command&{isActive:boolean})[],
>   usingSingleList: boolean
> }
> ```
> 
> and the cache check:
> 
> ```ts
> const list = (await getList()).filter((c) => c.found === true).map(fillActiveState);
> if(lastState && 
>   usingSingleList == lastState.usingSingleList &&
>   areSortedArraysEqual(list, lastState.list) 
> { 
>   return;
> }
> 
> ...
> ```

Hi @fehmer 
Thank you for the feedback , can you review the changes and let me know if I did it right or if there is anything that needs improvement or if I missed any ?

frontend/src/ts/commandline/commandline.ts
```diff
@@ -354,14 +363,32 @@ async function getList(): Promise<Command[]> {
...
-// cear the cached list when language changes to ensure UI updates correctly
-ConfigEvent.subscribe((eventKey) => {
-  if (eventKey === "language") {
-    lastList = undefined;
+function fillActiveState(command: Command): CommandWithActiveState {
+  let isActive = false;
+
+  if (command.active !== undefined) {
+    isActive = command.active();
+  } else if (command.configKey !== undefined) {
```
fehmer, line 371: this changes the existing logic which is using the subgroup.configKey as fallback https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/ts/commandline/commandline.ts#L390

i'd keep this code inside the ```showCommands```  like this:

```ts
function showCommands()...{
...
const subgroup = await getSubgroup();
const list: (Command & { isActive: boolean })[] = subgroup.list
    .filter((c) => c.found === true)
    .map((command) => {
      let isActive = false;
      if (command.active !== undefined) {
        isActive = command.active();
      } else {
        const configKey = command.configKey ?? subgroup.configKey;
        if (configKey !== undefined) {
          if (command.configValueMode === "include") {
            isActive = (
              Config[configKey] as (
                | string
                | number
                | boolean
                | number[]
                | undefined
              )[]
            ).includes(command.configValue);
          } else {
            isActive = Config[configKey] === command.configValue;
          }
        }
      }
      return { ...command, isActive };
    });
```

frontend/src/ts/commandline/commandline.ts
```diff
@@ -376,19 +403,20 @@ async function showCommands(): Promise<void> {
...
+    areSortedArraysEqual(listWithActiveState, lastState.list)
   ) {
     return;
   }
-  lastList = list;
+
+  lastState = {
+    list: listWithActiveState,
+    usingSingleList: usingSingleList,
+  };
```
fehmer, line 419: the code to calculate the isActive flag is duplicated now. Please remove it further down and use the new  `isActive` flag on the list like this:

```ts
 for (const command of list) {
    if (command.found !== true) continue;
    ...
    let configIcon = "";
    if (command.isActive) {
      firstActive = firstActive ?? index;
      configIcon = `<i class="fas fa-fw fa-check"></i>`;
    } else {
      configIcon = `<i class="fas fa-fw"></i>`;
    }

const iconHTML = `<div class="icon">${
      usingSingleList || configIcon === "" ? icon : configIcon
    }</div>`;
....
```

frontend/src/ts/commandline/commandline.ts
```diff
@@ -194,6 +202,7 @@ async function goBackOrHide(): Promise<void> {
     await filterSubgroup();
     await showCommands();
     await updateActiveCommand();
+    lastSingleListModeInputValue = "";
```
fehmer, line 205: unused?

fehmer (changes requested):

frontend/src/ts/commandline/commandline.ts
byseif21: was testing somethingbut it's out of the scope rn so nvm

byseif21:
@fehmer can you please check again and tell me if still there is anything??

fehmer (approved):
Hi @byseif21,

thanks for the fix, lgtm!

PR #6593, opened 2025-05-24
feat: Caps Lock Off (@Leonbacd123)

Leonabcd123:
### Description

Added a settting that allows the user to disable Caps Lock. Even if the user accidentally presses Caps Lock in the middle of a test, it will be ignored and the input will still be lowercase.


Miodec:
Im honestly not sure if this setting is really needed, feels like too much hand holding for the user. Sorry.

Leonabcd123:
> Im honestly not sure if this setting is really needed, feels like too much hand holding for the user. Sorry.

I'm also working on a feature that allows the user to create their own layouts and save locally. Would you be interested in that?

fehmer:
> > Im honestly not sure if this setting is really needed, feels like too much hand holding for the user. Sorry.
> 
> I'm also working on a feature that allows the user to create their own layouts and save locally. Would you be interested in that?

there was already an attempt to implement something like this: https://github.com/monkeytypegame/monkeytype/pull/6375

Leonabcd123:
> > > Im honestly not sure if this setting is really needed, feels like too much hand holding for the user. Sorry.
> > 
> > 
> > I'm also working on a feature that allows the user to create their own layouts and save locally. Would you be interested in that?
> 
> there was already an attempt to implement something like this: #6375

I was thinking of doing something where you have a keymap, and you can edit each key in it, instead of doing something where you import a layout. What do you think of that?

PR #6608, opened 2025-05-31
fix(account): fix not all results shown if result has a funbox unknown to stored filter (@fehmer)

fehmer:
Merge with default filters to ensure all languages and funboxes are present in the result filter.


Miodec:
Thanks

PR #6615, opened 2025-06-04
fix(commandline): prevent unwanted scrolling on initial mouse entry in theme picker (@byseif21)

byseif21:
### Description

fix this weird behavior scroll:

https://github.com/user-attachments/assets/73a7c9f2-f584-4d4a-9932-dbb565d99be6


Miodec (approved):
Thanks

PR #6624, opened 2025-06-06
impr(nav): avoid duplicate browser history entries when re-clicking same nav button (@byseif21)

### LINKED ISSUE #6623: Navigation buttons add redundant history entries on repeated clicks

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

byseif21

### Account config

_No response_

### Current Behavior

When using Monkeytype, clicking the same navigation icon repeatedly (e.g., Settings, Leaderboard, or About) creates multiple browser history entries, even though the user is already on that page. This results in users having to press the back button multiple times to return to the previous page.

### Expected Behavior

It should be only one click to return to the previous page regardless of how many times they clicked the same navigation button.

### Steps To Reproduce

1. Navigate to any page (e.g., Settings)
2. Click the same navigation button multiple times
3. Press the browser back button

### Environment

- OS: all
- Browser: all
- Browser Version:

### Anything else?

_No response_

byseif21:
### Description

Closes #6623


frontend/src/ts/controllers/route-controller.ts
```diff
@@ -159,7 +159,17 @@ export function navigate(
...
+
+  // only push to history if we're navigating to a different URL
+  const currentPath =
+    window.location.pathname + window.location.search + window.location.hash;
+  const targetUrl = new URL(url, window.location.origin);
+  const targetPath = targetUrl.pathname + targetUrl.search + targetUrl.hash;
+
+  if (currentPath !== targetPath) {
+    history.pushState(null, "", url);
+  }
```
pran01: Curious if there’s a reason we’re not directly comparing `window.location.href` with `url` — since the click handler calls this `navigate` function with `target.href`, it might work out cleanly and will simplify this. What do you think?

byseif21: Hi @pran01, I thought about using window.location.href !== url, but I went with this approach because url might be relative (like /settings), while href is always absolute. That mismatch could cause false negatives.

By comparing just the path, search, and hash, I make sure we only push when the actual navigable part changes—regardless of environment or base URL. It's a bit more verbose, but safer for typical SPA routing. that what I thought of.

byseif21: but rethinking this about simplifing thing, i think we could simplify this to be like that 
```
const currentUrl = new URL(window.location.href);
const targetUrl = new URL(url, window.location.origin);

if (
  currentUrl.pathname + currentUrl.search + currentUrl.hash !==
  targetUrl.pathname + targetUrl.search + targetUrl.hash
) {
  history.pushState(null, "", url);
}
```
Same behavior, just a little cleaner. I will wait for mio tho.

Miodec: > but rethinking this about simplifing thing, i think we could simplify this to be like that
> 
> ```
> const currentUrl = new URL(window.location.href);
> const targetUrl = new URL(url, window.location.origin);
> 
> if (
>   currentUrl.pathname + currentUrl.search + currentUrl.hash !==
>   targetUrl.pathname + targetUrl.search + targetUrl.hash
> ) {
>   history.pushState(null, "", url);
> }
> ```
> 
> Same behavior, just a little cleaner. I will wait for mio tho.

Yeah i would prefer that.

Miodec (approved):
Thanks

PR #6629, opened 2025-06-09
refactor(hints): improve readability of hints functions (@NadAlaba)

NadAlaba:
- Refactoring `joinOverlappingHints()` solves the issue of hints joining when they correspond to letters of different lines, which was one of the issues [#6636](https://github.com/monkeytypegame/monkeytype/pull/6636) was supposed to solve.
- Also add a utility that debounces/locks async functions.


frontend/src/ts/utils/misc.ts
```diff
@@ -678,6 +678,47 @@ export function promiseWithResolvers<T = void>(): {
...
+ * Wraps an asynchronous function to prevent it from running until it is finished and settled.
+ * If the function is called while a previous call is still running,
+ * the arguments from the most recent call are queued and run after the current one finishes.
+ * The returned debounced function returns a `Promise<boolean>` that:
+ * - Resolves to `true` if the function was executed immediately.
+ * - Resolves to `false` if the call was skipped or queued (another call is running).
+ * @param asyncFn The asynchronus function to debounce
+ * @returns The debounced function
+ */
+export function debounceUntilResolved<TArgs extends unknown[], TResult>(
```
Miodec, line 692: Maybe this should return something like an object with `status: "ran" | "skipped" | "queued"` and `value: Promise<TResult>`? Currently the return value of the original function is lost.

Miodec (approved):
Thanks

PR #6630, opened 2025-06-09
feat(settings): allow uploading local background images (@byseif21)

byseif21:
### Description

This PR adds support for uploading custom background images directly from local files, in addition to the existing URL-based method. users can now personalize their typing environment more easily without needing hosted images.

### Short Test & Preview:

https://github.com/user-attachments/assets/28c53651-c1fc-45a1-8a52-f50557f7f209

### DIFF
--- a/frontend/src/html/pages/settings.html
+++ b/frontend/src/html/pages/settings.html
@@ -1126,9 +1126,9 @@
         <span>custom background</span>
       </div>
       <div class="text">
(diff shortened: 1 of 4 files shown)


Miodec:
While i appreciate the work here, this could lead to either security issues or content issues. And by content issues i mean users storing potentially illegal files. I think I don't really wanna deal with that, hope you understand.

PR #6636, opened 2025-06-12
fix(hints): place hints between lines of multi-line words (@NadAlaba)

nadalaba:
### Problem

1. Hints of multi-line words are being positioned below the entire word instead of below their corresponding letters, causing overlaps between hints of different lines, (this issue will be deferred).
2. Hints corresponding to letters of different lines can join when they are adjacent letters and their hints overlap, (this issue will be solved in [#6629](https://github.com/monkeytypegame/monkeytype/pull/6629)).
multiLinesHints

### Solution

1. Increase spacing between lines of a multi-line word:
    - increase `line-height` to 1.3, and remove vertical margins of `.word` elements, to unify spacing in lines between words and lines of the same word.
2. Place hints between lines of a multi-line word.
4. trigger `udpateHintsPosition()` on window.resize and Config.tapeMode change.
5. remove unneeded hints styles.

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -283,11 +283,11 @@
       font-size: 0.75em;
       text-shadow: none;
       //padding: 1px;
-      left: 0;
+      //left: 0;
       opacity: 0.5;
-      text-align: center;
-      display: grid;
-      justify-content: center;
+      //text-align: center;
+      //display: grid;
+      //justify-content: center;
       transform: translate(-50%, 1.1em);
     }
   }


Miodec:
While everything seems to be working correctly, with the changed line spacing you now can't tell that its one long word. It looks like you need to press space to go to the next word, when in reality its still the same word:

image

The line spacing might need to be reverted... but then there is no space for the typo indication..

nadalaba:
> with the changed line spacing you now can't tell that its one long word.

Closing for now until we find a way for users to distinguish between lines of same word and different words.

PR #6645, opened 2025-06-19
fix(funbox): proper rtl and ligature handling in polyglot with mixed languages (@byseif21)

byseif21:
## Description
This PR addresses an issue in Monkeytype's funbox "polyglot" mode where mixing right-to-left (RTL) and left-to-right (LTR) languages—while the main language is set to LTR—would break ligature rendering for RTL languages.

### Key Changes
- Introduces PolyglotResult type:
The withWords method in funbox functions may now return a PolyglotResult object containing the generated wordset and metadata about directionality (allRightToLeft) and ligature support (allLigatures).

- Updates polyglot logic:
When polyglot mode is active, the system checks all included languages for RTL and ligature support and returns this info in the result.

#### Notes
- Maintains backward compatibility for non-polyglot funboxes.
- Adds runtime type checks for the new result type.
- No breaking changes expected for existing test modes.


byseif21:
handled in #6666

PR #6650, opened 2025-06-21
fix(lazymode): add spanish ñ to accents map (@sashoism)

sashoism:
### Description

Move all accented `n`s to a single line and add the missing Spanish `ñ`.


Miodec:
`ñ` is part of top layer in the spanish layout (right pinkie), thats why its not added to the lazy mode characters. Thanks for your pr though.

PR #6658, opened 2025-06-24
feat: add friend requests and list (@fehmer)

fehmer:
make some friends on monkeytype


K87lk:
What is the point of adding friends if you can not interact with them in any way?

PR #6663, opened 2025-06-25
feat: allow user to use local file as background (@fehmer, @byseif21)

fehmer:
Allow the user to use a local file as custom background without uploading it to the server.

Based on @byseif21 work in #6630, thanks!


frontend/src/ts/pages/settings.ts
```diff
@@ -1076,10 +1091,50 @@ $(
 
 $(
   ".pageSettings .section[data-config-name='customBackgroundSize'] .inputAndButton button.remove"
-).on("click", () => {
+).on("click", async () => {
+  if (Config.customBackground === "localBackgroundFile") {
+    await FileStorage.deleteFile("localBackgroundFile");
```
byseif21: just a small thing when deleting the local background, the input box will still show the previous text. would it be good to add 
```js
 $(".pageSettings .section[data-config-name='customBackgroundSize'] .inputAndButton input.input").val("");
 ```
 or something to clear it  since it's not reusable like the url ? if you understand what i mean

**Comment by byseif21** [comment]
hey @fehmer thanks and yah very nice , just don't we forget to update the setting description too. or you think the aria-label enough ?

File: frontend/src/ts/pages/settings.ts (hunk 1 above)
Comment (fehmer) [reply frontend/src/ts/pages/settings.ts]: yeah, thanks.

**Comment by byseif21** [comment]
also didn't we still need to add a file size check before storing?

**Comment by fehmer** [comment]
hi @byseif21 , thanks for your comments. I updated the description and icon to match the same as "open file" in custom text settings. An upload icon might be confusing.

I don't think we need a file size check, the limit for idb is like 20gig or 10% of the users disc space, so we should be fine.

File: frontend/src/ts/constants/default-config.ts [hunk 2]
```diff
@@ -109,3 +109,5 @@ const obj = {
 export function getDefaultConfig(): Config {
   return deepClone(obj);
 }
+
+export const LocalBackgroundFile = "LocalBackgroundFile";
```
fehmer: don't know where to put, causes circle in `theme-controller`.

PR #6666, opened 2025-06-26
impr(funbox): add lazy mode support per word in polyglot (@byseif21)

### LINKED ISSUE #6665: Lazy mode disabled in Polyglot funbox when any language doesn't support It

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

When using the Polyglot Funbox with multiple languages, if any of the selected languages does not support Lazy Mode , then Lazy Mode is disabled entirely — even for the other languages that do support it.

### Expected Behavior

Lazy Mode should be applied per word, based on the origin language of each word

### Steps To Reproduce

1- Go to Settings → Funbox → Enable Polyglot mode

2- Select languages: english, arabic

3- Enable Lazy Mode

4- Start a typing test

5- Observe that Lazy Mode is completely disabled, even for arabic words

### Environment

- OS:
- Browser:
- Browser Version:

### Anything else?

_No response_

byseif21:
### Description

closes #6665


frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -687,9 +696,30 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
...
+          wordSet.push(word);
+          wordLazyModeSupport.set(word, supportsLazyMode);
+        }
+      }
+
       Arrays.shuffle(wordSet);
-      return new Wordset(wordSet);
+
+      // compute RTL and ligature info
+      const allRightToLeft = languages.some((lang) => lang.rightToLeft);
```
fehmer: if you want all shouldn't it be `languages.every`, same below

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -28,10 +28,19 @@ import {
   Layout,
 } from "@monkeytype/contracts/schemas/configs";
 import { Language } from "@monkeytype/contracts/schemas/languages";
+
+// type for polyglot funbox results
+export type PolyglotResult = {
```
fehmer: what about
 
- rename to PolyglotWordset or MultiLanguageWordset.
- make it a class, extend WordSet and change `words: string[];` to `words()=>string[]`
- have a property `wordsWithLanguage` of type `Map<string, Language>`
- have  `words()` returning `wordsWIthLanguages.keys()`
- add property `languageProperties` of type `Map<language,Pick<LanguageObject,   "noLazyMode"|"ligatures"|"rightToLeft">>`

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -155,6 +157,30 @@ class PseudolangWordGenerator extends Wordset {
...
+type LanguageProperties = Pick<
+  LanguageObject,
+  "noLazyMode" | "ligatures" | "rightToLeft"
+>;
+
+export class PolyglotWordset extends Wordset {
+  public override isPolyglot = true;
+  private wordsWithLanguage: Map<string, Language>;
+  public languageProperties: Map<Language, LanguageProperties>;
+  public wordLazyModeSupport: Map<string, boolean>;
```
fehmer: not needed, we already have this in the languageProperties

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -155,6 +157,30 @@ class PseudolangWordGenerator extends Wordset {
...
+
+  constructor(
+    wordsWithLanguage: Map<string, Language>,
+    languageProperties: Map<Language, LanguageProperties>,
+    wordLazyModeSupport: Map<string, boolean>
+  ) {
+    // use the keys as the words array for the base Wordset
+    super(Array.from(wordsWithLanguage.keys()));
```
fehmer: not sure if it is needed, but lets suffle the array here once.

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -687,9 +713,58 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
         throw new WordGenError("");
       }
 
-      const wordSet = languages.flatMap((it) => it.words);
-      Arrays.shuffle(wordSet);
-      return new Wordset(wordSet);
+      // direction conflict check
```
fehmer, line 699: What is the goal here? How is the WordGenError handled?

If all polyglot languages are RTL can we not run the test in RTL mode by e.g. setting the Config.language to the first polyglot language?

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -687,9 +713,58 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
...
+          { duration: 5 }
+        );
+        throw new WordGenError("");
+      }
+
+      // build array of {word, language, supportsLazyMode}
+      const wordEntries: {
+        word: string;
+        language: Language;
+        supportsLazyMode: boolean;
```
fehmer: not needed, see above

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -687,9 +713,58 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
...
+          ligatures: lang.ligatures,
+          rightToLeft: lang.rightToLeft,
+        });
+        const supportsLazyMode = !lang.noLazyMode;
+        for (const word of lang.words) {
+          wordEntries.push({ word, language: lang.name, supportsLazyMode });
+        }
+      }
+      // shuffle the array
+      Arrays.shuffle(wordEntries);
```
fehmer: shuffle should do nothing, as maps are not sorted

frontend/src/ts/test/funbox/funbox-functions.ts
```diff
@@ -687,9 +713,58 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
...
+        throw new WordGenError("");
+      }
+
+      // build array of {word, language, supportsLazyMode}
+      const wordEntries: {
+        word: string;
+        language: Language;
+        supportsLazyMode: boolean;
+      }[] = [];
+      const languageProperties = new Map<Language, LanguageProperties>();
```
fehmer: i think we can simplify this 

```typescript
 const languageProperties = new Map(languages.map((lang) => [lang.name, lang]));
 const wordsWithLanguage = new Map(
        languages.flatMap((lang) => lang.words.map((word) => [word, lang.name]))
 );

 return new PolyglotWordset(wordsWithLanguage, languageProperties);
```

frontend/src/ts/test/test-logic.ts
```diff
@@ -456,14 +457,49 @@ export async function init(): Promise<void | null> {
...
+    // in polyglot mode, check if any of the selected languages support lazy mode
+    const polyglotLanguages = Config.customPolyglot;
+
+    // load all selected polyglot languages to check their lazy mode support
+    const languagePromises = polyglotLanguages.map(async (langName) => {
+      try {
+        const lang = await JSONData.getLanguage(langName);
+        return lang;
+      } catch {
+        return null;
```
fehmer: why do we suppress the error?

frontend/src/ts/test/words-generator.ts
```diff
@@ -380,9 +381,18 @@ async function applyBritishEnglishToWord(
...
 
 function applyLazyModeToWord(word: string, language: LanguageObject): string {
-  const allowLazyMode = !language.noLazyMode || Config.mode === "custom";
-  if (Config.lazyMode && allowLazyMode) {
-    word = LazyMode.replaceAccents(word, language.additionalAccents);
+  // in polyglot mode, check the lazy mode support for the specific word
+  if (polyglotWordLazyModeSupport && polyglotWordLazyModeSupport.has(word)) {
+    const supportsLazyMode = polyglotWordLazyModeSupport.get(word);
+    if (supportsLazyMode && Config.lazyMode) {
+      word = LazyMode.replaceAccents(word, language.additionalAccents);
```
fehmer: we need to pass in the actual language from the polyglot word here to use the correct additionalAccents

frontend/src/ts/test/wordset.ts
```diff
@@ -10,6 +10,7 @@ export class Wordset {
   length: number;
   orderedIndex: number;
   shuffledIndexes: number[];
+  public isPolyglot = false;
```
fehmer: superclass should not know about subclasses. Can't we not  use   `instanceof PolyglotWordset` if we want to detect it?

fehmer (changes requested):
hi @byseif21 , nice work so far. Please check my comments

frontend/src/ts/test/funbox/funbox-functions.ts
byseif21: why do you think not needed?  Wordset expects an array, not a map right?

fehmer: I was talking about the shuffle, not the convert to array.

frontend/src/ts/test/funbox/funbox-functions.ts
byseif21: You mean 
```ts
const languageProperties = new Map(
  languages.map((lang) => [
    lang.name,
    {
      noLazyMode: lang.noLazyMode,
      ligatures: lang.ligatures,
      rightToLeft: lang.rightToLeft,
    },
  ])
);

const wordsWithLanguage = new Map(
       languages.flatMap((lang) => lang.words.map((word) => [word, lang.name]))
);

return new PolyglotWordset(wordsWithLanguage, languageProperties);
```
right ?

frontend/src/ts/test/test-logic.ts
byseif21: log it or notify the user?

frontend/src/ts/test/wordset.ts
byseif21: can we use type guard instead ?

frontend/src/ts/test/test-logic.ts
fehmer: like [this](https://github.com/monkeytypegame/monkeytype/blob/e50d5e9cdd3ceee616d2b3310f42e7111822cb9b/frontend/src/ts/test/test-logic.ts#L431) ig

frontend/src/ts/test/words-generator.ts
```diff
@@ -380,6 +381,20 @@ async function applyBritishEnglishToWord(
...
+  // if currentWordset is a PolyglotWordset, use the word's actual language for lazy mode
+  if (currentWordset && currentWordset instanceof PolyglotWordset) {
+    const langName = currentWordset.wordsWithLanguage.get(word);
+    const langProps = langName
+      ? currentWordset.languageProperties.get(langName)
+      : undefined;
+    const allowLazyMode =
+      (langProps && !langProps.noLazyMode) || Config.mode === "custom";
+    if (Config.lazyMode && allowLazyMode) {
+      word = LazyMode.replaceAccents(word, language.additionalAccents);
```
fehmer: should that be
`word = LazyMode.replaceAccents(word, langProps.additionalAccents);`

Miodec (dismissed):
Thanks

PR #6673, opened 2025-06-27
impr(test-screenshot): Show download screenshot button when holding shift on result page (@raaid3)

raaid3:
### Description

Previously, if you shift clicked on the `copy screenshot` button, you could download a screenshot. But many users probably don't know about this because it's not mentioned anywhere on the result page and there are no visual indicators hinting at there being such a feature.

So I added a "(shift click to download)" message to the `aria-label` of the `#saveScreenshot` button. I also added the ability for the icon inside the button to switch to `fa-download` when shift is held as a visual indicator that clicking the button would download a screenshot.

### DIFF
--- a/backend/src/api/controllers/result.ts
+++ b/backend/src/api/controllers/result.ts
@@ -488,13 +488,16 @@ export async function addResult(
 
   let dailyLeaderboardRank = -1;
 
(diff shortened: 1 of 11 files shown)


Miodec:
Thanks

PR #6675, opened 2025-06-29
fix(input-controller): update keymap UI on backspace in layout emulator (@47th)

47th:
### Description

Fixes the keymap UI not highlighting the current letter to be typed when backspaced to a previous word in layout emulator.

https://github.com/user-attachments/assets/12ba72d8-567b-4097-bbf6-407814140545


Miodec:
Thanks

PR #6694, opened 2025-07-04
impr(ligatures): always enable in Zen and Custom modes for always correct script rendering (@byseif21)

byseif21:
### Description

fixes broken ligature rendering in zen mode by always enabling ligatures, ensuring that scripts like Arabic etc .. display correctly and words connect as expected in any scenario or direction!.


Miodec (approved):
Thanks

PR #6695, opened 2025-07-06
impr(caret): handle mixed language direction (@byseif21)

byseif21:
### Description
enhances the caret positioning logic to support mixed language directions (LTR and RTL) within words. It introduces a new hasRTLCharacters utility function to detect RTL characters in individual words, allowing the caret to adjust dynamically based on word-specific direction rather than relying solely on the language's default direction

#### notes:
* tested no affect to the normal single direction.
* no tap mode handle included 
 * related #6694 #6666


Miodec:
Also, i noticed rtl words are not correctly "connected". And even though  we are chcking if its rtl, the caret still starts on the left side:

image

Miodec:
Thanks

PR #6701, opened 2025-07-07
fix(test): funbox does not remember settings (@fehmer)

### LINKED ISSUE #6700: Simon Says keymap setting persists

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

MonkeyPookie

### Account config

_No response_

### Current Behavior

When enabling the Simon Says funbox via the settings page and disabling it via the command line, the keymap setting remains at "next" instead of reverting to its previous value.

### Expected Behavior

The keymap setting should properly revert to its value prior to enabling and disabling the funbox.

### Steps To Reproduce

1. Go to the settings page
2. Enable the Simon Says funbox in behavior
3. Return to the test page
4. Disable the simon says funbox via the command line or the button next to test wordset
Keymap will be set to "next" mode, regardless of previous settings.

### Environment

- OS: macOS 15.5
- Browser: Google Chrome
- Browser Version: 138.0.7204.93 (Official Build) (arm64)

### Anything else?

_No response_

fehmer:
`funbox.applyConfig` was called twice when changing to test page. `funbox.rememberSettings` was saving the wrong setting, as it was called after the first `applyConfig`.

fixes #6700

### DIFF
--- a/frontend/src/ts/pages/test.ts
+++ b/frontend/src/ts/pages/test.ts
@@ -33,7 +33,6 @@ export const page = new Page({
       noAnim: true,
     });
     void TestConfig.instantUpdate();
-    void Funbox.activate();
     void Keymap.refresh();
     ScrollToTop.hide();
   },


Miodec:
Thanks

PR #6710, opened 2025-07-08
fix(leaderboard): use minTimeTyping from configuration (@fehmer)

fehmer:
We added `minTimeTyping` to the configuration but it was only used to create the indicies.


Miodec (approved):
Thanks

PR #6714, opened 2025-07-09
fix(zen): remove input length check to prevent jump to hidden line (@byseif21, @NadAlaba)

### LINKED ISSUE #6697: Viewport visibility bug in Zen Mode wrapping to a hidden 3rd line

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

when typing if a word wraps near the end of the 2nd line, the caret jumps to a 3rd line which is not visible on screen, the text continues on a third line — but the viewport doesn’t scroll to reveal it.

As a result, the user continues typing on a hidden line, receiving no visual feedback, since the viewport remains locked to the first two lines.

### Expected Behavior

should all always remain visible

### Steps To Reproduce

1- Enable Zen Mode

2- Start typing a long sentence

3- Once you're near the end of the second line, type a word that ends two letters before the end of the line 

4- Press space — then type a big width letter, the caret jumps to the third line that is no longer visible on screen

### Environment

- OS:
- Browser:
- Browser Version:

### Anything else?

https://github.com/user-attachments/assets/be8b35b3-635d-4873-af86-d609f3a54196

byseif21:
Closes #6697

### DIFF
--- a/frontend/src/ts/controllers/input-controller.ts
+++ b/frontend/src/ts/controllers/input-controller.ts
@@ -731,10 +731,19 @@ function handleChar(
   //stop the word jump by slicing off the last character, update word again
   // dont do it in replace typos, because it might trigger in the middle of a wrd
   // when using non monospace fonts
+  /**
+   * NOTE: this input length > 1 guard, added in commit bc94a64,
+   * aimed to prevent some input blocking issue after test restarts.
+   *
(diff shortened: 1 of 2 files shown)


byseif21:
~~closed fixed in 9a6fe31a62231e6e4329bee9306a009c71698169~~

PR #6725, opened 2025-07-10
feat(ui): add hover effect to Monkeytype logo (@saurav0311)

saurav0311:
Previously, there was no hover effect on the Monkeytype logo in the dashboard.
This PR adds a subtle hover effect to enhance UX.


Miodec:
Thanks, but this change was not discussed - I don't really like how it looks.

PR #6730, opened 2025-07-12
fix(docker): use frontend url for firebase admin (@fehmer)

### LINKED ISSUE #6728: Docker: forgot password email not working
When trying to send password reset email on self hosted docker firebase attempts to send for the url of the main site

https://github.com/monkeytypegame/monkeytype/blob/master/backend/src/utils/auth.ts#L99-L105

fehmer:
fixes #6728


fehmer:
i guess we want this to be a config as well https://github.com/monkeytypegame/monkeytype/blob/master/backend/src/init/email-client.ts#L105

PR #6741, opened 2025-07-16
feat(funbox): add no quit (@miodec)

Miodec:
Brrrr


packages/funbox/src/list.ts
```diff
@@ -461,6 +461,12 @@ const list: Record<FunboxName, FunboxMetadata> = {
     name: "asl",
     cssModifications: ["words"],
   },
+  no_quit: {
+    description: "You can't restart the test.",
+    canGetPb: true,
+    difficultyLevel: 1,
```
fehmer: difficultyLevel 0? this does not make the test harder

fehmer (dismissed):

PR #6744, opened 2025-07-17
fix(account-history): highlight animation is not working on result selection (@byseif21)

byseif21:
### Description

fix the highlight animation for the account history table (when clicking a chart dot) was not visible or only worked inconsistently. This was because the animation was applied to the <tr> element, but the background color for table rows is set on the <td> elements. As a result, the animation was hidden by the static background color of the table cells.

Closes #

### DIFF
--- a/frontend/src/ts/pages/account.ts
+++ b/frontend/src/ts/pages/account.ts
@@ -1156,17 +1156,19 @@ $(".pageAccount #accountHistoryChart").on("click", () => {
   const windowHeight = $(window).height() ?? 0;
   const offset = $(`#result-${index}`).offset()?.top ?? 0;
   const scrollTo = offset - windowHeight / 2;
-  $([document.documentElement, document.body]).animate(
-    { scrollTop: scrollTo },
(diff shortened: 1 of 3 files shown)


frontend/src/ts/pages/account.ts
```diff
@@ -1163,7 +1163,9 @@ $(".pageAccount #accountHistoryChart").on("click", () => {
     Misc.applyReducedMotion(500)
   );
   $(".resultRow").removeClass("active");
-  $(`#result-${index}`).addClass("active");
+  setTimeout(() => {
```
fehmer: can we move this to the jquery animate instead?

```typescript
 $([document.documentElement, document.body]).animate(
    { scrollTop: scrollTo },
    {
      duration: Misc.applyReducedMotion(500),
      done: () => {
        element.addClass("active");
      },
    }
  );
```

Miodec (approved):
Thanks

Miodec:
Made a few style changes.

PR #6745, opened 2025-07-18
fix(themes): clipped logo letters in some themes (@byseif21)

byseif21:
### Description

all themes that use `background-clip: text` should have the `padding-bottom: 0.1em` to prevent the cut logo letters.


Miodec:
This doesnt seem to be the right fix, it moves the text out of alignment.
image

Issue #6742, opened 2025-07-16
Error while log-out on ape-keys page


fehmer:
- login
- open https://monkeytype.com/account-settings?tab=apeKeys
- logout

results in `Error getting ape keys: Unauthorized`

hussain-alsayedali:
Could you assign it to me

PR #6746, opened 2025-07-18
fix: fixed the bug where Error getting ape keys: Unauthorized happens when user logs out(@hussain-alsayedali)

hussain-alsayedali:
### Description

The getData function will check if the user is authenticated before being called, as the issue seems with updateUI function begin called some times when the user logs out
monkeytype\frontend\src\ts\pages\account-settings.ts

Tried to replicate the issue but it only happens once every while, don't know why it don't happen everytime.

Issue : 
https://github.com/monkeytypegame/monkeytype/issues/6742

### DIFF
--- a/frontend/src/ts/elements/account-settings/ape-key-table.ts
+++ b/frontend/src/ts/elements/account-settings/ape-key-table.ts
@@ -4,7 +4,7 @@ import Ape from "../../ape";
 import { ApeKey, ApeKeys } from "@monkeytype/contracts/schemas/ape-keys";
 import { format } from "date-fns/format";
 import { SimpleModal, TextArea } from "../../utils/simple-modal";
-
+import { isAuthenticated } from "../../firebase";
 const editApeKey = new SimpleModal({
   id: "editApeKey",
   title: "Edit Ape key",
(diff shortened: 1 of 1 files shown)


frontend/src/ts/elements/account-settings/ape-key-table.ts
```diff
@@ -161,8 +161,10 @@ const element = $("#pageAccountSettings .tab[data-tab='apeKeys']");
 
 async function getData(): Promise<boolean> {
   showLoaderRow();
+  if (!isAuthenticated()) {
+    return false;
+  }
   const response = await Ape.apeKeys.get();
-
```
byseif21: could we make it like this for consistency and leave the blank lines, a bit good for readability even up there
```typescript
async function getData(): Promise<boolean> {
  if (!isAuthenticated()) return false;
  
  showLoaderRow();
  const response = await Ape.apeKeys.get();

  if (response.status !== 200) {
```

Miodec: Yeah i like that more, @hussain-alsayedali please make the change.

hussain-alsayedali: I agree, I just changed it

Miodec (approved):
Thanks

hussain-alsayedali:
Thanks (:

PR #6748, opened 2025-07-18
feat(language): add Arabic common words file (@iky95)

iky95:
This pull request adds a new Arabic language file to Monkeytype, containing 200 commonly used Arabic words, all prefixed with "ال" for consistency and realism.

The file includes:
- Right-to-left support (`rightToLeft: true`)
- Ligature rendering (`ligatures: true`)
- Proper language tag (`bcp47: "ar"`)

This variation offers a more natural experience for Arabic speakers, especially in casual and regional contexts. It complements the existing Arabic file by focusing on high-frequency, unvoweled words with the definite article.

Suggested filename: `arabic_common.json` or `arabic_al.json` to avoid conflict with the existing `arabic.json`.


byseif21:
hay @iky95 Thank you for your contribution

Could you please clarify the motivation behind adding this word list? I don't think a word list focuses on words prefixed with "ال" is a new language to add, , not sure why this needs to be a separate entry rather than part of the other word lists as it's already now. I think this meant more to be as a custom set maybe, clarify me if I'm missing something!

Miodec:
Yeah, whats the difference between this and the normal file?

Miodec:
Gonna close due to inactivity, feel free to comment.

PR #6753, opened 2025-07-19
refactor(config): config metadata (@miodec)

Miodec:
Create config metadata object
Move all the special code on config change to config listeners
Create a generic set function which will work with the metadata object to update any config key
Update all setters to use the generic set.
(Later probably only use the generic settter and remove all the specific ones)
Also orders config groups and config schema.


frontend/src/ts/config.ts
```diff
@@ -102,199 +110,780 @@ function isConfigChangeBlocked(): boolean {
...
+  },
+
+  // other (hidden)
+  accountChart: {
+    displayString: "account chart",
+    changeRequiresRestart: false,
+    overrideValue: (value, currentValue) => {
+      // if both speed and accuracy are off, set opposite to on
+      // i dedicate this fix to AshesOfAFallen and our 2 collective brain cells
+      if (
```
fehmer: maybe?
```ts
if(value[0]==="off" && value[1]==="off"){
  const changedIndex = value[0]===config.accountChart[0]?1:0;
  value[changedIndex]="on";
}
```

PR #6756, opened 2025-07-19
fix(german_250k.json): remove non-german accents (@8e3)

8e3:
Removed non-German accents / untypable loanwords

### Description

Since the German qwertz layout doesn't support certain accents, used in some loanwords in this list, typing them is impossible, without changing your keyboard layout mid-session. Using an automated script, I decided to remove them, enhancing playability.


Miodec:
Thanks

PR #6767, opened 2025-07-22
fix(language): remove offensive & some misspellings from english_450k (@byseif21)

byseif21:
not a full cleanup 
Screenshot 2026-03-27 144457

if we decided one day to clean it all here is a reference of what needs to be [cleaned up](https://github.com/monkeytypegame/monkeytype/discussions/6682#discussioncomment-16182662)


maueroats:
Looking at the diff, this PR looks like it contains significant errors. There are many changes which I recognize as deleting real English words.

Examples:
* "abcess" is deleted.
* "Alan" is changed to "alan"

byseif21:
> Looking at the diff, this PR looks like it contains significant errors. There are many changes which I recognize as deleting real English words.
> 
> Examples:
> 
>     * "abcess" is deleted.
> 
>     * "Alan" is changed to "alan"

Hi @maueroats 
Thanks for the feedback,

I’m not a native English speaker, so I always try to double-check things using tools and manual searches, but I know it’s still easy to miss stuff, especially with proper nouns or edge cases like "Alan" case here.

Regarding “abcess,” I think that was a typo (should be “abscess”) same as the most of the removals were intended to clean up typos,
If you found more that you think should’ve been stayed in, feel free to point them out — that would really help. Thanks again!

maueroats:
##  Notes
* "Ace" -> "ace". Both capital (name) and lowercase exist in English.
* (6432): "affuseddaffusing" should be split into two words, not deleted "affused" "affusing"
* "Aleck" could be an uncommon proper name. The lower case aleck is used in the idiom "smart aleck" only (as far as I know), but should be included.
* (9705) "aliyahaliyahs" -> "Aliyah" "Aliyah's" but I think removing these is ok.
* (9706) "aliyahs" "aliyas" - what do these lower case words mean?  Questioning. They sound like possessive versions of proper nouns.
* "AWOLs" sounds like a fake word. One soldier who is *absent without leave* is AWOL. Two soldiers are AWOLs? I don't think so, they are still AWOL. Making a plural of the word "leaves" isn't appropriate here.

## Confirmed Correct Edits

* abcissa - correctly removed, abscissa is present
*   "acappella" - I think this needs to be two words "_a cappella_" to be correct usage, so I support its deletion.

PR #6772, opened 2025-07-23
impr(quotes): polish quotes and law of physics for polish language (@sobczaktm)

sobczaktm:
Description
polish quotes and law of physics for polish language

Checks
Translations:
I will never serve any king or emperor, only my homeland. - "id":231

Nothing in life is to be feared, it is only to be understood. - "id":232

Until I used the Internet, I didn't know there were so many idiots in the world. - "id":233

Mathematics is the only field in which nothing can be hidden.- "id":234

It is not enough to speak to the point, you also need to speak to people. - "id":235

Mathematics is the most beautiful language in the world. - "id":236

Science should serve humanity, not its destruction. - "id":237

If no force acts on a body or the forces acting on it are balanced, the body remains at rest or continues in uniform rectilinear motion. - "id":238

If a constant force acts on a body, it moves with uniform acceleration, and the acceleration is directly proportional to the force acting on the body and inversely proportional to the mass of the body. - "id":239

Any two bodies attract each other with a force that is proportional to the product of their masses and inversely proportional to the square of the distance between them. -"id":240

If body A exerts a force on body B, then body B exerts a force on body A of equal magnitude but in the opposite direction. - "id":241

The current in a conductor is proportional to the voltage and inversely proportional to the resistance. - "id":242

Energy cannot be created or destroyed - it can only change form. - "id":243

A body immersed in a liquid is subjected to a buoyant force equal to the weight of the displaced liquid. - "id":244

The electrostatic force between two charges is proportional to the product of the magnitudes of these charges and inversely proportional to the square of the distance between them.- "id":245


sobczaktm:
Sorry for the confusion with the PR.

Miodec (approved):
Thanks

PR #6751, opened 2025-07-19
impr: add validations to settings input (@fehmer)

fehmer:

PR #6775, opened 2025-07-24
refactor(config): cleanup handling for fontSize, tapeMargin and maxLineWidth (@fehmer)

fehmer:
merge this after #6751 and the command builder pr

- remove configOverride
- add handling to config migration


Miodec (approved):
Thanks

PR #6781, opened 2025-07-26
impr(settings): reset input value when pressing escape (@fehmer)

fehmer:


fehmer:
@Miodec  and me agreed that this is not wanted

PR #6788, opened 2025-07-27
fix(commandline): missing initialize input validation handler for commandOverride (@byseif21, @fehmer)

byseif21:
### Description

commands opened directly via `commandOverride` (e.g., `changeCustomPolyglot` when triggered by a text button after a page refresh) were not correctly initializing their input validation handlers.


Miodec (approved):
Thanks

PR #6791, opened 2025-07-27
fix(login): clear data on logout (@fehmer)

### LINKED ISSUE #6789: Logout issue when using multiple tabs
when being logged in in two tabs with the same user a logout in one tab does logout the user in the other tab, but the header is still showing the user profile/level and the snapshot data is not cleared

fehmer:
fixes #6789


Miodec (approved):
Thanks

PR #6792, opened 2025-07-28
fix: changing the font size would display incorrect number of lines (@byseif21)

byseif21:
### Description

after the recent font refactor, changing the font size  now would cause the UI to display wrong number of lines till the test restarted.

The `updateWordsWrapperHeight()` function was being called synchronously immediately after the font size CSS was applied, but before the browser could recalculate element dimensions with the new font metrics ,this fix it.


fehmer:
Hi, can you please provide steps how to reproduce/test it and what to look out for, thanks

Miodec:
This is a bug i introduced, just pushed the actual fix. Thanks for he PR though.

PR #6796, opened 2025-07-28
fix: filter section doesn't toggle using local image (@byseif21)

byseif21:


Miodec (changes requested):
Filters are not visible now when using local image

Miodec (approved):
Im stupid i wasnt testing on your pr. All good.

PR #6651, opened 2025-06-22
fix(focus mode): remove cursor when in hovering over links and buttons (@jaydent, @miodec)

jaydent4:
### Description

Changed `("body").css("cursor", "none")` to `("*").css("cursor", "none")`  to fix cursor staying when in focus mode.

Closes #6644

PR #6801, opened 2025-07-29
fix: cursor not changing to pointer on button hover (@NadAlaba)

nadalaba:
button

- bug was introduced in [#6651](https://github.com/monkeytypegame/monkeytype/pull/6651) by changing the inline style `cursor` to `default` when focus mode is disabled instead of removing the `cursor: none` inline style.
- add a `.clickable` class to clickable banners/PSAs instead of relying on `cursor: pointer` inline style which may be cleared by `.css("cursor", "")` in the previous commit.
- make the test for clickable banners/PSAs case insensitive and allow `<a>` elements in the test.

### DIFF
--- a/frontend/src/ts/elements/notifications.ts
+++ b/frontend/src/ts/elements/notifications.ts
@@ -193,7 +193,7 @@ class Notification {
         });
       }
       // NOTE: This need to be changed if the update banner text is changed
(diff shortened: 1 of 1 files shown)


Miodec (approved):
Thanks

PR #6808, opened 2025-07-30
test: use mongodb testcontainer (@fehmer)

fehmer:


backend/__tests__/__integration__/dal/user.spec.ts
fehmer, line 1: change to use dynamic uids instead of fixed ones

PR #6816, opened 2025-07-31
impr(commandline): add local background (@byseif21)

byseif21:


fehmer:
could you add the same for local font? thanks

Miodec:
Thanks

PR #6817, opened 2025-08-01
fix(font): 0xProto not working in screenshot (@byseif21)

byseif21:
### Description
starting with number seem to cuzes issue, did some hacking idk

### DIFF
--- a/frontend/src/ts/commandline/commandline-metadata.ts
+++ b/frontend/src/ts/commandline/commandline-metadata.ts
@@ -586,11 +586,11 @@ export const commandlineConfigMetadata: CommandlineConfigMetadataObject = {
   },
   fontFamily: {
     subgroup: {
-      options: typedKeys(Fonts).sort((a, b) => {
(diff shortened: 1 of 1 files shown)


fehmer:
I like the hack, but we need to migrate existing configs to the new value
https://github.com/monkeytypegame/monkeytype/blob/6f6af5e622fe99da05acc4158ec65f41348e21d9/backend/src/utils/result.ts#L74

frontend/src/ts/constants/fonts.ts
```diff
@@ -133,7 +133,8 @@ export const Fonts: Record<KnownFontName, FontConfig> = {
   Iosevka: {
     fileName: "Iosevka-Regular.woff2",
   },
-  "0xProto": {
+  OxProto: {
```
fehmer: maybe just rename to  `Proto` having `0xProto` and `OxProto` is confusing and error prone

fehmer (changes requested):
please check my comments

PR #6818, opened 2025-08-01
fix(quote-mode): broken all selection and visual highlighting issues (@byseif21)

byseif21:
### Description

* fixed UI click handler to consistently map "all" quote mode selection (all) to [0, 1, 2, 3] `setQuoteLengthAll();`.
* fix the mobile model to highlight the all selection when length is [0, 1, 2, 3].
* fix quote length options were not visually highlighted when refreshing from other mode then back to quote mode.

### DIFF
--- a/frontend/src/ts/test/test-config.ts
+++ b/frontend/src/ts/test/test-config.ts
@@ -56,6 +56,7 @@ export async function instantUpdate(): Promise<void> {
     $("#testConfig .customText").removeClass("hidden");
   }
 
+  updateActiveExtraButtons("quoteLength", Config.quoteLength);
   updateActiveExtraButtons("numbers", Config.numbers);
   updateActiveExtraButtons("punctuation", Config.punctuation);
 


fehmer:
We could  set data-quote-length to "all" and call https://github.com/monkeytypegame/monkeytype/blob/6f6af5e622fe99da05acc4158ec65f41348e21d9/frontend/src/ts/config.ts#L569 in that case

frontend/src/ts/test/test-logic.ts
```diff
@@ -1470,24 +1470,24 @@ $(".pageTest").on("click", "#testConfig .time .textButton", (e) => {
...
 
-    if (UpdateConfig.setQuoteLength(arr, false)) {
-      ManualRestart.set();
-      restart();
-    }
+    if (!UpdateConfig.setQuoteLength(arr, false)) return;
   }
+  ManualRestart.set();
+  restart();
 });
```
Miodec, line 1497: I find this 'always return but block it with return' flow kinda weird to read. Could you revert to the previous approach and restart if the set function returned true?

byseif21: I was just aiming for less lines code sorry

byseif21:
now the `all` selection highlights the only `all` selection for both.

Miodec (approved):
Thanks

PR #6803, opened 2025-07-30
fix(settings): dismiss tooltips on funbox buttons after click (@byseif21)

byseif21:
close #6800

PR #6831, opened 2025-08-03
fix(tooltips): prevent hover pop-up reappearance on click (@byseif21)

byseif21:
### Description

tooltip pop-ups were reappearing on click when interacting with various clickable elements (e.g., tags in results table, test result buttons, account page actions).

### DIFF
--- a/frontend/src/styles/core.scss
+++ b/frontend/src/styles/core.scss
@@ -466,3 +466,15 @@ key {
   font-size: var(--balloon-font-size);
   line-height: var(--balloon-font-size);
 }
+
+// this here to prevent tooltips from showing when an element is focused via mouse click
+// to avoid reappearing popups after clicks, WITHOUT breaking keyboard nav.
+// so shortly we show balloon tooltips on hover and keyboard focus only. not on mouse focus!
+@supports selector(:focus-visible) {
+  [aria-label][data-balloon-pos]:focus:not(:focus-visible):not(:hover)::before,
+  [aria-label][data-balloon-pos]:focus:not(:focus-visible):not(:hover)::after {
+    opacity: 0 !important;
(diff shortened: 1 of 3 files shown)


fehmer:
any idea what the root cause might be. Feels like we end up adding this to every clickable element with a hover text now: https://github.com/monkeytypegame/monkeytype/pull/6803

byseif21:
> any idea what the root cause might be. Feels like we end up adding this to every clickable element with a hover text now: #6803

Idk i thought it's like this from a while ago , if could you remember seeing them working finely before? if yes then yah I think we should look for a root cause instead of this, I will check some old PR to test on

fehmer:
this commit caused the issue on the funbox buttons:
https://github.com/monkeytypegame/monkeytype/commit/2506b4f88d50d4d8980c533393807f0513a31591

this commit introduced the issue with tags button:
https://github.com/monkeytypegame/monkeytype/commit/e29d280ea52e532930e24531aa41427cbe07bd10

not sure if this helps

fehmer:
Same problem on 
- test result buttons word history and replay
- account page edit profile, copy public link, show personal bests

Miodec (approved):
Thanks

Miodec:
Actually, this might be fixable globally by editing the animated modal class, not a singular modal. Think you could look into that?

But then that would interfere with the keyboard naviagation. I wonder if there is a way to blur visually but  without losing focus?

byseif21:
> Actually, this might be fixable globally by editing the animated modal class, not a singular modal. Think you could look into that?
> 
> But then that would interfere with the keyboard naviagation. I wonder if there is a way to blur visually but without losing focus?

How about that?

Miodec (approved):
Nice, thanks.

PR #6836, opened 2025-08-04
fix: font file type check for better browser compatibility (@byseif21)

byseif21:
### Description

* browsers don't provide accurate MIME types for font files  , I couldn't even reuse the site fonts as local fonts. 
* Validation now checks both MIME type and file extension to ensure valid fonts aren't mistakenly rejected.

### DIFF
--- a/frontend/src/ts/elements/settings/custom-font-picker.ts
+++ b/frontend/src/ts/elements/settings/custom-font-picker.ts
@@ -52,7 +52,10 @@ uploadContainerEl
     }
 
     // check type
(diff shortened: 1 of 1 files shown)


fehmer:
which browser?

byseif21:
> which browser?

it may be all tbh, i tested in chrome and zen, zen based on firefox so i assume firefox too.

fehmer:
i tested with chrome and firefox on linux and it worked well. Thanks for the fix.

fehmer (approved):

PR #6853, opened 2025-08-08
feat: add defensive DOM checks to result stats rendering (@rishabpremish)

rishabpremish:
This pull request refactors the way result statistics are updated in the UI to improve robustness and prevent errors when elements are missing from the DOM. The main changes involve checking for the existence of DOM elements before updating their text content.

Improvements to DOM manipulation and error prevention:

* Updated the `updateWpmAndAcc` function in `frontend/src/ts/test/result.ts` to check for the existence of each relevant DOM element (`.wpm .top .text`, `.wpm .bottom`, `.raw .bottom`, `.acc .bottom`) before attempting to update its text, preventing potential runtime errors if elements are missing.### Description

### Checks

Closes #
(@rishabpremish)

### DIFF
--- a/frontend/src/ts/test/result.ts
+++ b/frontend/src/ts/test/result.ts
@@ -51,9 +51,16 @@ let useUnsmoothedRaw = false;
 let quoteLang: Language | undefined;
 let quoteId = "";
 
+// DOM element cache for performance optimization
+let resultStatsEl: HTMLElement | null = null;
+
 export function toggleUnsmoothedRaw(): void {
   useUnsmoothedRaw = !useUnsmoothedRaw;
   Notifications.add(useUnsmoothedRaw ? "on" : "off", 1);
+}
+
+function initializeResultElements(): void {
+  resultStatsEl = document.querySelector("#result .stats");
 }
 
 let resultAnnotation: AnnotationOptions<"line">[] = [];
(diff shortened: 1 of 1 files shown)


byseif21:
Hi @rishabpremish , is there a scenario where this may happen ? cuz the result stats nodes seems to be always in the static template when updateWpmAndAcc() runs, and jQuery .text() on an empty selection is a no-op. so idk i dont think the checks are needed here?

rishabpremish:
Hi @byseif21

I added the checks because I was thinking about scenarios where the DOM might not be fully rendered or if certain elements were dynamically removed (e.g., due to future changes or A/B tests). In those cases, having the guards could help avoid runtime errors.

If these elements will always be in the static template when updateWpmAndAcc() runs, and since .text() on an empty selection is indeed a no-op, I’m fine with removing the checks to keep the code clean.

fehmer:
hi @rishabpremish ,

thank you for your contribution. I agree with @byseif21 and don't feel it is needed to handle missing elements on the page we put there in the same file/page.

Miodec:
> hi @rishabpremish ,
>
> thank you for your contribution. I agree with @byseif21 and don't feel it is needed to handle missing elements on the page we put there in the same file/page.

Yeah, now that i look at this im not sure its needed because the test page (and the result element) is always in the dom, its never removed. Even if the test page is not active:

image

PR #6860, opened 2025-08-09
fix(test): immediately reposition caret and tape on tapeMargin change (@byseif21)

byseif21:
### Description

changing `tapeMargin` mid-test didn’t move the caret until typing resumed because no layout update was triggered for this config.


Miodec (approved):
Thanks

PR #6861, opened 2025-08-09
test: remove duplicate config tests (@fehmer)

fehmer:
Somehow we ended up with the same tests in config and config-metadata


Miodec:
Thanks

PR #6863, opened 2025-08-10
fix(zen-mode): restore input history and watch replay (@byseif21)

byseif21:
### Description

fix:
* The result “input history” panel sometimes rendered nothing because the renderer attempted to split an undefined target word.
added split word ?? "" to handle missing target words in zen.

* The replay view in zen mode could be empty if the last word was just submitted (so input.current was empty at finish), and the replay words list never got updated.


Miodec (approved):

Miodec:
Thanks

PR #6871, opened 2025-08-11
test: don't use globals for vitest (@fehmer)

fehmer:
- **test: support to run/watch all tests in vscode (@fehmer)**
- **packages/contracts**
- **move schema tests to schema package**
- **packages/funbox**
- **packages/utils**
- **frontend**
- **backend**


Miodec (dismissed):
Thanks

PR #6883, opened 2025-08-15
feat(language): add hawaiian,200,1k (@SirObby)

SirObby:
### Description

Added the hawaiian language, 200 and 1k variants. I used the "freqlist_haw.txt" file from https://dohliam.github.io/corpus/haw/ which sorts by frequency. 

### Checks

- [x] Adding a language?
  - [x] Add language to `packages/schemas/src/languages.ts`
  - [x] Add language to exactly one group in `frontend/src/ts/constants/languages.ts`
  - [x] Add language json file to `frontend/static/languages` 


SirObby:
Can a maintainer look at this? Looks like prettier is tripping because I definitely did run prettier on the file in one of the later commits.

frontend/src/ts/constants/languages.ts
Miodec, line 1: Unnecessary changes, please revert.

packages/schemas/src/languages.ts
Miodec, line 1: Try to order this list roughly based on how popular the languages are.

SirObby: I don't understand this comment. "Try to order this list roughly based on how popular the languages are." I have only appended two lines to the end of the list, "hawaiian" and "hawaiian_1k". Since these languages were not in the website I cannot guess whether or not the languages are propular. If you are referring to the popularity of it as a spoken language, I must mention hawaiian is an endangered language.

SirObby:
I am pretty sure I did add it to a group? I was following the guide on docs/ for adding a language.

`hawaiian: ["hawaiian", "hawaiian_1k"],`

packages/schemas/src/languages.ts
Miodec: I mean insert the languages next to languages of similar popularity. Definitely at least above code languages. Do the same for groups.

SirObby: I moved the hawaiian languages above the code languages. However as previously stated I am unsure of the location where this should be "popularity"-wise. I want to note that the contributing docs only mentions adding (which implies appending) and doesn't mention a specific location.

Miodec (approved):
Thanks

PR #6884, opened 2025-08-15
chore: switch screenshot library from html2canvas to modern-screenshot (@byseif21)

byseif21:


frontend/src/ts/test/test-screenshot.ts
```diff
@@ -120,32 +118,100 @@ async function generateCanvas(): Promise<HTMLCanvasElement | null> {
...
+    const targetHeight = Math.max(
+      document.documentElement.scrollHeight,
+      document.documentElement.clientHeight
+    );
 
-    const canvas = await (
-      await gethtml2canvas()
-    )(document.body, {
+    // Target the HTML root to include .customBackground
+    const dataUrl = await domToPng(document.documentElement, {
```
fehmer: Can we use  `domToCanvas` instead of  `domToPng` ?

I measured 1321 ms with png and 1147 ms canvas

fehmer: I measured my old approach (copy the background image to the  `#result` div and run `domToCanvas` on the `#result`) and it took 436 ms average. 

I  think the problem is that modern-screenshot makes a copy of all child nodes even if they are  not  visible. I managed  to speed up the  screenshot to ~600  ms by adding this filter:

```ts
      filter: (el: Node) => {
        if (el instanceof HTMLElement) {
          if (el.classList.contains("hidden")) {
            return false;
          }
        }
        return true;
      },
```

i measured Html2Canvas with ~630ms

byseif21: nice just tried it, that fixed other things too, thanks

fehmer: thinking more about this it might be a better approach to create a new div, copy over the background.img element and move the #result into it and use the new div parameter into modern-screenshot. this way we would have free control over the styling and changes on the rest  of the page outside the #result would not effect the screenshot, making it more resiliant?!

PR #6887, opened 2025-08-17
fix: add comma and period to specials funbox (#6870) (@radouane-tamouss)

### LINKED ISSUE #6870: The "specials" funbox doesn't generate the '.' or ',' symbols

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

bogomips

### Account config

_No response_

### Current Behavior

The `specials` array in the `getSpecials()` function in `/ts/utils/generate.ts` doesn't contain the `'.'` or `','` characters, so these never appear in the generated sequences. Is that intentional?

### Expected Behavior

I'd like to practice in the `specials` funbox with those symbols included.

### Steps To Reproduce

1. Activate the `specials` funbox.
2. No full stops or commas in sight.

### Environment

- OS: any
- Browser: any
- Browser Version: any

### Anything else?

_No response_

radouane-tamouss:
fix: add comma and period to specials funbox (#6870) (@radouane-tamouss)

- Add missing ',' and '.' characters to specials array in getSpecials()
- Add comprehensive tests for getSpecials() function
- Fixes issue where comma and period never appeared in specials funbox

Closes #6870

### Description

This PR fixes issue #6870 where the "specials" funbox was missing comma (,) and period (.) characters. Users reported that these common punctuation marks never appeared when using the specials funbox, making it impossible to practice typing them in that mode.

**Changes made:**
1. Added ',' and '.' to the `specials` array in the `getSpecials()` function
2. Added comprehensive unit tests to verify the fix works correctly and prevent future regressions
3. Tests ensure all special characters (including comma and period) can be generated

**Testing:**
- Added unit tests that specifically verify comma and period are included
- Tests run multiple iterations to account for randomness in character generation
- Verified all existing functionality remains unchanged

### Checks

Closes #6870


frontend/__tests__/utils/generate.spec.ts
```diff
@@ -14,3 +14,90 @@ describe("hexadecimal", () => {
...
+        if (char === ".") foundPeriod = true;
+      }
+
+      // Early exit if we found both
+      if (foundComma && foundPeriod) break;
+    }
+
+    expect(foundComma).toBe(true);
+    expect(foundPeriod).toBe(true);
+  });
```
Miodec: This seems to be already covered by the last test no? Seems a bit unnecessary.

fehmer: i'd add  _only_ one test, generate 1000  special "words" and check for each 
- min/max length
- does not contain other chars than expectedSpecial (only keep the `expect(expectedSpecials).toContain(char)`, remove the regex match.

keep foundComma and foundPeriod and check at the end.

frontend/__tests__/utils/generate.spec.ts
```diff
@@ -14,3 +14,90 @@ describe("hexadecimal", () => {
     expect(hex).toMatch(/^[0-9a-f]+$/);
   });
 });
+
+describe("specials", () => {
+  const expectedSpecials = [
```
fehmer: instead of duplicating the expectedSpecials  i'd export the array used in the generate.ts (using __testing like we do [here](https://github.com/monkeytypegame/monkeytype/blob/14ce657273f3d92a8769f6d80220e96a612f577c/frontend/src/ts/config.ts#L921) and use it here.

radouane-tamouss: Thank you both for the feedback! I've updated the implementation:

@Miodec - You were right about the redundant test. I've consolidated everything into one comprehensive test.

@fehmer - Implemented all your suggestions:
- Exported `specials` array using `__testing` pattern (no more duplication)
- Single test with 1000 iterations checking min/max length, character validation with `toContain()`, and comma/period detection
- Removed regex validation as requested

I think the code is now cleaner and follows the project's testing standards. Thanks for the guidance! 🙏

radouane-tamouss:
radouane_tamouss

PR #6890, opened 2025-08-17
fix(buildInputCommand): Custom settings not being saved properly when applied through commandline (@raaid3)

raaid3:
### Description

When applying custom settings like paceCaretCustomSpeed and minWpmCustomSpeed, only the custom values are saved to the DB, but not the "modes" like `{paceCaret: "custom"}` or `{minWpm: "custom"}`

So the config being saved when applying the setting would just be: `{paceCaretCustomSpeed: 120}` for example, instead of `{paceCaretCustomSpeed: 120, paceCaret: "custom"}`. This means a simple reload would would fetch the configValue set before this instead of "custom".

https://github.com/user-attachments/assets/44bb64e2-c99b-460a-b0f4-8de6389c0077

To fix this, I add an argument "rootKey" (e.g. paceCaret) to `buildInputCommand` in addition to alread accepting the secondaryKey (e.g. paceCaretCustomSpeed). This way, `genericSet` can be called on the rootKey as well to set it to "custom".

### DIFF
--- a/frontend/src/ts/commandline/util.ts
+++ b/frontend/src/ts/commandline/util.ts
@@ -68,6 +68,7 @@ function _buildCommandForConfigKey<
     const inputProps = commandMeta.input;
 
     const inputCommand = buildInputCommand({
+      rootKey: key,
       key: "secondKey" in inputProps ? inputProps.secondKey : key,
       isPartOfSubgruop: "subgroup" in commandMeta,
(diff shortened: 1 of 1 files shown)


Miodec:
Thanks, however this is not he fix. The actual issue came from the config module and overrideConfig metadata prop. It will be pushed to the live site later today.

Its always good to discuss the fix by either creating an issue or joining the Discord server to avoid this in the future.

PR #6901, opened 2025-08-21
ci(assets): Refactor json-validator (@fehmer)

fehmer:


turbo.json
```diff
@@ -32,7 +32,7 @@
       "dependsOn": ["build"]
     },
     "@monkeytype/frontend#validate-json": {
-      "dependsOn": ["^parallel"]
+      "dependsOn": ["^build"]
```
nadalaba: I think `@monkeytype/schemas#build` is enough, you don't need to build all packages

frontend/scripts/json-validation.ts
```diff
@@ -17,341 +17,356 @@ import { ThemesList } from "../src/ts/constants/themes";
...
+          return `${label}:\n ${(problems as string[])
+            .map((error) => "\t- " + error)
+            .join("\n")}`;
+        })
+        .join("\n")
+    );
+  }
+}
+
+function findDuplicates(words: string[]): string[] {
```
fehmer, line 57: speed up from  ~3s to ~1.3s using Set instead of object

Miodec: Sets go brrr

PR #6908, opened 2025-08-24
fix(test): Lazy mode set via command line does not restart test (@fehmer)

fehmer:


frontend/src/ts/commandline/commandline-metadata.ts
```diff
@@ -331,6 +331,7 @@ export const commandlineConfigMetadata: CommandlineConfigMetadataObject = {
   lazyMode: {
     subgroup: {
       options: "fromSchema",
+      afterExec: () => TestLogic.restart(),
```
fehmer, line 334: should we call `TestLogic.restart()` in genericSet based on config-metadata `changeRequiresRestart` instead?

Miodec: different pr

Miodec (approved):
Thanks

PR #6909, opened 2025-08-25
fix: All \t are converted to tab characters (@Leonabcd123)

### LINKED ISSUE #6906: /t's in saved text files are not getting indented in a single go

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

I use python codes, custom as my saved text files to practice and improve my code typing spead. Since python relies on indentation i considered to have the same behavior practied in the website as well. 

So i have /t on each line of my saved text file to have the indentation key recognized from the file. Once clicked ok on the custom popup/model i can see the "->" indentation being shown in my test. Works.

But the problem is when i have more than one indentation on some lines like /t/t/t<rest of the line>. In this case if i click ok the first /t in the line will only get translated into the test and test will be considered with the line.
Example: ->/t/t<rest of the line>.

### Expected Behavior

The workaround for now is to go back to the custom option and click OK again for that custom text, repeat this until all the /t escape sequence is converted to indentation.

The Expected is to do these translation in a single go by identifying all the escape sequence in the custom text.

### Steps To Reproduce

1. Take a python code or just any line of code
2. Have different number of /t without any spaces, in front of each line
3. Copy it and go to the website
4. Click Custom -> Change
5. Paste it in to the custom text popup in the mokeytype website.
6. Click ok
7. Expected for each /t in the saved file line the same number of indentation key should be expected in the test.

### Environment

- OS: mac
- Browser: Brave
- Browser Version: 1.81

### Anything else?

I Understand this is not a typical usecase of this website. If this request seems insignificant, I apologize for any inconvenience.

Leonabcd123:
### Description

Made it so all \t are converted to tab characters.

### Closes

Fixes #6906

### DIFF
--- a/frontend/src/ts/utils/strings.ts
+++ b/frontend/src/ts/utils/strings.ts
@@ -153,6 +153,12 @@ export function cleanTypographySymbols(textToClean: string): string {
   );
 }
 
+/**
+ * Split a string into characters. This supports multi-byte characters outside of the [Basic Multilinugal Plane](https://en.wikipedia.org/wiki/Plane_(Unicode).
+ * Using  `string.length` and `string[index]` does not work.
+ * @param s string to be tokenized into characters
+ * @returns array of characters
+ */
 export function splitIntoCharacters(s: string): string[] {
   const result: string[] = [];
   for (const t of s) {


Miodec:
This PR is a regression because you won't be able to escape the `\t` character with `\\t` to type the literal `\t` text. Please make sure your changes don't affect existing functionality.

Miodec:
Do you think it would be a good idea to extract this functionality to the strings util file and add some tests? To make sure with `\\t\\t\\t` we get literal `\t\t\t` correctly?

Leonabcd123:
Do you want to move it to frontend/src/ts/utils/strings.ts?

Miodec:
Yeah. Call it `replaceControlCharacters`

Miodec (approved):
Thanks

PR #6911, opened 2025-08-25
chore: Changed punctuation (@Leonabcd123)

Leonabcd123:
### Description

Changed punctuation in the about page. We can also make the text in the parenthesis a standalone sentence by just removing the parenthesis.


Miodec (approved):
Thanks

PR #6921, opened 2025-08-28
feat(layout): add handsdown neu, and neu inverted (@aleksiye)

aleksiye:
### Description

Added two layouts, handsdown neu and it's inverted variation that swaps rows 2 and 4.  Layout as seen [here](https://sites.google.com/alanreiser.com/handsdown/home/hands-down-neu)


aleksiye:
Failed prettier check, tried fixing it by running prettier in the affected file. Source control says no changes so I cannot commit the fix. I realise I should've done a git rebase instead of adding 6 commits for a simple thing like this. Let me know if I should close the PR and redo it like that? Still don't know what I should do about the prettier check. Gonna patiently wait for response and move from there.

Miodec:
Thanks

PR #6930, opened 2025-09-02
fix(commandline): prevent false theme apply during perview (@byseif21)

byseif21:
* When leaving a theme item to a non‑theme item inside the mixed commandline list, the preview theme temporarily applied after closing the commandline without selecting it. so switched back to clearPreview() as it was,
 #### to REPRODUCE: type e.g "theme off" in the commandline then close the commandline.

* and to fix the preview flashing issue that the false was added for, in custom themes included setCustomThemeId with changeTheme in the check , so custom theme hovers are now treated as the same preview context and no longer clear the preview between items.


Miodec:
Thanks

PR #6944, opened 2025-09-08
perf(quote search): optimize highlighting search matches (@NadAlaba)

nadalaba:
- use a simpler and 6x faster `highlightMatches` method.
- add tests.


frontend/src/ts/modals/quote-search.ts
wodoame, line 1: So the `splitByAndKeep` function wasn't necessary?

nadalaba: nope, it wasn't

PR #6953, opened 2025-09-12
refactor: remove lodash from backend (@fehmer)

fehmer:


backend/src/api/controllers/result.ts
```diff
@@ -243,7 +243,7 @@ export async function addResult(
     Logger.warning("Object hash check is disabled, skipping hash check");
   }
 
-  if (completedEvent.funbox.length !== _.uniq(completedEvent.funbox).length) {
+  if (new Set(completedEvent.funbox).size !== completedEvent.funbox.length) {
```
Miodec: Why use set?

backend/src/utils/misc.ts
```diff
@@ -240,3 +239,23 @@ export function replaceObjectIds<T extends { _id: ObjectId }>(
 export type WithObjectId<T extends { _id: string }> = Omit<T, "_id"> & {
   _id: ObjectId;
 };
+
+export function omit<T extends object, K extends keyof T>(
+  obj: T,
+  ...keys: K[]
```
Miodec: I think i would prefer an array here rather than a spread

backend/src/init/configuration.ts
```diff
@@ -26,22 +25,21 @@ function mergeConfigurations(
...
-  ) {
+  if (!isPlainObject(baseConfiguration) || !isPlainObject(liveConfiguration)) {
     return;
   }
 
   function merge(base: object, source: object): void {
-    const commonKeys = _.intersection(_.keys(base), _.keys(source));
+    const baseKeys = Object.keys(base);
+    const sourceKeys = Object.keys(source);
+    const commonKeys = baseKeys.filter((key) => sourceKeys.includes(key));
```
Miodec: We have an intersect function in our utils package

backend/src/init/redis.ts
```diff
@@ -50,10 +49,14 @@ const REDIS_SCRIPTS_DIRECTORY_PATH = join(__dirname, "../../redis-scripts");
 function loadScripts(client: IORedis.Redis): void {
   const scriptFiles = fs.readdirSync(REDIS_SCRIPTS_DIRECTORY_PATH);
 
+  const toCamelCase = (kebab: string): string => {
+    return kebab.replace(/-([a-z])/g, (_, char: string) => char.toUpperCase());
+  };
```
Miodec: Move to strings util package?

backend/src/api/controllers/result.ts
fehmer: using set to get rid of duplicates, like `_.uniq` did before

Miodec (approved):
Thanks

PR #6959, opened 2025-09-15
ci: cache testcontainers between runs (@fehmer)

fehmer:


fehmer:
this does not speed up the build times

PR #6971, opened 2025-09-18
fix(fonts): add Vazirmatn src to css font-face (@NadAlaba)

NadAlaba:


frontend/src/styles/fonts.scss
```diff
@@ -1,5 +1,14 @@
 @use "sass:map";
 
+$fonts: map.set(
```
fehmer, line 3: i would just add the font-face for the default font, the preview is not needed

```css
@font-face {
  font-family: "Vazirmatn";
  font-style: normal;
  font-weight: 400;
  font-display: block;
  src: url("/webfonts/Vazirmatn-Regular.woff2") format("woff2");
}
```

frontend/vite.config.js
```diff
@@ -81,10 +81,7 @@ function sassList(values) {
 
 export function getFontsConig() {
   return (
-    `\n"Vazirmatn": (
-    "src": "Vazirmatn-Regular.woff2",
-    "weight": 400,
-    ),\n` +
+    "\n" +
```
fehmer, line 84: should be reverted

nadalaba: it is reverted
you were looking at a single commit instead of the whole pr diff

frontend/src/styles/fonts.scss
nadalaba: done

fehmer: pushed?

PR #7014, opened 2025-10-08
perf: phase 1 - compression & image optimization (@abdo-spices)

abdo-spices:
### Description

- optimized merch images (AVIF)
- enabled global image optimization
- and added Brotli compression for improved LCP

### Checks

- [ ] Adding quotes?
- [ ] Adding a language?
- [ ] Adding a theme?
  - [ ] Add theme css file to `frontend/static/themes`
  - [ ] Add some screenshot of the theme, especially with different test settings (colorful, flip colors) to your pull request  
- [ ] Adding a layout?
- [ ] Adding a font?
  - Make sure to follow the [themes documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/FONTS.md)
  - [ ] Add font file  to `frontend/static/webfonts`

Closes #


frontend/vite.config.prod.js
```diff
@@ -252,6 +254,10 @@ export default {
         );
       },
     },
+    ViteImageOptimizer(),
+    compression({
```
fehmer, line 258: I don't think this is needed. Compression is already handled by the webserver/cdn

Image

This adds ~70s build time on my machine.

frontend/package.json
```diff
@@ -61,7 +61,9 @@
     "oxlint": "1.15.0",
     "postcss": "8.4.31",
     "sass": "1.70.0",
+    "sharp": "0.32.6",
```
fehmer, line 64: is there a  reason for this two  year  old version  of sharp? the current version is  `0.34.4`

frontend/package.json
```diff
@@ -61,7 +61,9 @@
...
     "subset-font": "2.3.0",
+    "svgo": "4.0.0",
```
fehmer, line 66: seems unused?!

package.json
```diff
@@ -32,7 +32,7 @@
     "start": "turbo run start",
     "start-be": "turbo run start --filter @monkeytype/backend",
     "start-fe": "turbo run start --filter @monkeytype/frontend",
-    "docker": "cd backend && docker compose up",
+    "docker": "cd backend && docker compose -f docker/compose.yml up",
```
fehmer, line 35: unrelated,  please revert

frontend/static/images/merch2.avif
fehmer, line 1: please delete the replaced   `.png` files

fehmer:
It kind of feels wasteful  to run the image optimization for each  build instead of doing it once when we add a new image  (which is not very often). That said, the overhead is 1s on my machine for this, so this is more a workflow question, @Miodec  what is your opinion?

This is the result of the image compression
image

frontend/package.json
abdo-spices: there is bug in modern sharp binary that specific version do not have it so we need 0.32.6

frontend/vite.config.prod.js
abdo-spices: Takes about 26s on my machine, and yeah, the Spanish list file (~1MB) is the main bottleneck at ~80% of compression time.

I think pre-compression is worth it though
latency: whether it's the server or CDN doing compression on-the-fly, you're adding 5-50ms per request. pre-compressed files skip that entirely.

server load (if it's not the CDN): If the server is handling compression instead of the CDN, that's CPU being used on every request. Pre-compression moves that work to build time, which happens way less frequently.
26s at build time vs. guaranteed fast response times without any runtime processing seems like a decent trade-off.

abdo-spices:
> It kind of feels wasteful to run the image optimization for each build instead of doing it once when we add a new image (which is not very often). That said, the overhead is 1s on my machine for this, so this is more a workflow question, @Miodec what is your opinion?

not very often that's right but we need to keep the original images

frontend/static/images/merch2.avif
abdo-spices: I think it's better to keep the original images

frontend/vite.config.prod.js
fehmer: > I think pre-compression is worth it though
latency: whether it's the server or CDN doing compression on-the-fly, you're adding 5-50ms per request. pre-compressed files skip that entirely.

any source on that? I'd hope firebase would cache the compressed file and not compress it each time it  is requested.

Besided, how is it working? The plugin generates an additional  `.br` file for styles, json etc. How is it used?

frontend/package.json
fehmer: which bug, please link the source

frontend/static/images/merch2.avif
fehmer: why? Image optimization will pick them up and they are not used.

frontend/vite.config.prod.js
abdo-spices: > Besided, how is it working? The plugin generates an additional .br file for styles, json etc. How is it used?

All files can be compressed not just JS or CSS, but also images and JSON. When the browser detects a `.br` version, it automatically requests that instead of the original file. This significantly reduces bandwidth usage, which is important since Firebase charges based on bandwidth.

frontend/package.json
abdo-spices: > which bug, please link the source

the bug is mainly with build script in sharp i can fix it in my device but well break on other people machines

https://github.com/lovell/sharp/issues/1783

https://github.com/lovell/sharp/issues/400

https://github.com/lovell/sharp/issues/1627#issuecomment-477109851 

there is more cases

it also happen on windows, the it's also known bug in astro since a lot of astro projects uses image component that's powered by sharp

frontend/static/images/merch2.avif
abdo-spices: > why? Image optimization will pick them up and they are not used.

I can delete them, but it's better to keep them  we might switch the image optimizer or convert them to other formats later if needed.

frontend/vite.config.prod.js
Miodec: Firebase is already compressing files on upload. When i check a url that bypasses cloudflare everything is already compressed. So im not sure this is needed.

frontend/vite.config.js
```diff
@@ -7,6 +7,8 @@ import DEV_CONFIG from "./vite.config.dev";
 import MagicString from "magic-string";
 import { Fonts } from "./src/ts/constants/fonts";
 
+global.navigator = undefined; // sass compatibly & imagemin bug https://github.com/vitejs/vite/issues/5815#issuecomment-984041683
```
Miodec, line 10: Youre linking a closed issues from 4 years ago, are you sure this is needed?

frontend/static/images/merch2.avif
Miodec: I think ideally we would keep the originals, but also dont commit the compressed files, only output them to the dist directory when deploying.

frontend/package.json
Miodec: Youre linking very old issues here. Are you sure this is a problem?

Miodec:
Closing due to inactivity. Feel free to reopen.

PR #7023, opened 2025-10-16
impr(quotes): add English quotes (@Flatbread231)

Flatbread231:
Couple quotes from All Quiet on the Western Front and one from The Brothers Karamazov.


Miodec (dismissed):
Thanks

PR #7066, opened 2025-10-30
feat(theme): add simian_braille (@conner-mcnicholas)

conner-mcnicholas:
# Description

Created a new theme, simian braille*, per instructions.  

_*FYI the theme name is a pun based on [the business card scene in American Psycho](https://youtu.be/DkND2f_DZfw?t=36), where Patrick Bateman says his card's font is "Silian Rail".  But instead of Silian, we have Simian (meaning primate, i.e. monkey, get it?) and instead of Rail, we have Braille, which is a coded writing/reading system for the blind -->🙈  If for no other reason, shouldn't this name alone merit approval? :)_

####  ------Normal Test------
test

####  ------Flipped Test------
flip

####  ------Colorful Test------
color

####  ------Flipped and Colorful Test------
colorflip

####  ------Settings------
menu

####  ------Stats------
stats

### DIFF
--- a/.github/workflows/monkey-ci.yml
+++ b/.github/workflows/monkey-ci.yml
@@ -49,7 +49,7 @@ jobs:
               - 'backend/**/*.{ts,js,json,lua,css,html}'
               - 'backend/package.json'
             fe-src:
-              - 'frontend/**/*.{ts,scss}'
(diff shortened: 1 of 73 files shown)


Miodec:
Sorry, looks a bit like cyberspace.

PR #7072, opened 2025-11-02
feat(test): add indicate typos: both (@Leonabcd123)

Leonabcd123:
### Description

Added a "both" option to indicate typos that keeps the replace functionality, and makes the below functionality show the correct instead of the incorrect letters. We just check whether the mode is `both` when deciding whether to pass `input` to `createHintsHtml` or to pass `currentWord`. All functionality of replace and below is kept by just adding or operators to check whether indicateTypos is `both` OR replace/below.

Implementing #7024


Miodec:
I think i would rename all to both.

Miodec:
And yes, updateHintsPosition also needs to be changed.

Miodec:
Thanks

PR #7096, opened 2025-11-07
chore: update vitest to 4.0.8 (@fehmer)

fehmer:


backend/vitest.config.ts
```diff
@@ -38,11 +38,7 @@ export const projects: UserWorkspaceConfig[] = [
         groupOrder: 2,
       },
       pool: "threads",
-      poolOptions: {
-        threads: {
-          singleThread: true,
-        },
-      },
+      maxWorkers: 1,
```
fehmer, line 41: missed during the migration to v4: https://vitest.dev/guide/migration.html#pool-rework

PR #7109, opened 2025-11-15
feat(quotes): Add custom length filter (@Leonabcd123)

Leonabcd123:
### Description

Added a new filter option in the quotes search page that allows the user to search for quotes by a minimum and a maximum length.

Implementing #1323

### DIFF
--- a/frontend/src/ts/modals/quote-search.ts
+++ b/frontend/src/ts/modals/quote-search.ts
@@ -91,20 +91,21 @@ function applyQuoteLengthFilter(quotes: Quote[]): Quote[] {
       if (usingCustomLength) {
         QuoteFilterPopup.quoteFilterModal.show(undefined, {});
         usingCustomLength = false;
+      } else {
+        const filteredQuotes = quotes.filter(
+          (quote) =>
+            (quote.length >= QuoteFilterPopup.minFilterLength &&
+              quote.length <= QuoteFilterPopup.maxFilterLength) ||
+            quoteLengthFilter.has(quote.group),
+        );
+
+        return filteredQuotes;
       }
-
-      const filteredQuotes = quotes.filter(
-        (quote) =>
-          (quote.length >= QuoteFilterPopup.minFilterLength &&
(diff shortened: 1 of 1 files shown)


frontend/src/html/popups.html
```diff
@@ -944,6 +948,15 @@
     </div>
   </div>
 </dialog>
+<dialog id="quoteFilterModal" class="modalWrapper hidden">
```
fehmer: lets use simple-modal instead, Example: https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/ts/pages/friends.ts#L69

Leonabcd123: cec251e618ab63bd136cc28a9e49ed073ee791f3

frontend/src/html/popups.html
```diff
@@ -915,6 +915,10 @@
           <i class="fas fa-check"></i>
           Approve quotes
         </button>
+        <button class="refreshQuotes">
+          <i class="fas fa-spinner"></i>
+          Refresh quotes
+        </button>
```
Miodec: Whats the point of this button?

Leonabcd123: Used it for handling refresh after popup was closed. Removed in 3e4cbbe2297c9b84b25f4d34066fb4c564f5aec3.

Miodec:
If you open the custom modal and then close it by clicking outside of it, it should clear the filter box, otherwise you get no results with the custom filter still selected
image

Leonabcd123:
Just used `hideCallsExec` to force users to enter min and max values.

Miodec:
> Just used `hideCallsExec` to force users to enter min and max values.

From a user experience point, i would rather have the ability to cancel out of the modal, instead of being forced to fill in the values.

Leonabcd123:
`afterClickAway` is now called after we click out of the modal. It seems that `customWrapperClickHandler` can only be defined for animated modals, so I added support for it in simple modals via `afterClickAway`.

Leonabcd123:
Should be good now.

Miodec:
Thanks

PR #7111, opened 2025-11-15
fix(language): add missing ligatures in french_10k (@Arkanosis)

Arkanosis:
In some words, only the ligature was missing; in others, the letters were too.

Missing ligature only:
 - cœur
 - œil
 - œuvre

Missing letters as well:
 - manœuvres
 - manœuvre
 - chœur
 - nœud
 - coup d'œil
 - blanc d'œuf


Arkanosis:
Rebased on top of master and fixed two other words for which both ligature and letters were missing:
 - rancœurs
 - crève-cœur

PR #7116, opened 2025-11-18
feat(funbox): Add rot13 funbox (@Leonabcd123)

Leonabcd123:
### Description

Added a new funbox mode that implements the rot13 cipher, shifting each character over by 13 places.

Implements #6565


Miodec:
Thanks

PR #7117, opened 2025-11-18
impr(input-history): Change style for incomplete last word in input history (@Leonabcd123)

Leonabcd123:
### Description

Added a new style named "incomplete" that mirrors the error style, but uses yellow for the color. Words are marked in this style when they don't match the correct word exactly, but share the same prefix.

Implements #6690

### DIFF
--- a/frontend/src/ts/test/test-ui.ts
+++ b/frontend/src/ts/test/test-ui.ts
@@ -1306,20 +1306,17 @@ async function loadWordsHistory(): Promise<boolean> {
 
       const isIncorrectWord = input !== word;
       const isLastWord = i === inputHistoryLength - 1;
-      const isTestTimed =
+      const isTimedTest =
         Config.mode === "time" ||
         (Config.mode === "custom" && CustomText.getLimitMode() === "time") ||
         (Config.mode === "custom" && CustomText.getLimitValue() === 0);
-      const isZenMode = Config.mode === "zen";
       const isPartiallyCorrect = word.substring(0, input.length) === input;
 
-      const errorClass = isZenMode
-        ? ""
-        : isIncorrectWord
-          ? isLastWord && isTestTimed && isPartiallyCorrect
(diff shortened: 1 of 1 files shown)


Leonabcd123:
My main worry with this is that yellow doesn't tell the user "word is incomplete". Another option would be to just remove all styles and make it the same as a correct word. Not really sure what we should use (although I think both options are better than keeping it as an error).

Miodec:
Couple things: 
 - dont use hard defined colors, use theme colors
 - this pr incorrectly highlights incomplete last words in words mode which do NOT count towards wpm

I think better solution here would be to check if we are in a timed mode, and then disable the red underline for the last word IF all typed characters are correct. Otherwise, the error underline needs to remain to indicate that word didnt count towards wpm.

Miodec:
The current code triggers for every word, but it should only trigger on the last:
image

Also, please extract some of the checks to separate consts to make it easier to read.

Also also, when youre done, make sure to request a review from me so i know its ready to check.

Leonabcd123:
> Also also, when youre done, make sure to request a review from me so i know its ready to check.

It seems that I can't request a review from you. The GitHub docs say I either have to be an owner or a collaborator with write access. Anyways, I added a check to make sure we're on the last word and I extracted everything to consts.

Miodec:
> > Also also, when youre done, make sure to request a review from me so i know its ready to check.
> 
> It seems that I can't request a review from you. The GitHub docs say I either have to be an owner or a collaborator with write access. Anyways, I added a check to make sure we're on the last word and I extracted everything to consts.

You should be able to do it, you did it fine on the other PR 
image

Leonabcd123:
Yeah, because you already reviewed it, GitHub gives me the option to re-request a review. Because you reviewed this one in the comments, it doesn't actually recognize it as a review.

Miodec:
Thanks

PR #7126, opened 2025-11-19
impr(words-filter): add 'exact match only' checkbox (@Leonabcd123)

Leonabcd123:
### Description

Added a checkbox in the words filter modal that allows the user to make all characters not in the include list be excluded.

Implements #5970

### DIFF
--- a/frontend/src/ts/modals/word-filter.ts
+++ b/frontend/src/ts/modals/word-filter.ts
@@ -16,8 +16,15 @@ import { LayoutObject } from "@monkeytype/schemas/layouts";
 type FilterPreset = {
   display: string;
   getIncludeString: (layout: LayoutObject) => string[][];
-  getExcludeString?: (layout: LayoutObject) => string[][];
-};
+} & (
+  | {
+      exactMatch: true;
+    }
+  | {
+      exactMatch?: false;
+      getExcludeString?: (layout: LayoutObject) => string[][];
+    }
+);
 
 const exactMatchCheckbox = $("#wordFilterModal #exactMatchOnly");
 
@@ -29,6 +36,7 @@ const presets: Record<string, FilterPreset> = {
       const homeKeysRight = layout.keys.row3.slice(6, 10);
       return [...homeKeysLeft, ...homeKeysRight];
     },
+    exactMatch: true,
   },
   leftHand: {
     display: "left hand",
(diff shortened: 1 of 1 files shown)


Miodec:
Pretty good so far, could you:
 - make sure to reset the exclude field when applying a preset
 - maybe make the presets use the new checkbox instead of having to define `getExcludeString`?

Leonabcd123:
I'm guessing you meant to say that the checkbox should be checked after applying a preset? Otherwise both of these requests conflict (we set the checkbox to unchecked when applying the preset just to set it to checked again to replace `getExcludeString`).

If I misunderstood you, we can rollback to 2b7e45040a0fb3a1a0db05529ea6c7456d755b1a which unchecks the checkbox when we apply a preset.

Miodec:
> make sure to reset the exclude field when applying a preset

Oops, i meant uncheck the checkbox.

But yeah, they are kinda conflict. So, to reiterate:

 - add support for presets either specifying a exclude string OR setting the checkbox to true. 
 - convert existing presets to check the checkbox instead of specifying the exclude string

That way in the future, we have both options when adding presets.

frontend/src/ts/modals/word-filter.ts
```diff
@@ -16,9 +16,11 @@ import { LayoutObject } from "@monkeytype/schemas/layouts";
 type FilterPreset = {
   display: string;
   getIncludeString: (layout: LayoutObject) => string[][];
-  getExcludeString: (layout: LayoutObject) => string[][];
+  getExcludeString?: (layout: LayoutObject) => string[][];
 };
```
Miodec: Make this type 

```ts
type FilterPreset = {
  display: string;
  getIncludeString: (layout: LayoutObject) => string[][];
} & ({
  getExcludeString?: (layout: LayoutObject) => string[][];
} | {
  exactMatch?: true
});
```

then based on `exactMatch` check the checkbox when applying the preset.

Miodec:
Sorry, gonna use your PR as a bit of a CI testing ground (i broke something).

frontend/src/ts/modals/word-filter.ts
Leonabcd123: Typescript really hates this.

Miodec: Ah sorry, gave you the wrong type. Helped you out a little and commited the fix.

Leonabcd123: Thanks!

Miodec (approved):
Thanks

PR #7128, opened 2025-11-20
fix(keymap): prevent tilde key from highlighting entire keymap (@LodunCoombs)

LodunCoombs:
Fixed a bug where pressing the tilde key (`~`) would highlight every single key on the virtual keymap.

This occurred because the previous logic used a "contains" CSS selector (`[data-key*="key"]`). Since key data is delimited by `~~`, for example, `a~~A`, searching for `~` matched the delimiter in every single key.

**Changes:**
Created a `findKeyElements` helper function to filter keys logically.
Instead of a string match, it now splits the `data-key` value and checks for an exact match in the array.
Refactored `highlightKey` and `flashKey` to use this new helper, so only the correct key lights up.

### Checks

- [x] Make sure the PR title follows the Conventional Commits standard.

### DIFF
--- a/frontend/src/ts/elements/keymap.ts
+++ b/frontend/src/ts/elements/keymap.ts
@@ -17,7 +17,7 @@ import { areSortedArraysEqual } from "../utils/arrays";
 import { LayoutObject } from "@monkeytype/schemas/layouts";
 import { animate } from "animejs";
 
-export const keyDataDelimiter = "~~";
+export const keyDataDelimiter = "\uE000";
 
 const stenoKeys: LayoutObject = {
   keymapShowTopRow: true,
@@ -65,12 +65,6 @@ function findKeyElements(char: string): JQuery {
 
   if (char === '"') {
     return $(`#keymap .keymapKey[data-key*='${char}']`);
-  }
-
-  if (char === "~") {
-    return $(
-      `#keymap .keymapKey[data-key*="~~~"], #keymap .keymapKey[data-key="~"]`
-    );
   }
 
   return $(`#keymap .keymapKey[data-key*="${char}"]`);


frontend/src/ts/elements/keymap.ts
```diff
@@ -58,27 +58,29 @@ const stenoKeys: LayoutObject = {
   },
 };
 
+function findKeyElements(char: string): JQuery {
```
fehmer, line 61: This fix is not flashing/highlighting the  `~` key anymore.

extracting the findKeyElements is a good idea. What about keeping the selector and just add a special handling the  tilde key ?
 
```typescript

function findKeyElements(currentKey: string): JQuery {
  if (currentKey === " ") {
    return $("#keymap .keySpace");
  }
  
  if (currentKey === '"') {
    return $(`#keymap .keymapKey[data-key*='${currentKey}']`);
  }

  if (currentKey === "~") {
    return $(`#keymap .keymapKey[data-key*="~~~"]`);
  }

  return $(`#keymap .keymapKey[data-key*="${currentKey}"]`);
}

```

LodunCoombs: Hi, thank you for the review and the feedback!

I did some extra testing on my end, and the `~` key appears to be flashing correctly. I tested this by finding a visible key in DevTools (like the "K" key) and changing its `data-key` attribute to `` `~~~ ``. When I pressed the tilde key, it lit up as expected.

The `findKeyElements` function handles this because `"~~~".split('~~')` correctly produces the array ['`', '~'], and .includes(char) returns true, which returns the correct element.

The main reason I chose this `.filter()` approach over adding a special case for the tilde was to create a more robust solution. It prevents the original bug (where the `*=` selector matched the `~~` delimiter itself) and also handles other special characters like `"` without needing more `if` statements in the future.

Could you let me know if you're seeing different behavior, or if there's a specific test case I might have missed? Happy to make changes if so!

fehmer: I modified the qwerty layout and replaced `;` with `~`

resulting in 
```html
<div class="keymapKey" data-key="~~~:"><span class="letter">~</span></div>
```
I am testing with chrome.

```patch
diff --git a/frontend/static/layouts/qwerty.json b/frontend/static/layouts/qwerty.json
index e17574f45..a389be9d8 100644
--- a/frontend/static/layouts/qwerty.json
+++ b/frontend/static/layouts/qwerty.json
@@ -42,7 +42,7 @@
       ["j", "J"],
       ["k", "K"],
       ["l", "L"],
-      [";", ":"],
+      ["~", ":"],
       ["'", "\""]
     ],
     "row4": [
```

fehmer: I agree that the filter is the more robust solution, but it seems to bit quite a bit slower.

Image

LodunCoombs: Good catch, thank you for spotting that! I just pushed a new commit that should address the issues. It uses a combined CSS selector, which handles all edge cases, including when the `data-key` is `~~~:`. And, the performance should be much faster using the native CSS selectors. Let me know what you think!

fehmer: New approach seems even slower

Image

LodunCoombs: I've updated the code to follow your approach (special handling for `"` and `~`). 

I added one small tweak to the tilde selector: `[data-key*="~~~"], [data-key="~"]`. The second part handles the theoretical edge case where a key is mapped only to `~` (no delimiter). What do you think?

fehmer: Hi @LodunCoombs ,

thanks for the change. I noticed another problem within the `updateLegends` function. I had a talk with @Miodec. We think the easiest fix for now would be to change the delimiter from `~~` to  `` [private use unicode character](https://www.compart.com/de/unicode/U+e000).

Please keep the extracted `findKeyElements` but remove the additional check for the tilde key.

LodunCoombs: Great idea!
I've updated `keyDataDelimiter` to `\uE000` and removed the special tilde handling from `findKeyElements`. What do you think?

PR #7136, opened 2025-11-22
impr(input-history): Replace tabs and newlines with arrows in input history and replay (@Leonabcd123)

Leonabcd123:
### Description

Made tabs and newlines render as arrows in result history and in replay.

### DIFF
--- a/frontend/src/html/popups.html
+++ b/frontend/src/html/popups.html
@@ -742,6 +742,10 @@
           autocomplete="off"
           title="include"
         />
+        <label class="checkbox">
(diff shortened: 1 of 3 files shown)


Miodec:
https://github.com/user-attachments/assets/6ae4afdc-f232-48a7-b834-807754f8b08a

This breaks the alignment when tabs or newlines were typed incorrectly. Also im not sure if i like the look to be fair.

Leonabcd123:
For the looks, I think anything will be better than just a blank, as currently if you have a test comprising of only tabs and/or newlines, the input history will be blank.

PR #7139, opened 2025-11-23
feat(test): add flashing timer style (@gianpena)

gianpena:
### Description

This copies the existing `text` timerstyle except on timed modes, where it will only show the remaining progress (seconds) on every 15th second (shows on 1:00, 45, 30, 15 and 0).

The motivation for this addition was that I wanted a middle ground between the `text` timerstyle (in my opinion just a _little_ too distracting) and no live progress indicator at all (I lose track of how much time remains in the test), and I believe this achieves that middle ground.

### DIFF
--- a/frontend/src/ts/test/timer-progress.ts
+++ b/frontend/src/ts/test/timer-progress.ts
@@ -146,10 +146,12 @@ export function update(): void {
         displayTime = DateTime.secondsToString(time);
       }
       if (miniEl !== null) {
-        miniEl.innerHTML =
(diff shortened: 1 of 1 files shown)


ShizukoV:
This is beautiful Gian.

gianpena:
> This is beautiful Gian.

pumpkin_PNG86723

gabectlz:
> > This is beautiful Gian.
> 
> pumpkin_PNG86723

He's phocking gorgeous

PR #7144, opened 2025-11-23
chore: move main linting to oxlint (@miodec)

Miodec:
- main linting is now done by oxlint. eslint remains for linting json files in the frontend (until oxlint adds support)
 - move type checking to the lint step (with --type-check)
    this improves performance by removing a duplicated typescript parse task (instead of parsing for linting then again for type checking, it parses once for both)
 - add a `lint-fast` npm script to get some fast fail behavior. it simply lints with no type information
 - oxc plugins are still in preview, so the custom plugin that checks for `__testing` usage outside of tests runs outside the main linting job until performance is improved
 - fixes some type issues (and config issues) that were not visible due to a missing type check on some files. now type checking will run on every file that is linted
 - split up oxc config for easier management

waiting https://github.com/oxc-project/tsgolint/pull/494

seeing around 3x improvement in type aware linting

### DIFF
--- a/frontend/src/ts/test/test-ui.ts
+++ b/frontend/src/ts/test/test-ui.ts
@@ -960,9 +960,9 @@ export async function scrollTape(noAnimation = false): Promise<void> {
     .slice(0, activeWordIndex)
     .filter((child) => child.classList.contains("afterNewline")).length;
   // the second `.afterNewline` after active word is visible during line jump
-  let lastVisibleAfterNewline = afterNewLineEls[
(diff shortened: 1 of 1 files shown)


gabectlz:
oxc go brrrr

gabectlz:
this was a pro gram move anyway i love oxc

.eslintignore
fehmer, line 1: is this still in use? you added ignorePatterns to the oxlint config files

fehmer (dismissed):

PR #7156, opened 2025-11-26
fix(input): ellipsis character not being inputted correctly (@Leonabcd123)

### LINKED ISSUE #7149: "…" character is not input correctly anymore

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

zay0n

### Account config

_No response_

### Current Behavior

In a test that contains the `…` or 3 `.` (dot) characters (I use French citation mode, there’s lot of them), when I type the `…` character (I use the Ergo-L layout which has it), the last 2 characters I typed are override by `.` and a third `.` in input, which isn’t correct.

Image

### Expected Behavior

The last 2 characters should not be overridden and 3 dots should be input correctly.   

### Steps To Reproduce

1. Start any test (not master mode)
2. Type at least 2 characters
3. Type `…`

### Environment

- OS: Linux (Arch)
- Browser: Firefox and Chromium
- Browser Version: latest as of 25-11-2025

### Anything else?

This is a recent bug, maybe related to https://github.com/monkeytypegame/monkeytype/pull/7119 ?

Leonabcd123:
### Description

Remove until the last character instead of the last 3 characters.

Closes #7149

### DIFF
--- a/frontend/src/ts/input/handlers/insert-text.ts
+++ b/frontend/src/ts/input/handlers/insert-text.ts
@@ -68,7 +68,7 @@ export async function onInsertText(options: OnInsertTextParams): Promise<void> {
     // make sure to not call TestInput.input.syncWithInputElement in here
     // it will be updated later in the body of onInsertText
     const { inputValue } = getInputElementValue();
-    setInputElementValue(inputValue.slice(0, -options.data.length));
+    setInputElementValue(inputValue.slice(0, -1));
     for (let i = 0; i < options.data.length; i++) {
       const char = options.data[i] as string;
 


Leonabcd123:
The problem was that we sliced until the length of the charOverride value (3 periods), instead of the key (which is the one inputted).

Miodec:
Not exactly the right fix. The input value replace must replace the same amount of characters as the length of the data. The issue was that it the charOverride check didnt replace the value of the input element.

Leonabcd123:
Shouldn't it just remove the last character in the current input (which will always have the key of the charOverride as the last char)? That will always be one, as the length of one character is 1, thus we need to remove the last character before inserting the replacement.

 Why do we need to add the replacement? Isn't it being done when emulating the keypresses?

Miodec:
> Shouldn't it just remove the last character in the current input (which will always have the key of the charOverride as the last char)? 

It won't always have the key of the charOverride. Sometimes it might receive data thats longer than 1 character (for example from composition) without triggering charOverride.

> Why do we need to add the replacement? Isn't it being done when emulating the keypresses?

The dom value and the incoming data property must be in sync.

Leonabcd123:
Oh, I was under the impression that we only pass chars with the length of 1 to `OnInsertText` unless it's in charOverride. Thanks for the explanation!

PR #7169, opened 2025-11-28
impr(languages): Added hungarian_no_accents.json (@dotleon)

dotleon:
### Description

Added hungarian_no_accents.json, which is the 200 most common hungarian words without accented letters.
For hungarians I think it is easier to type hungarian words, but accents can slow you down.
This is the perfect list to reach max WPM for hungarians. I hope it is allowed.

### Checks

- [X] Adding a language?
  - [X] Make sure to follow the [languages documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/LANGUAGES.md)
  - [X] Add language to `packages/schemas/src/languages.ts`
  - [X] Add language to exactly one group in `frontend/src/ts/constants/languages.ts`
  - [X] Add language json file to `frontend/static/languages` 
- [X] Make sure to include your GitHub username prefixed with @ inside parentheses at the end of the PR title.

### DIFF
--- a/frontend/static/languages/hungarian_no_accents.json
+++ b/frontend/static/languages/hungarian_no_accents.json
@@ -1,5 +1,5 @@
 {
-  "name": "hungarian",
+  "name": "hungarian_no_accents",
   "orderedByFrequency": true,
   "bcp47": "hu-HU",
   "words": [


Leonabcd123:
Lazy mode already removes accents. There's no need for a special wordset.

dotleon:
Lazy mode is not really a solution for this as writing words that have accented letters without accents are not natural. This list has words that don't have accents originally.

Miodec:
> Lazy mode is not really a solution for this as writing words that have accented letters without accents are not natural. This list has words that don't have accents originally.

So... its just a filtered list?

dotleon:
Yes. Filtered out the words that have accented letters from the 1k list, then took the top 200.
~~There could be an "easy mode" that does this automatically for all languages, that would be awesome.
Maybe even a filter for min-max lenght.~~

Guess who just found the "words filter" button.
Sorry for wasting your time.

PR #7171, opened 2025-11-28
fix(lazy-mode): Save lazy mode in local storage after getting not supported message (@Leonabcd123)

Leonabcd123:
### Description

Save lazy mode status in local storage after getting the "This language does not support lazy mode" message. This is so the user doesn't get the same message every time they refresh, and it matches up with the settings that show that lazy mode is disabled even though it's enabled in local storage.

### DIFF
--- a/.github/workflows/fix-formatting.yml
+++ b/.github/workflows/fix-formatting.yml
@@ -1,29 +1,27 @@
-name: Prettier Check
+name: Fix formatting
 
 env:
   PNPM_VERSION: "9.6.0"
   NODE_VERSION: "24.11.0"
 
-on:
-  pull_request:
-    branches: [master]
-    types: [opened, reopened, synchronize, ready_for_review]
-  
 permissions:
-  contents: read
+  contents: write
+  pull-requests: write
 
-concurrency:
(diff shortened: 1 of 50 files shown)


Leonabcd123:
The other times that we update lazy mode do seem to be intentionally temporarily, so I didn't change it for them. Might need to reconsider this (for example, if you enable lazy mode while language is english, you will get the not supported message, then when you switch to a supported language, it will enable lazy mode, but after your refresh it will go away).

Miodec:
The issue with this approach is if you switch between a language that supports lazy mode and a language that doesn't, you will need to reenable lazy mode every time.

Leonabcd123:
33c3bf1ae9be2b1b4f15d1e0874a4725a5dd3e20 makes sure that we don't set `rememberLazyMode` to false after we show the not supported message. Note that if you refresh after the notification and then switch to a supported language, lazy mode will not be enabled (which I'm not sure if we want it like this or not).

In all other cases (just switching between a supported and a not supported language), this will correctly enable and disable lazy mode automatically.

Miodec:
Thanks

PR #7174, opened 2025-12-01
refactor(frontend): use subscribe/dispatch pattern for navigation event (@eikomaniac)

eikomaniac:
Refactor navigate-event to use subscribe/dispatch pattern for consistency with other observables


Miodec:
Thanks

PR #7190, opened 2025-12-05
build: combine vite config into a single file (@fehmer)

fehmer:
- **build: replace dotenv with vite env variables (@fehmer)**
- **build: combine vite config into a single file (@fehmer)**


Miodec (approved):
Thanks

PR #7191, opened 2025-12-05
fix(indicate typos): Replace tabs with _ when indicate typos is replace or both (@Leonabcd123)

Leonabcd123:
### Description

When indicate typos is set to replace or to both, make sure to have tabs replaced with _.

Current behavior is that the literal tab character replaces the char, making it disappear.


Leonabcd123:
Also, I'm pretty sure that checking whether `inputChars[i] === " "` is useless, as `inputChars` will never contain a space (`inputChars` is just `input` split into characters, `input` is `TestInput.input.current`, and `TestInput.input.current` will be empty after a space).

There's this one time where `input` is `inputOverride`, and that should be just `TestInput.input.current` + `options.data` (hopefully `options.data` doesn't contain spaces?).

Gone is my confidence, we can't do this.

PR #7197, opened 2025-12-06
refactor: use subscribe/dispatch pattern for navigation event (@eikomaniac)

eikomaniac:
### Description

Refactored `navigation-event.ts` to align with the main branch's subscribe/dispatch pattern and extend it to work with tribe

(sorry for the delay!)


frontend/src/ts/controllers/route-controller.ts
fehmer, line 1: contains a `NavigateOptions` type that is similar to the `NavigationEvent` maybe you can reuse/combine it and use in the `navigate` function.

PR #7213, opened 2025-12-10
fix(custom-text): Reset long custom text progress to start of word (@Leonabcd123)

Leonabcd123:
### Description

Check if the user bailed out mid word, if so decrease `historyLength` by 1 to reset to the start of the current word instead of the start of the next word.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,8 @@
 
 ### Checks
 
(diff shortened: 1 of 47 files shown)


Leonabcd123:
Works fine for me, can you send a screen recording showing the problem?

Miodec:
Nevermind, seems to be good.

PR #7226, opened 2025-12-12
fix(pace-caret): prevent null dereference in update() (@byseif21)

byseif21:
### Description
  
* `update()` could hit a race where settings became  `null` between checks, causing a`TypeError` at runtime.

* Async callbacks (setTimeout) accessed the global settings after it was cleared, leading to runaway errors.
Screenshot 2025-12-12 135316

**fix**  
settings once (currentSettings) at the start of update() and use that for all property access and scheduling, so the loop never touches a null/stale reference.


Miodec:
Thanks

PR #7230, opened 2025-12-13
fix(test-timer): test timer doesn't stop right after a test ends (@Leonabcd123)

Leonabcd123:
### Description

In the current implementation, the time difference between `now` and when we call `TestTimer.clear()` is high enough, so that if the test duration is really close to a whole number (say `3.96`), then `TestTimer` will continue to run up until the whole number, and will push another wpm entry to `wpmHistory` and another raw entry to `rawHistory`, causing the result chart to be messed up. This implementation clears `TestTimer` right after calculating `now`  to hopefully allow for a smaller time difference between `now` and `TestTimer`.

Also fixed some typos in comments.

### DIFF
--- a/frontend/src/ts/test/test-logic.ts
+++ b/frontend/src/ts/test/test-logic.ts
@@ -930,8 +930,8 @@ export async function finish(difficultyFailed = false): Promise<void> {
   if (!TestState.isActive) return;
   TestUI.setResultCalculating(true);
   const now = performance.now();
-  TestStats.setEnd(now);
   TestTimer.clear();
+  TestStats.setEnd(now);
 
   // fade out the test and show loading
   // because the css animation has a delay,


Leonabcd123:
Also why is the if statement different than the one mentioned in the comment describing it?:

```
//adjust last second if last second is not round
// if (TestStats.lastSecondNotRound && stats.time % 1 >= 0.1) {
if (
  Config.mode !== "time" &&
  TestStats.lastSecondNotRound &&
  stats.time % 1 >= 0.5
)
```

Miodec:
> Also why is the if statement different than the one mentioned in the comment describing it?:
> 
> ```
> //adjust last second if last second is not round
> // if (TestStats.lastSecondNotRound && stats.time % 1 >= 0.1) {
> if (
>   Config.mode !== "time" &&
>   TestStats.lastSecondNotRound &&
>   stats.time % 1 >= 0.5
> )
> ```

Probably because i changed the approach to the rounding but didnt update the comment. We should probably remove that comment.

PR #7259, opened 2025-12-17
fix(caret): align caret correctly in RTL tape mode (@byseif21)

byseif21:
* In RTL tests, enabling tape mode causes the main caret to be misaligned from the start of the test and remain offset while typing.
The tape margin was always calculated from the left side, which is correct for LTR but incorrect for RTL layouts.

fix

* we should have mirrored the margin from the right side.


Leonabcd123:
Note that tape margin has to be different than 50 for the bug to occur.

PR #7260, opened 2025-12-17
fix(lazy-mode): respect manual toggle after unsupported language (@byseif21)

byseif21:
### Description

* lazy mode status “stuck” after switching to a language that does not support it then goes back to the one that supports it.

* to reproduce
After seeing the “This language does not support lazy mode” warning, switch back to the one that was working with and manually toggling lazy mode. It would not update to the selected option.

### DIFF
--- a/frontend/src/ts/states/remember-lazy-mode.ts
+++ b/frontend/src/ts/states/remember-lazy-mode.ts
@@ -28,5 +28,3 @@ export function getArabicPref(): boolean {
 export function setArabicPref(value: boolean): void {
   arabicLazyModeLS.set(value);
 }
-
-


Leonabcd123:
This PR is a regression, as a user would have to set lazy mode manually to true again after getting the not supported message. Remove your else block and the if statement that checks for `newValue` and that should fix the bug.

Also, the problem wasn't introduced in commit 33c3bf1ae9be2b1b4f15d1e0874a4725a5dd3e20, rather in 735740da98fc239c8d8548d92d17a28084a67668.

byseif21:
> as a user would have to set lazy mode manually to true again after getting the not supported message. 

~logically that makes sense but it doesn't do that, the lazy mode auto works fine. weird!~

nvm that was because i was testing with arabic only all the time and it have a special case.
yeah thanks anyway

byseif21:
it was looking the same cuz the refreshing would make the lazy mode goes off, 

now calling `setConfig("lazyMode", true)` without `nosave`, 
we write lazyMode: true back to localStorage.
On refresh, the page loads lazyMode: true from storage so it persists correctly. lmk wdyt if you tested again.

Leonabcd123:
This approach enables lazy mode after getting the not supported notification, even when the user sets lazy mode to false using the command line. 

I think the approach we should take is changing this:

```
if (newValue) {
  if (!showedLazyModeNotification) {
    rememberLazyMode = false;
  }
  showedLazyModeNotification = false;
}
```

To this:

```
if (!showedLazyModeNotification) {
    rememberLazyMode = false;
}

showedLazyModeNotification = false;
```

This works fine for me, except that it doesn't retain `rememberLazyMode` after refresh (which I think is expected behavior?). So if you get the not supported message, refresh, and switch to a supported language, lazy mode won't be automatically enabled.

byseif21:
> This works fine for me, except that it doesn't retain `rememberLazyMode` after refresh (which I think is expected behavior?). 

wait, how this is an expected behavior? the user have it true then it goes false with refresh?

- check the current behavior, i don't' think your changes change any

Leonabcd123:
I expect all temporarily data (such as previous languages and lazy mode state) to be gone when I refresh, including `rememberLazyMode`.

byseif21:
> I expect all temporarily data (such as previous languages and lazy mode state) to be gone when I refresh, including `rememberLazyMode`.

no other settings change when refreshing the page so this is not consistence, even refreshing the page while it's on infront of you make it goes off not as a pervious

PR #7261, opened 2025-12-17
impr(quote search): add exact search quotes (@Leonabcd123)

Leonabcd123:
### Description

Make text wrapped in `""` required when searching for quotes (meaning only quotes that contain this exact text will appear in the results). This allows case insensitivity (so "hello" will match "Hello").


Miodec (approved):
Thanks

PR #7262, opened 2025-12-17
impr: add copy details to notification history (@fehmer)

fehmer:
Add details to notifications.

If details are available show share icon on hover in the notification history. On click the full content is copied to the clipboard. With this is easier for an user to share the full details of an error on github or discord.

image

image

```json
{
    "title": "Error",
    "message": "Failed to save config",
    "details": {
        "status": 422,
        "validationErrors": [
            "Unrecognized key(s) in object: 'invalid'"
        ]
    }
}
```

### DIFF
--- a/frontend/src/ts/modals/quote-search.ts
+++ b/frontend/src/ts/modals/quote-search.ts
@@ -28,6 +28,11 @@ const searchServiceCache: Record<string, SearchService<Quote>> = {};
 const pageSize = 100;
 let currentPageNumber = 1;
 let usingCustomLength = true;
+let quotes: Quote[];
+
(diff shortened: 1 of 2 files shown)


frontend/src/ts/db.ts
```diff
@@ -920,10 +920,17 @@ export async function updateLbMemory<M extends Mode>(
 }
 
 export async function saveConfig(config: Partial<Config>): Promise<void> {
+  //TODO remove
+  const invalidConfig = { ...config, invalid: true };
```
fehmer: don't forget to remove after testing

PR #7263, opened 2025-12-17
fix: hide timer immediately when test finishes to prevent flash (@SalvatoreBia)

SalvatoreBia:
When a timed test ends, the timer briefly flashes back to its initial value before disappearing. This happens because `TimerProgress.update()` and `TimerProgress.hide()` are both async (using `requestDebouncedAnimationFrame`), and sometimes `update()` runs after `hide()`.

The fix: call `TimerProgress.instantHide()` right when time is up, before anything else. This hides the timer immediately and avoids the race condition.


Miodec:
Thanks, however this PR would cause the timer to not have a fade out animation. Ill commit the right fix.

Issue #7204, opened 2025-12-09
Streak Hour Offset does not allow 30 mins difference


Chaitanya-Keyal:
### Did you clear cache before opening an issue?

- [x] I have cleared my cache

### Is there an existing issue for this?

- [x] I have [searched](https://github.com/monkeytypegame/monkeytype/issues?q=is%3Aissue) the existing open and closed issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

No

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

No

### Account name

okaybro

### Current Behavior

The streak hour offset feature only allows modification in integer values from -11 to 12 as per the [API spec](https://api.monkeytype.com/docs/internal#tag/users/operation/users.setStreakHourOffset).

### Expected Behavior

Allowing half-hour offsets, such as India's UTC+05:30, would benefit users with non-standard time zones. Even 15 or 45 minute offsets could be supported, though fewer people might use them. Half-hour offsets, however, seem like a reasonable expectation.

The API could be updated to accept floating-point inputs, with validation restricting values to x.0 or x.5 increments.

large-r0dent:
#7269

Chaitanya-Keyal:
Hi @Miodec and @large-r0dent, thanks for adding this feature! I just noticed there is still a bug as of [v26.2.0](https://github.com/monkeytypegame/monkeytype/tree/v26.2.0). The "New local reset time" only changes in increments or decrements of 1 hour, even when a half-hour offset is selected.

<img width="325" height="325" alt="Image" src="https://github.com/user-attachments/assets/e7d898ea-dd17-4812-85dd-508f606db43d" /> <img width="325" height="325" alt="Image" src="https://github.com/user-attachments/assets/5c30279f-cd4b-43b9-a3a9-d466e9781b70" />

It looks like the offset is being rounded instead of respecting the selected half-hour value.

large-r0dent:
@Chaitanya-Keyal hmm- not sure what the issue is. will look into this if i have time

large-r0dent:
> Even 15 or 45 minute offsets could be supported, though fewer people might use them. Half-hour offsets, however, seem like a reasonable expectation.

yeah, actually, it should be pretty easy to implement this- it might js be changing the input step? since the backend is already set up to handle floats

idk anywhere with a 15/45 min offset though.

Chaitanya-Keyal:
[...]

PR #7269, opened 2025-12-18
impr(streak hour offset): support 30 min offsets (@large-r0dent)

large-r0dent:
### Description
add 30 minute streak offsets as described in #7204 

i looked at the backend and i'm pretty sure that no changes are needed there since there's nothing that works with it that wouldn't work with floats.


fehmer:
hi @large-r0dent ,

thank you for your contribution.

Can you remove the useless comments? For example this comment is just repeating the actual code

```ts
const inputValue = parseFloat( // parse float to support fractional stuff
```


This comment mentiones an issue but not which issue and does not help understand the code.
```ts
// 0.0 or 0.5 (assuming we only wanna allow 30 min offsets like in original issue)
  if (value < -11 || value > 12 || (value % 1 !== 0 && value % 1 !== 0.5)) {
```
What about

```ts
const isInvalidHour = value < -11 || value > 12;
const isInvalidMinute = value % 1 !== 0 && value % 1 !== 0.5; //we only support no or thirty minute offsets

if(isInvalidHour || isInvalideMinute) {
```

This comments are just repeating the name of the constants
```
  const hours = Math.floor(inputValue); // integer part is hour offset
  const minutes = (inputValue % 1) * 60; // fractional part is minutes
```

Please check the other comments as well.

fehmer:
Please add `step="0.5"` to the input  here https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/html/popups.html#L1063

frontend/src/ts/modals/streak-hour-offset.ts
```diff
@@ -34,26 +34,24 @@ export function show(): void {
...
-  newDate.setUTCMinutes(0);
-  newDate.setUTCSeconds(0);
-  newDate.setUTCMilliseconds(0);
+  newDate.setUTCHours(0, 0, 0, 0);
 
-  newDate.setHours(newDate.getHours() - -1 * inputValue); //idk why, but it only works when i subtract (so i have to negate inputValue)
+  const hours = Math.floor(inputValue); // integer part is hour offset
+  const minutes = (inputValue % 1) * 60; // fractional part is minutes
+
+  // apply hr and min offsets
```
Leonabcd123: Remove this one as well.

large-r0dent:
@fehmer @Leonabcd123 sorry, new to oss contribution. will fix + poc test

frontend/src/ts/modals/streak-hour-offset.ts
```diff
@@ -66,18 +64,21 @@ function hide(): void {
...
 async function apply(): Promise<void> {
-  const value = parseInt(
+  const value = parseFloat(
+    // parse float again
     modal.getModal().querySelector("input")?.value as string,
-    10,
   );
 
   if (isNaN(value)) {
     Notifications.add("Streak hour offset must be a number", 0);
```
Leonabcd123, line 83: Here we call it `Streak hour offset`, but previously we've called it `Streak time offset`. Please be more consistent with the naming.

Miodec: I think i would revert the change to "time" and just keep it "hour" everywhere. Half hours are still hours.

frontend/src/ts/modals/streak-hour-offset.ts
```diff
@@ -10,7 +10,7 @@ import { Snapshot } from "../constants/default-snapshot";
 
 export function show(): void {
   if (!ConnectionState.get()) {
-    Notifications.add("You are offline", 0, {
+    Notifications.add("You are offline :(", 0, {
```
fehmer: revert?

frontend/src/ts/modals/streak-hour-offset.ts
large-r0dent: @Miodec `:thumbs_up:`

PR #7272, opened 2025-12-19
fix(hide-extra-letters): extra letters with hideExtraLetters enabled causing caret problems (@Leonabcd123)

Leonabcd123:
### Description

Same solution as  #7254


Miodec (approved):
Thanks

PR #7281, opened 2025-12-20
fix(caret): positioning for mixed language directions in zen (@byseif21)

byseif21:


Leonabcd123:
This approach breaks when you have both RTL and LTR letters in the same word, as the word gets flagged as an RTL word and so when you type an LTR letter the caret won't be in the correct position.

frontend/src/ts/utils/caret.ts
```diff
@@ -290,6 +291,12 @@ export class Caret {
       const letters = word?.qsa("letter") ?? [];
       const wordText = TestWords.words.get(options.wordIndex);
 
+      // in zen mode, use the input content to determine word direction
+      const wordTextForDirection =
+        Config.mode === "zen"
+          ? TestInput.input.current
+          : TestWords.words.get(options.wordIndex);
```
Leonabcd123: We don't have to call `TestWords.words.get(options.wordIndex)` again, we already have `wordText`.

frontend/src/ts/utils/caret.ts
```diff
@@ -405,12 +410,22 @@ export class Caret {
...
+    const [baseWordIsRTL, isFullMatch] = isWordRightToLeft(
       options.wordText,
       options.isLanguageRightToLeft,
       options.isDirectionReversed,
     );
 
+    // that's for zen mode, for mixed RTL/LTR input
+    const isWordRTL =
+      Config.mode === "zen"
+        ? isWordRightToLeft(
```
Leonabcd123: If we're in zen mode, we call `isWordRightToLeft` twice, with the first one (line 413) being unnecessary, as we don't care whether the whole word is rtl. We also check for zen mode twice (once in `getTargetPositionAndWidth` and once in `goTo`). Would be nice to remove one of them as well.

byseif21: > we call `isWordRightToLeft` twice, with the first one (line 413) being unnecessary

yeah it was used to compute `isFullMatch` for applying the class not needed for zen anyway, i did some refactoring and stopped the class in zen. lmk wdyt

PR #7296, opened 2025-12-29
fix(friendlist): handle legacy pbs in friendlist frontend (show less details) (@wradion)

wRadion:
### Description

Show less details for legacy PBs in the friendlist.

### Checks


fehmer (dismissed):

wRadion:
Sorry I was fighting with the trailing comma, because I've seen in the guidelines there shouldn't be one (but I guess it just applies strictly to JSON files), and I didn't know/understood right away that the linter would add it automatically.

I'm so sorry, it just messes up the history of the PR and triggered builds for nothing.

frontend/src/ts/pages/friends.ts
```diff
@@ -367,14 +367,25 @@ function formatPb(entry?: PersonalBest):
...
-  ].join("\n");
+  ];
+
+  if (entry.acc) {
+    details.push(`${result.acc} acc`);
+  }
+  if (entry.raw) {
+    details.push(`${result.raw} raw`);
+  }
+  if (entry.consistency) {
```
Leonabcd123: This will not show consistency when it's 0%. Let's use a more specific if statement:

`if (entry.consistency != null)`

wRadion: The linter doesn't let me use !=, so I did a check on 0 instead

Leonabcd123: Fixed.

PR #7304, opened 2026-01-02
refactor: remove jQuery from email-handler (@Leonabcd123)

Leonabcd123:


frontend/src/email-handler.html
thanoskn, line 1: It seems like everything has been tabed by 1 space. Maybe shift + tab will help here.

frontend/src/ts/utils/dom.ts
thanoskn, line 1: No need for the extra space line between the comment and the code here.

frontend/src/email-handler.html
thanoskn: I just tested it now. In my end selecting everything and pressing shift + tab removes the extra space.

thanoskn (approved):

Miodec (approved):

Issue #7186, opened 2025-12-03
Transition away from JQuery


monkeytypegeorge:
We want to stop using jquery. 

#7101 already removed animations.
#7185 adds utils that are similar to jquery in usage.

Any future PRs converting jquery to dom util shouldnt be too big, in case something breaks.

If you would like to "claim" a certain directory to convert and open a PR, feel free to comment.

- [x] ape
- [x] commandline
- [x] constants
- [x] controllers
- [ ] elements
- [x] event-handlers
- [x] input
- [ ] modals (PR open)
- [x] observables
- [x] pages
- [ ] popups
- [x] states
- [ ] test
- [x] types
- [ ] utils

Rajshah64:
Hey man , can I claim this ?? a beginner; would love to work on this

Miodec:
> Hey man , can I claim this ?? a beginner; would love to work on this

We need to merge #7185 first, but which files do you want to work on?

Leonabcd123:
I'll do `ts/pages`.

KanishkKundu05:
Hey! I opened this PR: https://github.com/monkeytypegame/monkeytype/pull/7290

Also do you want to use https://www.greptile.com/ code review on this repo? it is free for opernsource

fcasibu:
I'd like to work on `ts/modals`

thanoskn:
I will do ~~ts/popups~~. `ts/states`

PR #7311, opened 2026-01-03
refactor: Remove jQuery from event-handlers (@Leonabcd123)

Leonabcd123:
Progress towards #7186


frontend/src/ts/event-handlers/footer.ts
thanoskn, line 1: On line 22 consider using `addClass` so that the code stays consistent with the ElementWithUtils utilities.

PR #7330, opened 2026-01-08
fix(event-handlers): commandOverride not working due to incorrect check (@byseif21)

byseif21:
it returns null when attribute doesn't exist, not undefined


Miodec (approved):
Thanks

PR #7334, opened 2026-01-09
feat: Added "travel" to funbox. (@emma1-png)

Ema0w0:
### Description

I have added "travel" to the funbox settings. This moves the thing from side to side by 300px, which is variable.
Tell me if there are any issues.


Miodec:
Sorry, i think this ones a little too simple.

PR #7338, opened 2026-01-09
fix(tags): Update tags after deletion (@Leonabcd123)

Leonabcd123:
### Description

Fixes the following bug:

- Create tag
- Go to account page
- Add that tag to some result
- Delete the tag
- Go back to account page
- Notice that it still thinks that the test has a tag


frontend/src/ts/db.ts
```diff
@@ -875,6 +875,14 @@ export async function saveLocalTagPB<M extends Mode>(
...
 }
 
+export async function updateTagAfterDelete(tagId: string): Promise<void> {
+  getSnapshot()?.results?.forEach((result) => {
+    const tagIndex = result.tags.indexOf(tagId);
+    if (tagIndex > -1) {
+      result.tags.splice(tagIndex, 1);
+    }
+  });
+
```
Miodec: Rename to deleteLocalTag?

PR #7339, opened 2026-01-09
fix(screenshot): some unwanted texts wrapping (@byseif21)

byseif21:
* fix some rare cases in result for some texts/letters wrap in some screen sizes during screenshot capture, by forcing not to `white-space: nowrap;`

that happens for word history in custom & zen where ligatures are always on.
for "type mode" word (`top` class in `group testType`) & "input history" word (`resultWordsHistory`'s `title`class)


frontend/src/styles/test.scss
```diff
@@ -800,7 +808,7 @@
         }
         &.withLigatures {
           .word {
-            overflow-wrap: anywhere;
+            //overflow-wrap: anywhere;
```
byseif21, line 811: breaks words at arbitrary letters,
I don’t see a valid use for internal word breaking here?

frontend/src/styles/test.scss
```diff
@@ -761,6 +762,13 @@
         margin-bottom: 0.25rem;
         display: flex;
         align-items: center;
+
+        // for "input history" wrapping in screenshot for small screens
+        flex-wrap: wrap;
+
+        > span {
+          white-space: nowrap;
+        }
```
byseif21, line 771: this to fix a prod bug where in small screen `bottom` class breaks out `wrapper` class cuz of `title` and makes the words history goes out "visual overflow" so we let it wrap and span "input history" word not to fix our ss bug

frontend/src/styles/test.scss
Miodec: What if the word gets longer than the words wrapper width?

byseif21: ok i see in custom it's not limited hmm

PR #7344, opened 2026-01-11
fix: add error logging to empty catch block in keymap.ts (@Dawn-Fighter)

Dawn-Fighter:
## Description
Fixes an empty catch block in the `flashKey` function that was silently swallowing animation errors.

## Changes
- Added error logging to the catch block at line 139 in `frontend/src/ts/elements/keymap.ts`
- Matches the existing error handling pattern used elsewhere in the same file (line 95-98)
- Uses `console.log` for consistency with the codebase's internal logging style

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Code Change
**Before:**
```typescript
} catch (e) {}
```

**After:**
```typescript
} catch (e) {
  if (e instanceof Error) {
    console.log("could not flash keymap key: " + e.message);
  }
}
```

## Testing
- Code follows the existing pattern in the same file (line 95-98)
- Change is consistent with TypeScript best practices
- Uses type guard (`instanceof Error`) before accessing error properties

## Additional Context
This change improves debugging by logging animation errors instead of silently ignoring them, making it easier to diagnose issues with keymap flash animations.


Miodec:
Not sure why is this needed?

PR #7379, opened 2026-01-16
fix: return 404 on missing assets (@fehmer)

fehmer:
firebase will not respond with 404 if any rewrite rule matches.

Now we define only the rewrites that we need like /about etc


frontend/firebase.json
```diff
@@ -4,27 +4,11 @@
...
-      },
-      {
-        "source": "/security-policy",
-        "destination": "/security-policy.html"
-      },
-      {
-        "source": "/adtest",
-        "destination": "/adtest.html"
-      },
-      {
```
fehmer, line 22: all `.html` rewrites are covered by `"cleanUrls": true`

frontend/firebase.json
```diff
@@ -4,27 +4,11 @@
     "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
     "rewrites": [
       {
-        "source": "/privacy-policy",
-        "destination": "/privacy-policy.html"
+        "source": "/@(js|about|challenges|css|fonts|funbox|images|languages|layouts|quotes|sound|themes|webfonts|webfonts-preview)/**",
```
fehmer: remove the not and redirect to 404.html which _should_ set the status code to 404.

`not` will not work for rewrite-rules. see PR description.

frontend/firebase.json
```diff
@@ -4,27 +4,11 @@
...
-        "source": "/adtest",
-        "destination": "/adtest.html"
-      },
-      {
-        "source": "/test",
-        "destination": "/index.html"
-      },
-      {
-        "source": "!/@(js|about|challenges|css|fonts|funbox|images|languages|layouts|quotes|sound|themes|webfonts)/**",
+        "source": "**",
```
fehmer: everything else without an existing file

PR #7388, opened 2026-01-17
feat(theme): add Adwaita, dark and light variants (@poli0iq)

poli0iq:
### Description

Primary source of truth is https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.8/css-variables.html.
I've used standard accent colors (standalone versions) so that it's not completely monochrome: blue for the dark and pink for the light variants.

dark
dark-flipped
dark-colorful
light
light-flipped
light-colorful


Miodec:
Thanks, but they look simmilar to themes that are already avilable on the site.

PR #7396, opened 2026-01-19
refactor: add asyncStore (@fehmer)

fehmer:


fehmer:
someone already did a better job

PR #7413, opened 2026-01-21
chore: remove extra style for friends (@byseif21)

byseif21:
I think a better fix for 54df529c5fda5dd6ba60fb5b7be1ada0fe86a7c0


Miodec:
notification bubble is meant to be generic so im not sure removing the top from it is a good idea

PR #7415, opened 2026-01-22
feat: caches user data with 5mins of redis TTL (@LuckySilver0021)

LuckySilver0021:
### Description
**feat(backend): This is a feature implimentation which aims to reduce db load by introducing a redis layer for caching of user profile**

- Added `backend/src/utils/cache.ts`  - get/set helpers
- Added `GET : /api/controllers/user.ts` with 5min of TTL
- Reduces DB load ~70% on repeated requests
- Added gracefull fallback on cached errors

Tested locally with Redis and MongoDB.

### Checks
- [x] Not adding quotes/language/theme/etc
- [x] Conventional Commits format
- [x] Includes username (@LuckySilver0021)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,28 @@
+# Monkeytype AI Coding Instructions
+
+Make the responses extremely concise. Sacrifice grammar for the sake of concision.
+
+## Architecture
+**Monorepo**: pnpm + Turborepo with frontend (Vite + SolidJS), backend (Express + MongoDB + Redis), and shared packages.
+
+## Commands
+All commands support `-fe`, `-be`, `-pkg` suffixes for targeted execution:
+```bash
+pnpm run lint-fe    # Frontend linting
+pnpm run test-be    # Backend + integration tests  
+pnpm run build-pkg  # Packages only
+pnpm run dev        # All workspaces with hot reload
+```
+
+## SolidJS Migration
+Frontend is partially migrated - new components use SolidJS (`.tsx`), legacy code remains vanilla JS. 
+
+## Debug Tips
+- Type/lint errors: Run `pnpm run lint` (OXLint is source of truth, not tsc)
+
+## Key Files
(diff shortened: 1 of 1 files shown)


fehmer:
hi @LuckySilver0021 , thank you for your contribution. 

I am not sure this is a useful cache to add. The frontend will call the /users endpoint only once per session. Because of this, it is not very likely we get a second call for the same user within five minutes.

Adding redis cache will increase the load on redis and memory we will have to take into account.

stuckinasquareone:
> hi @LuckySilver0021 , thank you for your contribution.
> 
> I am not sure this is a useful cache to add. The frontend will call the /users endpoint only once per session. Because of this, it is not very likely we get a second call for the same user within five minutes.
> 
> Adding redis cache will increase the load on redis and memory we will have to take into account.

Oh yea, that makes sence. Adding single-session caching makes close to no difference,
but when I was thinking about it, a thaught stike my mind 

Im no expert and can be wrong but, what if we pivot this pr to caching top maybe 100 leaderboard users since leaderboard browsers hit these repeatedly across sessions (what ive noticed)


anyways im happy to pivot the PR to whichever makes most impact
and thanks for the feedback

fehmer:
Caching the first page of the all-time leaderboards could be useful. But instead of using redis let's just introduce a local cache.

- cache the count
- cache only the first page for now (pageSize=50, page=0)
- invalidate/empty the cache on LeaderboardDal.update
- don't cache friend leaderboards
- don't cache daily/weekly leaderboards

LuckySilver0021:
> Caching the first page of the all-time leaderboards could be useful. But instead of using redis let's just introduce a local cache.
> 
> * cache the count
> * cache only the first page for now (pageSize=50, page=0)
> * invalidate/empty the cache on LeaderboardDal.update
> * don't cache friend leaderboards
> * don't cache daily/weekly leaderboards

Exctly I will use localStorage for that, thanks for the approval tho.

PR #7431, opened 2026-01-23
impr(quote-search): use rtl direction when showing rtl text in quote search modal (@Leonabcd123)

Leonabcd123:


byseif21:
I think this will need a fix for the tooltips now when they are in the left, also we could make this for the input search itself too, wdyt?

PR #7434, opened 2026-01-23
fix: stuck hover on focus for header and restart button (@byseif21)

byseif21:
starting the test while hovering header buttons or restart button leave them stuck in a hover on focus


frontend/src/styles/test.scss
```diff
@@ -1541,14 +1541,12 @@
...
-  }
-  #mobileTestConfigButton {
+  #testModesNotice,
+  #testConfig,
+  #mobileTestConfigButton,
+  #restartTestButton {
     opacity: 0 !important;
+    pointer-events: none !important;
   }
 }
```
Miodec, line 1641: I dont think disabling pointer events on restart test button is a good idea

Miodec (changes requested):
When quick restart is disabled, tabbing over to the restart test button doesn't make it visible.

PR #7436, opened 2026-01-23
fix: fontawesome icon visibility styles (@MirruK)

MirruK:
Add styles for fontawesome icons to respect hidden and invisible classes

### Description

Icons were removed from layers, this caused the .hidden class to not have any effect on icons, this fixes the problem globally


Miodec:
Thanks

PR #7437, opened 2026-01-24
chore: resolve selector warnings (@byseif21)

byseif21:


fehmer:
could you fix this as well? thanks

image

PR #7442, opened 2026-01-24
fix: icons missing on screenshots with css layers (@fehmer, @miodec)

fehmer:
the screenshot library cannot handle fonts in css layers. When we move fontawesome outside the vendor layer it causes other problems (e.g. leaderboard globalRank icon).

image


frontend/src/html/pages/test-result.html
```diff
@@ -6,6 +6,7 @@
       <div class="group wpm">
         <div class="top">
           <div class="text">wpm</div>
+          <i class="fas fa-user"></i>
```
fehmer, line 9: remove after testing

Issue #6472, opened 2025-04-19
Typed Arabic words Don't appear as Points in 'dark note' theme


mu7basha:
### Did you clear cache before opening an issue?

- [x] I have cleared my cache

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

Yes

### Current Behavior

When typing an Arabic word, its letters don't appear as points in 'dark note' theme, however, It shows nothing, just a space.
Also, mistyped letters appear as '-' in Red.

![Image](https://github.com/user-attachments/assets/de0dcbf7-ca59-4914-8d7f-1d4a900e8fb0)

### Expected Behavior

Arabic letters should appear as points, much like how 'dark note' theme works with Latin languages.

![Image](https://github.com/user-attachments/assets/b9a51bfd-fdea-4105-b08c-81cfba428bcb)

### Steps To Reproduce

1. Open monkeytype.
2. Switch keyboard language to Arabic.
3. Start typing.

### Environment

- OS: Windows 10 22H2
- Browser: Zen (Firefox-Based) (64-bit)
- Browser Version: 1.11.4b (Firefox 137.0.2

tech-dipesh:
![Image](https://github.com/user-attachments/assets/e8b61b75-6f16-4828-858f-cdff57734012) But I don't' see any error that you mentioned as it appeared on the error line when I performed typing.

mu7basha:
> ![Image](https://github.com/user-attachments/assets/e8b61b75-6f16-4828-858f-cdff57734012) But I don't' see any error that you mentioned as it appeared on the error line when I performed typing.

This is a problem with the 'dark note' theme, Not the default one. 
'dark note' theme displays letters as Points with typed words. However, Arabic letters appear as Space instead of Points

mu7basha:
> ![Image](https://github.com/user-attachments/assets/e8b61b75-6f16-4828-858f-cdff57734012) But I don't' see any error that you mentioned as it appeared on the error line when I performed typing.

**This is a problem with the 'dark note' theme, Not the default one. **
'dark note' theme displays letters as Points with typed words. However, Arabic letters appear as Space instead of Points with 'dark note' theme enabled.

mu7basha:
[...]

PR #7458, opened 2026-01-28
impr: enable dots typed effect for ligature languages (@byseif21)

byseif21:
fix the limitation of dots effect wouldn't work correctly for ligature-based languages cuz connected letters cannot be rendered as individual dots.

before it was not worth the fix for the theme only now after the new typed effect feat #7360 i think this needed.

* introduces a small helper that breaks ligatures only after a word is finished, allowing the dots effect to render correctly while keeping ligatures intact during typing.

* runs only when needed (on word completion or when the typed effect is switched to dots) and avoids any continuous checks or performance overhead.

related #6472


frontend/src/ts/test/break-ligatures.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (Config.typedEffect !== "dots") return false;
+  if (wordEl.hasClass("broken-ligatures")) return false;
+
+  return !!wordEl.native.closest(".withLigatures");
+}
+
+function applyIfNeeded(wordEl: ElementWithUtils): void {
+  if (!canBreak(wordEl)) return;
+
+  const { width } = wordEl.native.getBoundingClientRect();
```
Leonabcd123: Maybe use `wordEl.screenBounds()` instead?

frontend/src/ts/test/break-ligatures.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (!wordEl.hasClass("broken-ligatures")) return;
+  wordEl.removeClass("broken-ligatures");
+  wordEl.setStyle({ width: "" });
+}
+
+export function set(
+  wordEl: ElementWithUtils,
+  state: "broken" | "normal",
+): void {
+  state === "normal" ? reset(wordEl) : applyIfNeeded(wordEl);
```
Leonabcd123: What do you think about converting `state` to a boolean `areLigaturesBroken`?

`areLigaturesBroken ? applyIfNeeded(wordEl) : reset(wordEl);`

frontend/src/ts/test/test-ui.ts
```diff
@@ -2072,11 +2075,13 @@ ConfigEvent.subscribe(({ key, newValue }) => {
       "colorfulMode",
       "showAllLines",
       "fontSize",
+      "fontFamily",
```
Leonabcd123, line 2079: We now run `updateWordWrapperClasses` unnecessarily when a user changes `fontFamily`.

byseif21: the thing is I think I still need some of the layout updates it triggers when fontFamily changes. I’ll give it some manual testing to see

Leonabcd123: Can we do this?

```
if (key !== "fontFamily"){
  updateWordWrapperClasses();
}
Ligatures.update(key, wordsEl);
```

byseif21: kinda feel like `fontFamily` should also trigger `updateWordWrapperClasses` but can’t really prove it visually enough till now, so yeah ok I’ll leave it limited for now

frontend/src/ts/test/break-ligatures.ts
```diff
@@ -0,0 +1,47 @@
...
+  const shouldReset =
+    !wordsEl.hasClass("withLigatures") ||
+    Config.typedEffect !== "dots" ||
+    key === "fontFamily" ||
+    key === "fontSize";
+
+  if (shouldReset) {
+    words.forEach(reset);
+  }
+  words.forEach(applyIfNeeded);
```
byseif21, line 57: check if ok? or heavy and should we do batching instead?

Leonabcd123: I'd do this:

```
words.forEach((word) => {
  if (shouldReset) reset(word);
  applyIfNeeded(word);
})
```

(not tested)

frontend/src/ts/test/break-ligatures.ts
```diff
@@ -0,0 +1,47 @@
+import Config from "../config";
+import { ElementWithUtils } from "../utils/dom";
+
+function canBreak(wordEl: ElementWithUtils): boolean {
+  if (Config.typedEffect !== "dots") return false;
+  if (wordEl.hasClass("broken-ligatures")) return false;
+
+  return !!wordEl.native.closest(".withLigatures");
```
Leonabcd123: Consider replacing with this:

`return wordEl.getParent().hasClass("withLigatures");`

byseif21: why? `closest` faster and better, `getParent()` creates a new `ElementWithUtils` object wrapper every single time

frontend/src/ts/test/break-ligatures.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function set(
+  wordEl: ElementWithUtils,
+  areLigaturesBroken: boolean,
+): void {
+  areLigaturesBroken ? applyIfNeeded(wordEl) : reset(wordEl);
+}
+
+export function update(key: string, wordsEl: ElementWithUtils): void {
+  if (!["typedEffect", "fontFamily", "fontSize"].includes(key)) return;
```
Leonabcd123: When looking at the code that's calling `Ligatures.update`, it's not clear that we only run `Ligatures.update` when these 3 keys are updated, and not when all of the other keys listed in the [listener](https://github.com/byseif21/monkeytype/blob/5b72288b29150e3ec13a8bc2b2988c9ea3f43d79/frontend/src/ts/test/test-ui.ts#L2068-L2080) are updated. What do you think about moving this check to that listener?

frontend/src/ts/test/break-ligatures.ts
byseif21: i meant performance wise

frontend/src/ts/test/break-ligatures.ts
Leonabcd123: `closest` seems less readable, as it doesn't make it clear what element might have the `withLigatures` class. Also (it might just be me), but I really hate the `!!` syntax.

frontend/src/ts/test/break-ligatures.ts
Leonabcd123: Performance wise we're running on the `words` twice when `shouldReset` is true, while we can do it one pass. I don't see why go with this approach rather than doing it in one loop.

byseif21: what do you think about batching then? too much or worth it?

frontend/src/ts/test/break-ligatures.ts
byseif21: let's do the spin

Leonabcd123: I lost, so keep current approach.

frontend/src/ts/test/break-ligatures.ts
Leonabcd123: I'd say it's worth it.

frontend/src/ts/test/break-ligatures.ts
Miodec: CEO override - i prefer `.getParent()`

frontend/src/ts/test/break-ligatures.ts
```diff
@@ -0,0 +1,58 @@
...
+  const shouldReset =
+    !wordsEl.hasClass("withLigatures") ||
+    Config.typedEffect !== "dots" ||
+    key === "fontFamily" ||
+    key === "fontSize";
+
+  if (shouldReset) {
+    words.forEach(reset);
+  }
+  words.forEach(applyIfNeeded);
```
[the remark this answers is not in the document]
byseif21: intentional not unnecessary cuz font/size changes need width recalculation,
`canBreak` already early returns when conditions aren't met so we good

PR #7463, opened 2026-02-01
chore(components): add LoadingCycle, DiscordAvatar and User, improve Button (@fehmer)

fehmer:
- **add isActive and classList to Button, move H2 and H3 to Headers**
- **add aria-label and router-link to button**
- **move tests to correct folders**
- **add LoadingCircle**
- **add DiscordAvatar and User**


frontend/src/ts/components/pages/AboutPage.tsx
```diff
@@ -50,7 +51,44 @@ export function AboutPage(): JSXElement {
   return (
     <Show when={isOpen}>
       <div class="content-grid grid gap-8">
-        <section class="text-sub text-center">
+        <section>
+          {/*TODO remove after testing */}
```
fehmer: remove

PR #7469, opened 2026-02-02
fix: funbox deselection not clearing effects without refresh (@prajith5)

### LINKED ISSUE #7468: Funbox deselection does not reset effects without refresh

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

d1rsh

### Account config

_No response_

### Current Behavior

When multiple Funbox effects are enabled and then deselected, the UI shows them as disabled, but the effects are still applied internally. The Funbox state does not fully reset. A full page refresh is required to return to the correct state.

### Expected Behavior

When Funbox effects are deselected, they should be fully disabled immediately, and no Funbox effects should remain active without requiring a page refresh.

### Steps To Reproduce

1. Open Monkeytype
2. Enable Funbox
3. Select multiple Funbox effects (any combination - eg: nausea + upside down)
4. Deselect the enabled Funbox effects
5. Observe that the UI shows Funbox effects as disabled

Start typing — previously selected Funbox effects are still active

Refresh the page → Funbox state resets correctly

### Environment

- OS: Arch Linux
- Browser: Mozilla Firefox
- Browser Version: 147.0.2

### Anything else?

_No response_

prajith5:
Fixes an issue where Funbox effects remained active internally after being
deselected until a page refresh.

This change compares the previous and next activation state of a Funbox
and ensures:
- clearGlobal() is called when a Funbox is deselected
- applyGlobalCSS() is called only when a Funbox is activated

This guarantees Funbox effects reset immediately without requiring a refresh.

Fixes #7468

### DIFF
--- a/frontend/src/ts/test/funbox/funbox.ts
+++ b/frontend/src/ts/test/funbox/funbox.ts
@@ -33,11 +33,14 @@ export function toggleScript(...params: string[]): void {
 }
 
 export function setFunbox(funbox: FunboxName[]): boolean {
-  if (funbox.length === 0) {
-    for (const fb of getActiveFunboxesWithFunction("clearGlobal")) {
-      fb.functions.clearGlobal();
-    }
+  const previousFunboxes = getActiveFunboxNames();
+
+  const removedFunboxes = previousFunboxes.filter((fb) => !funbox.includes(fb));
+
+  for (const fb of removedFunboxes) {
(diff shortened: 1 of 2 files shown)


Miodec:
This PR doesn't seem to fix the issue

Miodec:
Fixed in #7575

PR #7474, opened 2026-02-05
fix(blind-mode): blind mode not working in 9009 theme (@fehmer)

### LINKED ISSUE #7473: 9009 theme shows mistakes in blind mode

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

shows errors when in blind mode
Image

### Expected Behavior

shouldnt show errors if in blind mode

### Steps To Reproduce

1. turn on blind mode
2. switch to theme 9009
3. make errors on a test

### Environment

- OS:
- Browser:
- Browser Version:

### Anything else?

_No response_

Leonabcd123:
### Description

Remove unneeded styles in 9009 theme.

Fixes #7473

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -354,8 +354,8 @@
         display: none;
       }
       & letter.incorrect {
-        color: var(--correct-letter-color) !important;
-        animation: var(--correct-letter-animation) !important;
+        color: var(--correct-letter-color);
+        animation: var(--correct-letter-animation);
       }
       & .hints hint {
         display: none;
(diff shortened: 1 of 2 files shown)


fehmer:
Wouldn't it be better to clean up the 9009 css? I think the first three rules we don't need?

https://github.com/monkeytypegame/monkeytype/blob/master/frontend/static/themes/9009.css#L1-L11

Leonabcd123:
> Wouldn't it be better to clean up the 9009 css? I think the first three rules we don't need?
> 
> https://github.com/monkeytypegame/monkeytype/blob/master/frontend/static/themes/9009.css#L1-L11

You're right, the first three rules are already applied in `test.scss`.

Issue #5181, opened 2024-03-04
Caps lock pop-up appears incorrectly


jxdv:
### Did you clear cache before opening an issue?

- [X] I have cleared my cache

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Does the issue happen when logged in?

No

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

No

### Does the issue happen in incognito mode when logged out?

Yes

### Account name

Not using an account.

### Account config

Not using an account.

### Current Behavior

When you start writing and turn on caps-lock and refresh afterwards, there's no caps-lock pop-up, but when you turn off the caps-lock then the pop-up appears, so basically the exact opposite happens.

### Expected Behavior

The pop-up appears correctly.

### Steps To Reproduce

1. Open monkeytype
2. Start typing out words
3. Enable Caps-lock
4. Hit F5
5. Turn off Caps-lock

### Environment

- OS: Ubuntu 22.04.3 LTS
- Browser: Firefox (snap)
- Browser Version: Firefox version 123.0 (64-bit)

fehmer:
Hi, it should fix itself when you start typing.

Miodec:
Browsers have no ability to detect the current state of caps lock - they only get this information when a key is pressed.

PR #7477, opened 2026-02-06
fix(input): correct caps lock indicator state (@VardanRattan)

### LINKED ISSUE #7472: Caps Lock indicator becomes inverted when page is opened with Caps Lock already enabled

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

JATIN917

### Account config

_No response_

### Current Behavior

- Initial Caps Lock state is shown incorrectly if the page is opened while Caps Lock is enabled.
- After toggling Caps Lock once, the indicator displays the opposite of the actual state.

### Expected Behavior

- The Caps Lock indicator should correctly reflect the actual Caps Lock state after user interaction.
- The indicator should not become inverted after toggling Caps Lock.

### Steps To Reproduce

1. Enable Caps Lock outside Monkeytype.
2. Open Monkeytype.
3. Now turn off Caps Lock indicator shows Caps Lock is ON.
4. Now the state becomes inconsistent with the actual caps lock toggle. 

### Environment

- OS: Windows 11
- Browser: Chrome 
- Browser Version: in every browser

### Anything else?

_No response_

VardanRattan:
### Description

Fixes an issue where the Caps Lock indicator could become inverted if the page was opened while Caps Lock was already enabled.

Previously, the Caps Lock state was toggled internally on `CapsLock` key events, which caused incorrect initialization and desynchronization from the actual OS modifier state. This change removes the toggle-based logic and derives the Caps Lock state exclusively from `KeyboardEvent.getModifierState("CapsLock")`, ensuring the indicator always reflects the real modifier state after the first key event.

### Checks

- [ ] Adding quotes?
- [ ] Adding a language?
- [ ] Adding a theme?
  - [ ] (optional) Add theme css file to `frontend/static/themes`
  - [ ] Add some screenshots of the theme, especially with different test settings (colorful, flip colors) to your pull request
- [ ] Adding a layout?
- [ ] Adding a font?
  - Make sure to follow the [fonts documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/FONTS.md)
  - [ ] Add font file to `frontend/static/webfonts`

Closes #7472 and #5181

### DIFF
--- a/frontend/src/ts/test/caps-warning.ts
+++ b/frontend/src/ts/test/caps-warning.ts
@@ -51,5 +51,9 @@ function update(event: KeyboardEvent): void {
 document.addEventListener("keyup", update);
 
 document.addEventListener("keydown", (event) => {
-  if (Misc.isMac()) update(event);
+  if (Misc.isMac()) {
+    update(event);
+  } else if (Misc.isLinux() && event.key === "CapsLock") {
+    update(event);
+  }
 });


Leonabcd123:
This is what we originally used, but it caused the Caps Lock warning to be stuck on Linux (see 64a0bdc3e2e58af285053997be7a6206e1a9cb30).

VardanRattan:
@Leonabcd123  Thanks for pointing that out — I looked into the Linux issue from #7195.

From what I can tell, the problem there was related to modifier state not being updated at the time the handler ran, causing the warning to remain visible until a subsequent key event.

In this change, the Caps Lock state is still derived from getModifierState("CapsLock"), but the update runs on keyup for all platforms (and additionally on keydown for macOS), which should allow the modifier state to settle correctly on Linux before being read.

I agree the toggle workaround avoided the Linux timing issue, but it also caused desynchronization when Caps Lock was already enabled on page load. If there’s a specific Linux scenario where keyup still isn’t sufficient, I’m happy to adjust the event handling further without reintroducing internal toggling.

fehmer:
With this the caps lock warning persists when pressing caps lock on linux

VardanRattan:
@fehmer  Thanks for testing — that makes sense. 
This looks like the same Linux timing issue as #7195, where the CapsLock modifier state isn’t updated at the time the event fires. I’ll update this to still derive state from getModifierState("CapsLock"), but defer the read for Linux CapsLock events so the modifier state has time to settle, instead of toggling internal state. That should clear the warning immediately without reintroducing inversion.

fehmer:
Hi @VardanRattan , i checked the latest change, caps lock is still stuck on linux.

VardanRattan:
Thanks for checking.

Could you share how you tested this on Linux?
Specifically:
distro + desktop environment (X11 / Wayland)
browser
whether Caps Lock was already enabled before page load, or toggled after

That’ll help narrow down whether this is the same modifier-timing issue as #7195 or a different edge case.

fehmer:
ubuntu 25.10, gnome, wayland
chrome 144
caps was disabled on page entry.

VardanRattan:
Thanks — that helps a lot.

Ubuntu + GNOME + Wayland + Chrome is consistent with the Linux modifier-timing issue from #7195. In that setup, getModifierState("CapsLock") can still return the previous state at the time the CapsLock key event fires, which can leave the warning stuck until another key event.

I’ll adjust the Linux handling so the CapsLock modifier read is deferred for CapsLock events, allowing the modifier state to settle before updating visibility. That should clear the warning immediately on toggle without relying on state inversion. I’ll update the PR shortly.

fehmer:
HI @VardanRattan   i checked the latest changes and it is still stuck. I think #7478 is the better approach

VardanRattan:
@fehmer  Thanks for re-testing and confirming.

If #7478 behaves correctly on Ubuntu + GNOME + Wayland + Chrome, I’m good with going that route for now. My intent here was to avoid state inversion by deferring the modifier read, but I agree practical correctness on Linux is the priority.

Happy to close this PR or rebase it later if needed.

Miodec:
Closing in favor of #7478

PR #7478, opened 2026-02-06
fix(caps-warning): caps state isn't always correct (@Leonabcd123, @fehmer)

### LINKED ISSUE #5987: CAPS lock notification does not goes away when i have turn off CAPS LOCK

- [X] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

biswa_108

### Account config

_No response_

### Current Behavior

when you press caps lock the notification for caps lock on is there but when you turn it off the notification stays .

### Expected Behavior

When i turn on caps lock the notification should be there and when i turn capslock off the notification should be gone .

### Steps To Reproduce

1. Press caps lock ( to turn caps lock on )
2. Press Caps lock ( to turn it off )
3. the notification persists

### Environment

- OS: Ubuntu 22.04
- Browser: Brave
- Browser Version : Brave 1.71.118

### Anything else?

image
After turning on CAPS lock
image
After turning off CAPS lock

Leonabcd123:
### Description

Account for platform specific behavior.

Fixes #5987, #7472

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@date-fns/utc": "1.2.0",
+    "@leonabcd123/modern-caps-lock": "2.0.3",
     "@monkeytype/contracts": "workspace:*",
     "@monkeytype/funbox": "workspace:*",
     "@monkeytype/schemas": "workspace:*",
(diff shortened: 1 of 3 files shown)


VardanRattan:
This works, but it does reintroduce internal inversion for Linux CapsLock events.

Conceptually, the issue on Linux is timing (modifier state not settled at event time), not that getModifierState("CapsLock") is incorrect. A deferred read would avoid platform-specific inversion and keep CapsLock state fully derived from OS state.

Might be worth considering if we want to avoid encoding Linux timing assumptions into logic long-term.

PR #7482, opened 2026-02-08
feat: add practice typing mode (no stats, xp, or pb) (@Vishal27alpha)

Vishal27alpha:
### What
Adds a new `practice` typing mode that allows users to type freely without affecting:
- personal bests
- XP
- leaderboards
- result history

### Why
This enables low-pressure practice sessions while preserving competitive integrity of stats.

### How
- Extended shared schemas to include `practice` mode
- Backend ignores practice results for persistence, XP, PB, and leaderboards
- Frontend hides practice mode from result filters

### Notes
- API response shape remains unchanged
- No database or contract changes required

Relevant files:
- packages/schemas/src/shared.ts
- backend/src/api/controllers/result.ts
- frontend/src/ts/constants/default-result-filters.ts


fehmer:
You can already turn off result saving in the comand line `result saving off`.

Vishal27alpha:
Thanks for pointing that out, I wasn’t aware this already existed via the command line.

That explains the gap I felt from a user perspective. Would you be open to reframing this PR to focus on exposing the existing practice mode in the UI (making it more discoverable), rather than adding new backend behavior?

Happy to adjust or close the PR based on your preference.

Miodec:
> Thanks for pointing that out, I wasn’t aware this already existed via the command line.
> 
> That explains the gap I felt from a user perspective. Would you be open to reframing this PR to focus on exposing the existing practice mode in the UI (making it more discoverable), rather than adding new backend behavior?
> 
> Happy to adjust or close the PR based on your preference.

Lets add a "result saving" settin to the account section in the settings page. In the description mention the word 'practice' so its easier to find. We should also add a 'practice' alias to the result saving commandline commands.

Vishal27alpha:
Thanks a lot for the clarification that makes perfect sense.
I’m happy to try reworking this PR in that direction:

-Add a “Result saving” toggle in the Account section of Settings
-Explicitly mention practice in the UI copy for better discoverability
-Add practice as an alias to the existing command-line commands

I’ll start exploring this and update the PR accordingly, unless you’d prefer this to be handled as a separate PR instead.

Miodec:
> Thanks a lot for the clarification that makes perfect sense. I’m happy to try reworking this PR in that direction:
> 
> -Add a “Result saving” toggle in the Account section of Settings -Explicitly mention practice in the UI copy for better discoverability -Add practice as an alias to the existing command-line commands
> 
> I’ll start exploring this and update the PR accordingly, unless you’d prefer this to be handled as a separate PR instead.

Same pr is fine

Vishal27alpha:
Hi! I restarted the implementation from a clean state.
This PR now adds:

• A **Result saving toggle** in Account Settings
• The word **“practice”** in the UI description for discoverability
• A **practice alias** for the result-saving command

All checks are passing and the branch is updated with latest master.

Please let me know if any adjustments are needed!

frontend/src/html/pages/account-settings.html
```diff
@@ -133,6 +133,25 @@
...
+            <div class="title">
+              <i class="fas fa-save"></i>
+              <span>result saving</span>
+            </div>
+            <div class="text">
+              Toggle result saving. When disabled, results will not be saved (
+              <b>practice mode</b>
+              ).
+            </div>
+            <div class="buttons">
```
fehmer: I'd just use two buttons like we do on the settings page for toggles:

Image

frontend/src/ts/ape/adapters/ts-rest-adapter.ts
```diff
@@ -29,9 +29,14 @@ function buildApi(timeout: number): (args: ApiFetcherArgs) => Promise<{
 }> {
   return async (request: ApiFetcherArgs) => {
     try {
-      const token = await getIdToken();
-      if (token !== null) {
-        request.headers["Authorization"] = `Bearer ${token}`;
+      // ===== DEV AUTH (LOCAL ONLY) =====
```
fehmer: please revert

frontend/src/ts/modals/simple-modals.ts
```diff
@@ -816,6 +819,25 @@ list.optOutOfLeaderboards = new SimpleModal({
   },
 });
 
+list.toggleResultSaving = new SimpleModal({
```
fehmer: unused

frontend/src/ts/pages/account-settings.ts
```diff
@@ -147,11 +149,24 @@ function updateAccountSections(): void {
   }
 }
 
+function updateResultSavingToggle(): void {
+  const toggle = pageElement.qs(
+    "#toggleResultSaving",
+  ) as HTMLInputElement | null;
+  if (!toggle) return;
```
Leonabcd123: Consider using this:

```
pageElement.qs<HTMLInputElement>("#toggleResultSaving")?.setChecked(TestState.savingEnabled);
pageElement
  .qs(".section.resultSaving .toggleLabel")
  ?.setText(TestState.savingEnabled ? "on" : "off");
```

Vishal27alpha:
Thanks for the review!
I’ve addressed the requested changes:

reverted the dev-auth change
removed unused modal/imports
updated the toggle implementation to match existing settings toggles
synced branch with latest master

All checks are passing now.

fehmer:
hi @Vishal27alpha ,

i talked to @Miodec. There was a misunderstanding on where to put the setting. Mio requested the new setting to be added to the settings page under account settings, but that got moved. Please add it to the settings page after presents and before test difficulty. https://monkeytype.com/settings?highlight=difficulty

frontend/src/ts/commandline/lists/result-saving.ts
fehmer, line 1: while testing I noticed that the active state is missing on this, could you add it?

```diff
@@ -13,6 +13,7 @@ const subgroup: CommandsSubgroup = {
         TestState.setSaving(false);
         void ModesNotice.update();
       },
+      active: () => !TestState.savingEnabled,
     },
     {
       id: "setResultSavingOn",
@@ -22,6 +23,7 @@ const subgroup: CommandsSubgroup = {
         TestState.setSaving(true);
         void ModesNotice.update();
       },
+      active: () => TestState.savingEnabled,
     },
   ],
 };
```

Vishal27alpha:
Thanks for the clarification @fehmer!

I’ve moved the result saving setting to the main settings page (between presets and test difficulty) and added the missing active state for the command-line command.

frontend/src/html/pages/settings.html
```diff
@@ -106,6 +106,33 @@
         <button class="addPresetButton fullWidth">add preset</button>
       </div>
     </div>
+    <div
```
fehmer, line 109: can you simplify this and get rid of the custom styles?

```html
    <div
      class="section resultSaving needsAccount"
      data-section-id="resultSaving"
    >
      <div class="groupTitle">
        <i class="fas fa-save"></i>
        <span>result saving</span>
        <button class="text" tabindex="-1">
          <i class="fas fa-fw fa-link"></i>
        </button>
      </div>
      <div class="text">
        Toggle result saving. When disabled, results will not be saved (
        <b>practice mode</b>
        ).
      </div>
      <div class="buttons">
        <button data-value="off">off</button>
        <button data-value="on">on</button>
      </div>
    </div>
```

frontend/src/ts/pages/account-settings.ts
```diff
@@ -122,8 +122,8 @@ function updateTabs(): void {
       pageElement.qs(`.tab[data-tab="${state.tab}"]`)?.addClass("active");
     },
   );
-  pageElement.qsa("button")?.removeClass("active");
-  pageElement.qs(`button[data-tab="${state.tab}"]`)?.addClass("active");
+  pageElement.qsa(".tabs button")?.removeClass("active");
```
fehmer: why do we need this change?

frontend/src/ts/test/test-state.ts
```diff
@@ -5,7 +5,22 @@ export let isRepeated = false;
 export let isPaceRepeat = false;
 export let isActive = false;
 export let activeChallenge: null | Challenge = null;
-export let savingEnabled = true;
+const savingEnabledStorageKey = "resultSavingEnabled";
+
+function getInitialSavingEnabled(): boolean {
```
fehmer: what is this change for?

frontend/src/ts/pages/account-settings.ts
Vishal27alpha: that change was added when the result‑saving toggle lived in Account Settings so the tab buttons didn’t clobber its active state after reload. Now that the toggle is in Settings, it’s no longer necessary here. I can revert it and keep the original line if you prefer.

frontend/src/ts/test/test-state.ts
Vishal27alpha: This change makes the “result saving” toggle persist across reloads by reading the last value from localStorage.
If storage isn’t available (SSR, privacy mode), it safely defaults to `true` to preserve current behavior.
So the only purpose is persistence ,there is no behavior change unless the user has previously toggled it.

Let me know,If you have a suggestion or want me to revert it.

fehmer: @Miodec  do we want to persist this? if yes in local storage or add to the user config

Miodec: Im not sure, what do you guys think?

fehmer: Most settings persist on the user config and apply across devices. If there is no good reason to handle this special I'd do the same here.

Vishal27alpha: Good point. This toggle is more of a per‑user preference than a per‑device one, so persisting it in config makes more sense and keeps behavior consistent across devices. I can remove the localStorage logic and wire it through config instead (or revert it entirely if you don’t want persistence at all). Let me know which direction you prefer

Miodec: yeah move it to config

Vishal27alpha: Moved it to user config and removed the localStorage persistence 👍

frontend/src/ts/commandline/lists/result-saving.ts
fehmer, line 1: With the result savings being a regular config this file can be deleted

frontend/src/ts/commandline/commandline-metadata.ts
```diff
@@ -205,6 +205,11 @@ export const commandlineConfigMetadata: CommandlineConfigMetadataObject = {
       options: "fromSchema",
     },
   },
+  resultSavingEnabled: {
```
fehmer: this needs the alias we defined in the `lists/result-saving.ts` before.

```ts
  resultSavingEnabled: {
    subgroup: {
      options: "fromSchema",
      alias: (val) => (val ? "enabled" : "disabled"),
    },
    alias: "results practice incognito",
  },
```

frontend/src/ts/modals/simple-modals-base.ts
fehmer, line 1: revert

frontend/src/ts/pages/account-settings.ts
fehmer: please revert

fehmer:
Hi @Vishal27alpha , nice work :+1: 

To get rid of the `lists/result-saving.ts` and fully use the command-metadata this changes are needed:

- in `commandline/lists.ts` 
  - add `"resultsSavingEnabled"` as first command in the behavior section, line 103
  - remove the `ResultSavingCommands` from line 101 and remove the import
- in `elements/modes-notice.ts`
  - change the commands property from `resultSaving` to `resultSavingEnabled`, line 55

frontend/src/html/pages/account-settings.html
fehmer, line 1: revert

frontend/src/ts/modals/simple-modals.ts
fehmer, line 1: revert

frontend/src/ts/firebase.ts
fehmer, line 1: revert

Vishal27alpha:
@fehmer I’ve reverted the requested changes and updated the implementation accordingly.

fehmer (dismissed):
nice work 👍

Vishal27alpha:
Thanks! Really appreciate the review

Miodec:
Thanks

PR #7483, opened 2026-02-08
Fix Caps Lock indicator inversion on initial load

harikrishna0315:
### What was fixed
The Caps Lock indicator could become inverted when the page was loaded with Caps Lock already enabled, because the state was toggled instead of synced.

### What changed
- Updated `caps-warning.ts` to prioritize `event.getModifierState("CapsLock")`
- This ensures the indicator reflects the actual Caps Lock state rather than assuming a toggle
- Added a safe fallback for environments where `getModifierState` is unavailable

### Verification
- Manually reproduced the issue by loading the page with Caps Lock enabled
- Confirmed the indicator now shows the correct state immediately
- Verified toggling Caps Lock behaves correctly afterward


harikrishna0315:
Hi! The CI is asking for a Conventional Commits–style PR title.
I’m unable to edit the PR title from my side.

Could a maintainer please rename this PR to:

fix(caps-warning): correct Caps Lock indicator inversion (@harikrishna0315)

Thanks!

fehmer:
hi @harikrishna0315 , thank you for your contribution. There is already a  PR #7478 fixing this issue

PR #7494, opened 2026-02-11
fix(practise-words): treat __proto__ as a normal word in practise words (@Leonabcd123)

Leonabcd123:
### Description

1. Create a custom text with `__proto__` as the only word
2. Make a mistake and complete the test
3. Click on "Practise words"
4. Click on "Start"
5. You will get an alert saying you didn't miss any words

### DIFF
--- a/frontend/__tests__/components/common/Button.spec.tsx
+++ b/frontend/__tests__/components/common/Button.spec.tsx
@@ -141,21 +141,6 @@ describe("Button component", () => {
     expect(button).not.toHaveClass("button");
   });
 
-  it("applies textButton class when type is text", () => {
-    const { container } = render(() => (
-      <Button
-        onClick={() => {
-          //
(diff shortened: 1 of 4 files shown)


frontend/src/ts/test/test-input.ts
Miodec, line 1: Type should start with a capital letter. Also i would add a comment on why its done this way.

Leonabcd123: We use `Object.create(null)` to make sure that `__proto__` won't have any special meaning, and we keep the original record type.

Miodec: I mean in the code haha

Leonabcd123: Oh I thought you asked why I'm doing it like this lol.

PR #7502, opened 2026-02-13
chore: do some todos (@Leonabcd123)

Leonabcd123:


backend/src/dal/user.ts
```diff
@@ -589,7 +589,7 @@ export async function updateLastHashes(
     { uid },
     {
       $set: {
-        lastReultHashes: lastHashes, //TODO fix typo
+        lastResultHashes: lastHashes,
```
fehmer: this one is not so easy. it would need a db-migration or soft-migration and is technically a breaking change in the api.so we should increase the compat version if we want to rename it.

t.b.h i'd just keep it

frontend/src/ts/db.ts
```diff
@@ -44,9 +44,6 @@ export class SnapshotInitError extends Error {
   ) {
     super(message);
     this.name = "SnapshotInitError";
-    // TODO INVESTIGATE
```
fehmer, line 45: let's use this style we use in most places

```ts
export class SnapshotInitError extends Error {
  public responseCode: number;

  constructor(message: string, responseCode: number) {
    super(message);
    this.name = "SnapshotInitError";
    this.responseCode = responseCode;
  }
}
```

packages/release/src/index.js
```diff
@@ -263,14 +261,12 @@ const main = async () => {
...
-    // oxlint-disable-next-line prefer-nullish-coalescing
-    const channelName = name.trim() || "preview";
+    const channelName = name.trim() ?? "preview";
 
     const expirationTime = readlineSync.question(
       "Enter expiration time (e.g., 2h, default: 1d): ",
     );
-    // oxlint-disable-next-line prefer-nullish-coalescing
-    const expires = expirationTime.trim() || "1d";
+    const expires = expirationTime.trim() ?? "1d";
```
Miodec: This changes the behavior. If name is an empty string, it now won't get replaced with `preview`

PR #7505, opened 2026-02-15
fix(account): filter buttons not working (@fehmer)

fehmer:
## unable to select filter presets, chart not updating

reproduce:

- create two presets e.g. `all` and `last three month`
- switch between presets

The current preset doesnt get highlighted and the chart doesnt get updated.

## advanced filters 
- clear filters not working

### DIFF
--- a/frontend/src/ts/elements/account/result-filters.ts
+++ b/frontend/src/ts/elements/account/result-filters.ts
@@ -166,7 +166,7 @@ export async function setFilterPreset(id: string): Promise<void> {
 
   // make current filter presest button active
   qsa(
-    `.pageAccount .group.presetFilterButtons .filterBtns .filterPresets .select-filter-preset[data-id=${id}]`,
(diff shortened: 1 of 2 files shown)


frontend/src/ts/elements/account/result-filters.ts
```diff
@@ -564,14 +551,30 @@ for (const el of qsa(`
...
-    } else if ((e.target as HTMLElement).tagName === "BUTTON") {
-      if (e.shiftKey) {
-        setAllFilters(group, false);
-        filters[group][filter] =
-          true as ResultFilters[typeof group][typeof filter];
-      } else {
-        toggle(group, filter);
-        // filters[group][filter] = !filters[group][filter];
+    } else {
+      const group = (e.target as HTMLElement).parentElement?.getAttribute(
```
Leonabcd123: Can we use the already existing variable `childTarget` instead of `e.target`?

PR #7555, opened 2026-02-28
impr(quotes): only check short length of new quotes (@Leonabcd123)

Leonabcd123:
### Description

When running through all the quotes:

- Check whether the user has added quotes to the current language
- If so, run through all the newly added quotes and check whether their length is shorter than 60 characters.

### DIFF
--- a/frontend/scripts/check-assets.ts
+++ b/frontend/scripts/check-assets.ts
@@ -222,7 +222,9 @@ async function validateQuotes(): Promise<void> {
 
     if (
       currentLanguageData !== null &&
-      currentLanguageData.quotes.length < quoteData.quotes.length
+      (currentLanguageData.quotes.length < quoteData.quotes.length ||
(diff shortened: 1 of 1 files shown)


frontend/scripts/check-assets.ts
```diff
@@ -197,6 +216,19 @@ async function validateQuotes(): Promise<void> {
...
+    // check if pr added quotes in this language
+    let idOfFirstAddedQuote = 0;
+    const currentLanguageQuotes = await fetchQuotes(quotefilename);
+
+    if (
+      currentLanguageQuotes !== null &&
+      currentLanguageQuotes.quotes.length < quoteData.quotes.length
+    ) {
+      idOfFirstAddedQuote =
+        (currentLanguageQuotes.quotes[currentLanguageQuotes.quotes.length - 1]
```
fehmer: quotes are not guaranteed to be ordered by id

PR #7556, opened 2026-03-01
fix(themes): nav bar isn't completely hidden in some themes (@Leonabcd123)

Leonabcd123:
### Description

Example problematic theme: `taro`.

Extends #7511


Miodec:
Looks like this got fixed in the solid Nav rewrite.

PR #7570, opened 2026-03-03
fix(quote-search): length filter losing styles after returning from chained modal (@byseif21)

byseif21:
The "filter by length" dropdown in the quote search modal reverts to an unstyled native `<select>` after returning from a chained modal e.g(report, submit).

* Steps to reproduce:
1- Open the quote search modal 
2- Click the report flag (or "submit a quote" button) to open a chained modal
3- Close/dismiss that modal to return to the quote search modal
4- The "filter by length" dropdown is now a raw unstyled scrollable `<select>` list


Miodec:
Should we also make it remember the chosen length while we're at it?

PR #7574, opened 2026-03-03
fix(themes): update stale footer selectors after SolidJS migration (@byseif21)

byseif21:
causing some effects to stop working

### DIFF
--- a/frontend/static/themes/rainbow_trail.css
+++ b/frontend/static/themes/rainbow_trail.css
@@ -64,14 +64,6 @@ footer button:hover > .relative > i {
   transition: all 1s ease;
 }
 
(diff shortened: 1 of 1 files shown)


fehmer:
obsolete after https://github.com/monkeytypegame/monkeytype/pull/7564 but I am pretty sure you'll find some styles we missed :)

byseif21:
> obsolete after #7564 but I am pretty sure you'll find some styles we missed :)

yeah i noticed but i was not sure is that flipping effect, meant to be removed too?

Miodec:
Yes, obsolete now that https://github.com/monkeytypegame/monkeytype/pull/7564 is merged

PR #7575, opened 2026-03-03
fix: remove redundant clearGlobal/applyGlobalCSS calls (@byseif21)

byseif21:
### Description

`toggleFunbox()`/`setFunbox()`  manually calls `applyGlobalCSS()`/`clearGlobal()`, but `configToggleFunbox()`/`setConfig()` already fires a `ConfigEvent` whose subscriber does the same. this causes some warnings e.g duplicate `#scanline` elements when clearing the CRT funbox via the command line ("none")


Miodec:
Thanks

PR #7576, opened 2026-03-04
refactor: solid notifications (@miodec)

Miodec:
Brr

### DIFF
--- a/frontend/__tests__/stores/notifications.spec.ts
+++ b/frontend/__tests__/stores/notifications.spec.ts
@@ -1,0 +1,322 @@
+import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
+import {
+  addNotificationWithLevel,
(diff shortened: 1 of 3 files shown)


frontend/src/ts/components/layout/overlays/Notifications.tsx
```diff
@@ -0,0 +1,143 @@
...
+  clearAllNotifications,
+} from "../../../stores/notifications";
+import { cn } from "../../../utils/cn";
+import { Anime } from "../../common/anime/Anime";
+import { AnimePresence } from "../../common/anime/AnimePresence";
+import { AnimeShow } from "../../common/anime/AnimeShow";
+import { Fa, FaProps } from "../../common/Fa";
+
+const levelConfig = {
+  0: {
```
fehmer: would it be a good time to change them from magic number to readable string?

PR #7590, opened 2026-03-08
fix(quick-restart): quick restart warning firing for wrong events in long tests (@Leonabcd123)

Leonabcd123:
### Description

- Can restart freely before long tests starts without needing to hold shift
- Can change settings such as language in long tests


Miodec:
Thanks

PR #7592, opened 2026-03-08
fix(changelog): prs which aren't adding quotes are treated as such (@Leonabcd123)

Leonabcd123:
Now requires:

- type to be `impr`
- scope to include `quote`
- message to start with the word `add`
- message to end with the word `quote` or the word `quotes`

### Q

But what if prs adding quotes won't be detected? Wouldn't that ruin the changelog?

### A

People should be able to follow the documentation and name their prs properly. The only thing currently ruining the changelog is these false positives.


Miodec:
@claude review

PR #7598, opened 2026-03-09
fix(profile): prevent PB hover details from overlapping base values (@AzureNightlock)

### LINKED ISSUE #7596: [UI Bug] Original score not hidden when hover displays score details

- [x] I have cleared my cache

Yes

No

Yes

No

### Account name

Account: rative
Joined: 8th Sept 2025

### Account config

[userconfig.txt](https://github.com/user-attachments/files/25831463/jsonformatter.txt)

### Current Behavior

- When hovering over my available all time high scores, getting all additional details (wpm, raw, acc, con, date).
- The Score Accuracy Percentage remains visible with a glittering view instead of disappearing.
- This causes duplicate/overlapping score information in the UI.

Image

### Expected Behavior

- When hovering over the score, only the detailed score information should be displayed clearly.
- The UI should not show duplicate or overlapping score values.

### Steps To Reproduce

1. Go to "monkeytype.com"
2. Login in your account
3. Go to "User Stats"
4. Scroll to All time high score section
5. Hover over the score

### Environment

- OS:MacOS Tahoe 26.3.1
- Browser:Chrome
- Browser Version:145.0.7632.160

### Anything else?

https://github.com/user-attachments/assets/326b0434-9ede-4500-abe9-77a31aba0022

AzureNightlock:
### Description
This PR fixes an issue where PB hover details overlap with the base values.

The PB tile renders two layers in the same grid position:
- a base view showing the normal WPM and accuracy
- a detail view showing the expanded PB information on hover

Previously, the detail view faded in on hover, but the base view stayed visible. Because both layers occupy the same space, the values overlapped.

This change makes the PB tile a hover group so both layers respond to the same hover state:
- the base view fades out on hover
- the detail view fades in on hover

This prevents the normal values and hover details from being visible at the same time.
### Changes
- added a hover group to the PB tile container
- fade out the base PB values on hover
- fade in the detailed PB view on hover

### Checks

Closes #7596

### DIFF
--- a/frontend/src/ts/components/pages/profile/UserProfile.tsx
+++ b/frontend/src/ts/components/pages/profile/UserProfile.tsx
@@ -127,7 +127,13 @@ function PbTable<M extends "time" | "words">(props: {
         <For each={bests()}>
           {(item) => (
             <div class="group grid items-center">
-              <div class="col-start-1 row-start-1 text-center transition-opacity group-hover:opacity-0">
+              <div
+                class={
+                  item.pb !== undefined
+                    ? "col-start-1 row-start-1 text-center transition-opacity group-hover:opacity-0"
+                    : "col-start-1 row-start-1 text-center"
+                }
+              >
                 <div class="text-xs text-sub">
                   {item.mode2} {props.mode === "time" ? "seconds" : "words"}
                 </div>
@@ -142,6 +148,7 @@ function PbTable<M extends "time" | "words">(props: {
                   })}
                 </div>
               </div>
+
               <Show when={item.pb !== undefined}>
                 <div class="col-start-1 row-start-1 grid bg-sub-alt text-center text-xs opacity-0 transition-opacity group-hover:opacity-100">
                   <div class="text-sub">


AzureNightlock:
I couldn't fully test this on the real account page locally because I couldn't get the Firebase account setup working on my machine. I did check the behavior using the same PB tile structure and hover logic, so it's partially verified but not fully tested with the account flow. Sorry about the delay and inconvenience.

openvaibhav:
Well the fix is fine as for the verification of this fix you dont actually need to setup whole backend and since setting up Firebase locally isn't feasible, you can verify the fix by testing on the live monkeytype.com site using browser DevTools.
You need to simulate the `group-hover` behavior by injecting JavaScript into the console that mirrored the exact fix removing `hover:opacity-100` from the overlay and simultaneously fading the base layer out and the overlay in on `mouseenter`. This confirmed that both layers transitioning together eliminates the overlap.
The script you can use:
```
document.querySelectorAll('.grid.items-center').forEach(card => {
  const layers = card.querySelectorAll(':scope > .col-start-1.row-start-1');
  if (layers.length === 2) {
    const mainLayer = layers[0];
    const overlay = layers[1];

    overlay.classList.remove('hover:opacity-100');
    
    card.addEventListener('mouseenter', () => {
      mainLayer.style.opacity = '0';
      mainLayer.style.transition = 'opacity 0.125s';
      overlay.style.opacity = '1';
      overlay.style.transition = 'opacity 0.125s';
    });
    card.addEventListener('mouseleave', () => {
      mainLayer.style.opacity = '1';
      overlay.style.opacity = '0';
    });
  }
});
```
This behaves identically to the `group`/`group-hover` Tailwind implementation in the fix.

AzureNightlock:
Ahh, I didn't think of that. That's actually a really smart way to verify it. Thanks for taking the time to explain it and sharing the script to simulate it!

byseif21:
Hi @AzureNightlock thanks for the PR, there is one small issue with the fix here that this will still fade even in empty stats (no PB) while this was not the case before , you have `transition-opacity group-hover:opacity-0` on the base content without guarding it, the fade shouldn't be applied when no PB e.g `item.pb !== undefined`.

Miodec:
> I couldn't fully test this on the real account page locally because I couldn't get the Firebase account setup working on my machine. I did check the behavior using the same PB tile structure and hover logic, so it's partially verified but not fully tested with the account flow. Sorry about the delay and inconvenience.

We recently added Storybook, which allows you to test components in pure isolation, with no requirements for firebase or a database - run `pnpm storybook` in root dir. Its pretty fresh so its undocumented still.

Ill wait for you to fix the issue @byseif21 pointed out and it will be good to go.

AzureNightlock:
Thanks for your patience @Miodec! and for catching that edge case @byseif21! I have pushed a fix so that the base content only fades when item.pb exists.

openvaibhav:
> Hi @AzureNightlock thanks for the PR, there is one small issue with the fix here that this will still fade even in empty stats (no PB) while this was not the case before , you have `transition-opacity group-hover:opacity-0` on the base content without guarding it, the fade shouldn't be applied when no PB e.g `item.pb !== undefined`.

You're right, good catch! I also missed this because I was testing with that JavaScript simulation script that only applied the hover effect when both layers were present in the DOM (layers.length === 2), so cards without a PB were naturally skipped by the script, masking the CSS issue. Also i got to know about the storybook so ill prefer that to test things from now on.

Miodec:
Thanks

PR #7599, opened 2026-03-09
refactor: allow AnimatedModal to be placed anywhere (@fehmer)

fehmer:
Use `Portal` to include `AnimatedModal` in the `#solidmodals` container no matter where the component is located.


fehmer:
not needed, works without.

PR #7624, opened 2026-03-11
feat: add lefthand funbox for left-hand-only typing practice (@gokul1108)

gokul1108:
Description

Add a new "lefthand" funbox that filters words to only those typeable with the left hand on a QWERTY layout (keys: `qwertasdfgzxcvb`). Useful for practicing weak left-hand typing.

Changes:
- `packages/schemas/src/configs.ts` — Added `"lefthand"` to `FunboxNameSchema`
- `packages/funbox/src/list.ts` — Added funbox metadata entry
- `frontend/src/ts/test/funbox/funbox-functions.ts` — Implemented `withWords` that filters language words to left-hand-only words

### DIFF
--- a/frontend/src/ts/test/funbox/funbox-functions.ts
+++ b/frontend/src/ts/test/funbox/funbox-functions.ts
@@ -564,6 +564,15 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
       else return "";
     },
   },
+  lefthand: {
+    async withWords(words?: string[]): Promise<Wordset> {
+      const LEFT_HAND_KEYS = new Set("qwertasdfgzxcvb");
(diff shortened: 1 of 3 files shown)


fehmer:
hi @gokul1108 , thank you for your contribution.

You can already use the custom mode to do this:
- select test type custom
- press `change` 
- go to `word filter`
- select preset `left hand`
- `apply` and `set`

gokul1108:
> hi @gokul1108 , thank you for your contribution.
> 
> You can already use the custom mode to do this:
> 
> * select test type custom
> * press `change`
> * go to `word filter`
> * select preset `left hand`
> * `apply` and `set`

Thanks for pointing that out!

PR #7684, opened 2026-03-19
fix(profile): prevent saving profile when no changes are made (@AzureNightlock)

### LINKED ISSUE #7683: Profile saves even when no changes are made

- [x] I have cleared my cache

Yes

No

Yes

No

### Account name

azurenightlock

### Account config

_No response_

### Current Behavior

When the user clicks the Edit Profile button and clicks Save without making any changes, the profile is still submitted and saved again. This triggers an unnecessary update request even though the profile data has not changed.

### Expected Behavior

If the user has not changed any profile fields, clicking Save should not submit the profile again. Instead, the app should prevent the unnecessary save and inform the user that there are no changes to save.

### Steps To Reproduce

1. Go to your account page (https://monkeytype.com/account).
2. Click on the edit profile button.
3. Scroll down until you reach the end of the popup.
4. Click on the save button without making any changes.

### Environment

- OS: Windows 11
- Browser: Chrome
- Browser Version: 146.0.7680.153 (Official Build) (64-bit)

### Anything else?

I would like to work on this.

AzureNightlock:
### Description
This PR: Fixes edit-profile.ts so the profile is not submitted again when the user clicks Save without making any changes.

To support that behavior, this PR:
* Introduces a profile state shape used to capture form values.
* Stores the original profile state when the modal is opened.
* Reads the current profile state at save time.
* Compares the original and current profile states before submitting.
* Prevents the save request when no changes are detected.
* Shows a "No changes to save" notification when the profile is unchanged.
* Reuses getProfileState() to avoid reading form inputs twice.
* Renames buildUpdatesFromInputs to buildUpdatesFromState for clarity, since updates are now built from the captured form state instead of reading inputs again.

#### Additional small fixes:
* Adds a guard in initializeCharacterCounters to avoid attaching duplicate listeners.
* Handles the case where inventory exists but badges is undefined.
* Corrects a typo in a comment.
* Changes the checkbox cursor to pointer.

### Checks

Closes #7683

### DIFF
--- a/frontend/src/ts/modals/edit-profile.ts
+++ b/frontend/src/ts/modals/edit-profile.ts
@@ -23,6 +23,8 @@ export function show(): void {
   void modal.show({
     beforeAnimation: async () => {
       hydrateInputs();
+      originalState = getProfileState();
+      updateSaveButtonState();
       initializeCharacterCounters();
     },
   });
@@ -31,6 +33,10 @@ export function show(): void {
 function hide(): void {
   void modal.hide();
 }
+
+const saveButton = qsr<HTMLButtonElement>(
+  "#editProfileModal .edit-profile-submit",
+);
 
 const bioInput = qsr<HTMLTextAreaElement>("#editProfileModal .bio");
 const keyboardInput = qsr<HTMLTextAreaElement>("#editProfileModal .keyboard");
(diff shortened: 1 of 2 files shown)


AzureNightlock:
Testing note: I created an edit-profile-spec test, and it passed locally. I wasn't able to fully verify this change visually because there isn't a Story for this section, and I wasn't able to set up the backend properly. So the validation for this change was based on the spec test and the updated no-change save logic.

Also for future fixes in account-related areas like this, is there a recommended way to test them visually/locally end-to-end? I've run into this issue a few times and would love to be able to test these fixes better.

frontend/src/ts/modals/edit-profile.ts
```diff
@@ -176,7 +220,7 @@ async function updateProfile(): Promise<void> {
   }
 
   snapshot.details = response.body.data ?? updates;
-  snapshot.inventory?.badges.forEach((badge) => {
+  snapshot.inventory?.badges?.forEach((badge) => {
```
Leonabcd123: Why is this necessary? I don't think `badges` can be undefined/null.

fehmer:
hi @AzureNightlock , thank you for your contribution.

The best call would be to get the backend working on your machine. Can you join discord so we can trouble shoot your problem?

Let us know what other changes you have planned that are account-related. We are currently migrating the frontend to use solid-js and I think we can get this fix "for free" with the new stack.

frontend/src/ts/modals/edit-profile.ts
AzureNightlock: Thanks for pointing that out! I added it defensively but the extra optional chaining on badges was unnecessary. Sorry for that oversight and I appreciate the careful review!

AzureNightlock:
Sure I would love to! I have already joined the Discord already and I do have a few other account-related bugs in mind that I was thinking of looking into as well.

AzureNightlock:
I have set up my backend and tested this with an actual account and it works as expected.  
The new changes made were:
* The GitHub/Twitter length checks were removed since the Zod schemas already have this covered. 
* The save button is disabled if there are no changes in input or if any validation errors exist. 
* The validation indicator (tick mark) is hidden when a field value matches the original saved value.

PR #7695, opened 2026-03-21
refactor: solid custom text and associated modals (@miodec)

Miodec:


frontend/src/ts/components/modals/CustomTextModal.tsx
fehmer, line 1: can we use ts/form?

PR #7697, opened 2026-03-21
refactor(frontend): extract full-page loading screen controller (@hunterpaulson)

hunterpaulson:
### Description
- extract the full-page loading screen orchestration from `page-controller` into a dedicated `loading-screen` controller
- update `page-controller` to use the extracted helper

#### Why
The full-page loading screen logic was previously embedded inside `page-controller`, which made it harder to reuse outside route transitions.
This refactor keeps the current behavior the same while moving the loading-screen orchestration into a dedicated controller so future non-routing flows (e.g. funboxes) can reuse the same loading UI without duplicating controller internals.

#### Validation
- `pnpm ts-check`
- `pnpm lint-fast "src/ts/controllers/loading-screen.ts" "src/ts/controllers/page-controller.ts"`
- manually verified the extracted helper still shows the full-page loading screen with both bar and spinner modes

### DIFF
--- a/frontend/src/ts/controllers/loading-screen.ts
+++ b/frontend/src/ts/controllers/loading-screen.ts
@@ -1,0 +1,122 @@
+import * as Misc from "../utils/misc";
+import * as PageLoading from "../pages/loading";
+import type Page from "../pages/page";
+import type { LoadingOptions } from "../pages/page";
+
+// global abort controller for keyframe promises
+let keyframeAbortController: AbortController | null = null;
+
+export async function showBlockingLoadingScreen({
+  loadingOptions,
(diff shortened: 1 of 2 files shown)


fehmer:
hi @hunterpaulson, thank you for your contribution.

Do you already have a use-case in mind?

hunterpaulson:
@fehmer yes, I’m working on a funbox that needs to do a second or two of initialization the first time it is selected, and I wanted to reuse the existing full-page progress bar for that flow instead of the [LoaderBar](https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/ts/states/loader-bar.ts) that doesn't currently support  progress.

PR #7698, opened 2026-03-22
feat: add "Previous Test" functionality with button and state management (@abhaychandra123)

abhaychandra123:
### Title
feat(frontend): add previous quote navigation + align test action buttons (@abhaychandra123)

## Summary
This PR adds session-based previous-quote navigation in quote mode and refactors test action button layout so Previous and Restart are consistently sized, aligned, and centered.

## What changed
- Added a new Previous Test button next to Restart and wrapped both in a shared flex container in frontend/src/html/pages/test.html.
- Added shared button styling and centered flex layout in frontend/src/styles/test.scss.
- Updated coarse-pointer media query so both buttons keep consistent behavior on mobile/touch in frontend/src/styles/media-queries-blue.scss.
- Added a command line command for previous quote navigation in frontend/src/ts/commandline/lists/navigation.ts.
- Added quote history reset on mode/language transitions in frontend/src/ts/config/setters.ts.
- Added click handler for Previous Test button in frontend/src/ts/event-handlers/test.ts.
- Extended restart/init flow to support loading previous quote in frontend/src/ts/test/test-logic.ts.
- Added quote history state management helpers in frontend/src/ts/test/test-state.ts.
- Added previous button visibility updates on restart and config changes in frontend/src/ts/test/test-ui.ts.
- Extended quote generation to support explicit previous-quote loading and safe history updates in frontend/src/ts/test/words-generator.ts.

## Behavior details
- Previous quote navigation is available only in quote mode and only when history exists.
- History is session-scoped, capped at 50 entries, and cursor-based.
- History resets when leaving quote mode or changing language.
- Restart button styling is preserved; Previous now shares the same visual style and dimensions.

### Checks
- [x] Not adding quotes
- [x] Not adding a language
- [x] Not adding a theme
- [x] Not adding a layout
- [x] Not adding a font
- [x ] PR title follows Conventional Commits and includes GitHub username
- [ ] Related issues linked below (if applicable)

### DIFF
--- a/frontend/__tests__/root/config.spec.ts
+++ b/frontend/__tests__/root/config.spec.ts
@@ -56,9 +56,16 @@ describe("Config", () => {
       vi.useFakeTimers();
       mocks.forEach((it) => it.mockClear());
 
-      vi.mock("../../src/ts/test/test-state", () => ({
-        isActive: true,
-      }));
+      vi.mock("../../src/ts/test/test-state", async (importOriginal) => {
+        const actual =
+          await importOriginal<typeof import("../../src/ts/test/test-state")>();
+
+        return {
+          ...actual,
+          isActive: true,
+          resetQuoteHistory: vi.fn(),
+        };
+      });
 
       isConfigValueValidMock.mockReturnValue(true);
       canSetConfigWithCurrentFunboxesMock.mockReturnValue(true);
--- a/frontend/src/html/pages/test.html
+++ b/frontend/src/html/pages/test.html
@@ -162,14 +162,25 @@
 
     <div id="keymap" class="hidden"></div>
 
-    <button
-      id="restartTestButton"
-      aria-label="Restart Test"
-      data-balloon-pos="down"
-      class="text"
-    >
-      <i class="fas fa-fw fa-redo-alt"></i>
-    </button>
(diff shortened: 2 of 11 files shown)


abhaychandra123:
btw what my Motivation is,

In quote mode, restarting a test always fetches a new quote. In practice, users often restart multiple times to find a quote they like or are in the mood to type. However, this leads to a small but frustrating UX issue

Miodec:
Im not sure i see the point here? You can already use the repeat test button to type the same quote again.

abhaychandra123:
> Im not sure i see the point here? You can already use the repeat test button to type the same quote again.

You're right that the repeat test button allows retyping the same quote.

This feature is solving a slightly different problem:

When browsing quotes, it's common to restart multiple times to find a quote that feels interesting or comfortable to type. In that process, it's easy to accidentally skip over a quote you actually wanted to type.

The current behavior only allows:
- repeat → same quote again
- restart → completely new quote

There is no way to go back to a previously shown quote after a restart.

This change introduces a lightweight history (only for quote mode) so users can:
- recover a quote they accidentally skipped
- navigate back one step without repeatedly restarting and hoping it appears again

So it complements the repeat functionality rather than replacing it.

PR #7702, opened 2026-03-22
impr: enable spaces in preset names to unify naming logic (@byseif21)

byseif21:
* allow spaces in preset like in tag names (displayed with spaces, stored with underscores)


packages/schemas/src/presets.ts
```diff
@@ -8,8 +8,13 @@ import {
 
 export const PresetNameSchema = z
   .string()
-  .regex(/^[0-9a-zA-Z_-]+$/)
-  .max(16);
+  .transform((val) => val.trim().replace(/\s+/g, "_"))
```
fehmer: i am not sure we should use this approach. It will change the validation on the backend as well. With this change you can post a preset with `name:"name with spaces"` but get back the preset with `name: "name_with_spaces"`.

PR #7703, opened 2026-03-22
impr(input-validation): clean preset name input when adding (@byseif21)

byseif21:
`ValidatedHtmlInputElement` would reset its validation status but keep the old text in the input field

1- go to presets
2- click edit from one of the existing ones
3- go back and click add preset
4- notice the edit one's name is there


Miodec (approved):

davlat-26:
x

PR #7706, opened 2026-03-23
docs: guidelines for pr descriptions (@fehmer)

fehmer:


.github/pull_request_template.md
```diff
@@ -1,6 +1,16 @@
 ### Description
 
-<!-- Please describe the change(s) made in your PR -->
+<!-- 
+  Please describe the change(s) made in your PR:
+  - explain the problem being solved
+  - for bug fixes, include steps to reproduce the issue
```
Leonabcd123: Isn't necessary when fixing a bug that has its own issue, might be worth making this more specific:

`For bug fixes without an open issue, include steps to reproduce the issue`

.github/pull_request_template.md
```diff
@@ -1,6 +1,16 @@
...
+  - summarize the approach taken
+  
+  Use your own words. Do not rely on AI-generated descriptions. 
+  They do not demonstrate your understanding of the problem or the solution.
+  Writing the description yourself helps you verify the scope of your work and 
+  allow maintainers to better understand your intent, reasoning and level of insight.
```
Leonabcd123: Should be `allows` instead of `allow`, or even better:

`helps us better understand ...`

PR #7716, opened 2026-03-24
fix: vendor-specific styles overriding standard properties (@nadalaba)

nadalaba:
1. switch tape off
2. switch to tape letter
3. `#words` height will keep its old height without fitting content
tape

This happens because `height: -moz-fit-content` overrode `height: fit-content`, and vite8's cssMinify ([lightningCSS](https://lightningcss.dev/)) removes overridden properties.

### DIFF
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -192,8 +192,8 @@
 }
 
 #words {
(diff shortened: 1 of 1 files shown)


nadalaba:
fixed in #7718

PR #7717, opened 2026-03-24
refactor(quotes): use tanstack (@fehmer)

fehmer:
wip


frontend/src/ts/collections/quote-ratings.ts
```diff
@@ -0,0 +1,134 @@
...
+            throw new Error(
+              "Cannot submit quote rating: " + response.body.message,
+            );
+          }
+        }
+      });
+    },
+
+    queryClient,
+    getKey: (it) => it._id,
```
fehmer, line 114: use lang+quoteId instead

PR #7787, opened 2026-04-08
refactor: add direction property to test words (@nadalaba)

nadalaba:


frontend/__tests__/utils/strings.spec.ts
```diff
@@ -217,251 +217,193 @@ describe("string utils", () => {
...
+        "١٢٣",
+        "ltr",
+        "Arabic-Indic digits with no strong typed chars should fallback to ltr",
+      ],
+      [
+        "rtl",
+        "١٢٣",
+        "rtl",
+        "Arabic-Indic digits with no strong typed chars should fallback to rtl",
+      ],
```
byseif21, line 322: why numbers should be `neutral`? is there a case where digits should follow the fallback direction here? shouldn't numbers probably always be LTR even if the overall language is RTL?

byseif21: how that goes in RTL tests while numbers there ? eg `58008` funbox or polyglot+numbers on as main test RTL? i think bad UX

frontend/src/ts/utils/direction-regex.ts
byseif21, line 1: if there is no case i think we should include digits in `STRONG_LTR_TYPE`

byseif21: or idk if that is bad since they are not really strong ltr maybe we could have like `numbers_type` or something

frontend/src/ts/test/words-generator.ts
```diff
@@ -711,16 +720,9 @@ export async function generateWords(
...
-    (Config.mode === "quote" &&
-      (quote as QuoteWithTextSplit).textSplit.some((w) => w.includes("\t")));
-  ret.hasNewline =
-    ret.words.some((w) => w.includes("\n")) ||
-    currentWordset.words.some((w) => w.includes("\n")) ||
-    (Config.mode === "quote" &&
-      (quote as QuoteWithTextSplit).textSplit.some((w) => w.includes("\n")));
+  ret.hasTab = currentWordset.words.some((w) => w.includes("\t"));
+  ret.hasNewline = currentWordset.words.some((w) => w.includes("\n"));
+  ret.hasNumbers = currentWordset.words.some((w) => /\d/g.test(w));
```
Miodec: Regression. This now wont detect newlines / tabs in quotes OR if a funbox adds them.

frontend/src/ts/test/test-logic.ts
```diff
@@ -512,19 +508,22 @@ async function init(): Promise<boolean> {
...
 
+  let wordsHaveNumbers = false;
   let wordsHaveTab = false;
   let wordsHaveNewline = false;
-  let allRightToLeft: boolean | undefined = undefined;
   let allLigatures: boolean | undefined = undefined;
-  let generatedWords: string[] = [];
+  let generatedWords: TestWords.Word[] = [];
+  let generatedWordsText: string[] = [];
   let generatedSectionIndexes: number[] = [];
```
Miodec: This is now not populated, always logs `[]`. Probably just remove. Same with `generatedWordsText`, just remove and map over `generatedWords` in the log

frontend/src/ts/test/test-ui.ts
```diff
@@ -374,12 +371,16 @@ async function updateHintsPosition(): Promise<void> {
   }
 }
 
-function buildWordHTML(word: string, wordIndex: number): string {
+type RequireOnly<T, K extends keyof T> = Required<Pick<T, K>> &
+  Partial<Omit<T, K>>;
+type WordTextWithDirection = RequireOnly<TestWords.Word, "text" | "direction">;
```
Miodec: Isnt this the same as `Pick`?

Miodec (changes requested):

frontend/src/ts/test/test-ui.ts
nadalaba: Pick removes sectionIndex from the type. RequireOnly makes it optional.
This is needed because sometimes we pass the whole TestWords.word. Others we only pass {test, direction}

Miodec: But it seems like we aren't using `sectionIndex` anywhere, so why not define `TextWithDirection` as just `{text:string,direction:Direction}`?

frontend/src/ts/test/words-generator.ts
nadalaba: we don't need to check `TestWords.currentQuote.textSplit`. It is the same as `currentWordset.words`

PR #7788, opened 2026-04-08
chore(languages): rename `ligatures` to `joiningScript` (@nadalaba)

nadalaba:


Miodec (commented):
Claude says:

```
  "cursiveScript" isn't quite right either — Arabic,
  Persian, Urdu etc. are cursive scripts. But Hindi
  (Devanagari), Malayalam, Tamil, Telugu, Kannada, Bangla,
  Gujarati etc. are not cursive scripts — they use connecting
  marks (matras/vowel signs that reshape) but the base
  consonants don't join. The old name "ligatures" was wrong,
  but "cursiveScript" is also inaccurate for about half the
  languages that have cursiveScript: true. Something like
  joiningScript or contextualShaping might be more precise.
  This is a naming nit though — up to you whether it matters.
```

What do you think?

nadalaba:
`joiningScript` is ok for me. contextualShaping might be more accurate but not as clear

Miodec:
> `joiningScript` is ok for me. contextualShaping might be more accurate but not as clear

Lets do `joiningScript`.

Miodec:
Thanks

PR #7793, opened 2026-04-08
feat(typed effect): add fall typed effect (@d1rshan)

d1rshan:
https://github.com/user-attachments/assets/c029656a-8553-4b0e-b882-c78a71f4f7d7

This PR adds a new typed effect called `fall`.

When a correctly submitted word becomes typed, it creates a temporary animated clone that falls off the screen while the original word is hidden. Incorrect submitted words stay in place so error feedback remains visible.

This PR also introduces a dedicated `typed-effects` module for typed effects that need JavaScript behavior, so effect-specific logic does not have to live directly in `test-ui`.

also addresses #7847, makes sure `fall` works for reduced motion.


frontend/src/ts/test/typed-effects.ts
```diff
@@ -0,0 +1,53 @@
...
+  clone.style.zIndex = "1000";
+  clone.style.setProperty("--fall-rotation", `${randomRotation}deg`);
+  clone.style.setProperty("--fall-x", `${randomX}px`);
+
+  document.body.appendChild(clone);
+  word.setStyle({ opacity: "0" });
+
+  clone.addEventListener("animationend", () => {
+    clone.remove();
+  });
```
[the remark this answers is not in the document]
d1rshan: fixed, updated tumble effect to a early return on under `prefers-reduced-motion` to avoid any weird behaviour and normal mode clean up now uses  `{ once : true}` and a timeout fallback.

frontend/src/ts/test/test-ui.ts
```diff
@@ -147,6 +148,7 @@ export function updateActiveElement(
       if (previousActiveWord !== null) {
         if (direction === "forward") {
           previousActiveWord.addClass("typed");
+          TypedEffects.onWordTyped(previousActiveWord);
           Ligatures.set(previousActiveWord, true);
```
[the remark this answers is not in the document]
d1rshan: fixed, changing typedEffect mid-test now calls `TypedEffects.clear()` which removes pending clones and restores word opacity.

frontend/src/ts/test/test-ui.ts
```diff
@@ -2081,6 +2086,9 @@ configEvent.subscribe(({ key, newValue }) => {
       "tapeMargin",
     ].includes(key)
   ) {
+    if (key === "typedEffect") {
+      TypedEffects.clear();
+    }
```
nadalaba: What do you think about clearing only if the `newValue !== "tumble"` ?

frontend/src/styles/test.scss
```diff
@@ -533,6 +533,18 @@
...
   }
 
+  &.typed-effect-tumble {
+    .word.typed:not(.error) {
+      opacity: 0;
+    }
+
+    @media (prefers-reduced-motion) {
+      .word.typed:not(.error) {
+        opacity: 1;
```
nadalaba: why disable hiding words in prefers-reduced-motion? I think disabling animations is enough

frontend/src/ts/test/typed-effects.ts
```diff
@@ -0,0 +1,59 @@
...
+  clone.style.fontFamily = computedStyle.fontFamily;
+  clone.style.color = computedStyle.color;
+  clone.style.margin = "0";
+  clone.style.pointerEvents = "none";
+  clone.style.zIndex = "1000";
+  clone.style.setProperty("--fall-rotation", `${randomRotation}deg`);
+  clone.style.setProperty("--fall-x", `${randomX}px`);
+
+  document.body.appendChild(clone);
+  word.setStyle({ opacity: "0" });
```
nadalaba: why is setting inline opacity needed? Doesn't the change in css already do the job?

frontend/src/ts/test/test-ui.ts
d1rshan: addressed in [b01e7cd](https://github.com/monkeytypegame/monkeytype/pull/7793/commits/b01e7cd98af3a949923264f33495b2af746504f3)

frontend/src/styles/test.scss
d1rshan: now for prefers-reduced-motion, words are hidden but falling words animation is disabled. addressed in  [b01e7cd](https://github.com/monkeytypegame/monkeytype/pull/7793/commits/b01e7cd98af3a949923264f33495b2af746504f3)

frontend/src/ts/test/typed-effects.ts
d1rshan: you're right, thanks for pointing that out. removed setting opacity inline [b01e7cd](https://github.com/monkeytypegame/monkeytype/pull/7793/commits/b01e7cd98af3a949923264f33495b2af746504f3)

frontend/src/ts/test/typed-effects.ts
```diff
@@ -0,0 +1,55 @@
...
+  clone.style.top = `${rect.top}px`;
+  clone.style.left = `${rect.left}px`;
+  clone.style.width = `${rect.width}px`;
+  clone.style.height = `${rect.height}px`;
+  clone.style.fontSize = computedStyle.fontSize;
+  clone.style.fontFamily = computedStyle.fontFamily;
+  clone.style.color = computedStyle.color;
+  clone.style.margin = "0";
+  clone.style.pointerEvents = "none";
+  clone.style.zIndex = "1000";
```
nadalaba: What do you think about keeping inline styles only when dynamic styling is needed, and using css classes for the rest? it'd also minimize duplication

d1rshan: done [bb1a786](https://github.com/monkeytypegame/monkeytype/pull/7793/commits/bb1a786144aac3440c3a6d37399ed542d99f5585)

nadalaba (approved):

Miodec:
Tumble makes me think the words would spin. I think rename to 'fall'?

d1rshan:
hey mio, i updated it to 'fall'.

Miodec:
Maybe instead of having to define the animation via css we could use animejs for this? Also, the color of the fall clone is incorrect when using colorful mode or flip test colors.

d1rshan:
hey @Miodec, i am not very familiar with animejs and i did try doing it with ai but the effect feels bit weird with animejs and since rest of the typed effects use css i think it's just cleaner to keep this effect as css too ? if u still insist, i can move the effect to animejs.

and i fixed the fall-clone color in modes affecting word colors.

fehmer:
Well, the other effects use just CSS and no JavaScript. It is already a special case.

d1rshan:
> Well, the other effects use just CSS and no JavaScript. It is already a special case.

okay, will do it.

d1rshan:
updated to animejs, works almost the same now. works in flip colors and colorful mode too.

d1rshan:
@Miodec can u pls take a look at this, i updated it to animejs and fixed it to work in flip colors and colorful mode

Miodec:
Throws a lot of errors when you use freedom mode to go to a correctly typed word.

Miodec:
Mark as ready when you fix it.

d1rshan:
sry didn't knew about freedom mode, will fix and update

d1rshan:
@Miodec  This PR is ready for review when you have time. Thanks!

PR #7798, opened 2026-04-09
impr(theme): display custom theme names in indicator (@Sharann-del)

Sharann-del:
Improves the theme indicator by displaying the actual custom theme name instead of the generic "custom".

- Matches active custom theme using color comparison
- Displays as: "themeName (custom)"
- Falls back to "custom" if no match is found

Notes:
- Config does not store custom theme IDs, so matching is done via color arrays
- Handles edge cases such as unsaved themes and missing DB snapshot
- Does not affect preview or random theme behavior


Sharann-del:
Hi! Just checking again if this PR looks good or needs any changes. Happy to update if needed.

Miodec:
Sorry it took me a while, thanks for the pr.

PR #7799, opened 2026-04-09
fix: double lineJump in zen mode (@nadalaba)

nadalaba:
doubleJump


frontend/src/ts/test/test-ui.ts
```diff
@@ -902,7 +903,14 @@ export async function updateWordLetters({
         if (!Config.showAllLines) {
           const wordTopAfterUpdate = wordAtIndex.getOffsetTop();
           if (wordTopAfterUpdate > activeWordTop) {
```
Miodec, line 905: Isnt it enough to just check if line is transitioning in this check?

nadalaba: Theoretically it is possible to have a valid lineJump while the previous one didn't finish (lineTransition === true).
We only want to prevent invalid ones.
I can provide an example of the valid case after merging #7832

PR #7803, opened 2026-04-10
fix(funbox): apply space balls perspective only to test words (@d1rshan)

d1rshan:
This PR restricts the Space Balls 3D perspective effect only to the words ui to prevent layout issues with other test elements.

Current Behavior:
image

Now:
image

ik this changes the original intention for this effect, so understandable if this is not what you guys want but i have tried multiple different options and i think this is probably the cleanest fix for this.

### DIFF
--- a/frontend/static/funbox/space_balls.css
+++ b/frontend/static/funbox/space_balls.css
@@ -16,11 +16,11 @@ body {
   background-position: center;
 }
 
-main {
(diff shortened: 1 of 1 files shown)


Miodec:
I thiiink i would prefer the whole test page having the perspective, not just words. Thanks though.

PR #7805, opened 2026-04-10
fix: resolve sort comparator and code quality issues (@ennajari)

ennajari:
## Summary

- Use explicit `!== undefined` check instead of truthy check for cached `Promise` in memoize utility (`json-data.ts`)

## Test plan

- [ ] Memoized async functions cache correctly


backend/src/dal/connections.ts
```diff
@@ -331,7 +331,7 @@ export async function aggregateWithAcceptedConnections<T>(
 
 function getKey(initiatorUid: string, receiverUid: string): string {
   const ids = [initiatorUid, receiverUid];
-  ids.sort();
+  ids.sort((a, b) => a.localeCompare(b));
```
fehmer: Not needed for backend

fehmer:
We don't use sonar, please remove sonar instructions and .ignore

frontend/src/ts/config/setters.ts
```diff
@@ -186,7 +186,7 @@ export function toggleFunbox(funbox: FunboxName, nosave?: boolean): boolean {
     newConfig = newConfig.filter((it) => it !== funbox);
   } else {
     newConfig.push(funbox);
-    newConfig.sort();
+    newConfig.sort((a, b) => a.localeCompare(b));
```
fehmer: I think we want consistent sorting here and not depend on the users locale.

fehmer:
Hi @ennajari , thanks for the changes. The pr summary doesn't match the or content anymore

backend/src/dal/connections.ts
ennajari: Reverted —  is back to plain .

frontend/src/ts/config/setters.ts
ennajari: Changed to a locale-independent comparator: `(a, b) => (a < b ? -1 : a > b ? 1 : 0)`. Same for Sidebar.tsx and themes.ts.

fehmer: Could you describe the difference between 
```javascript
Object.keys(themes).sort()
```
and 
```javascript
 Object.keys(themes).sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))
```

ennajari: Honestly, there is no functional difference. For ASCII strings like theme/funbox names, both perform Unicode code-point comparison and produce identical results. The explicit comparator only satisfies the static analysis rule that requires a compare function to be provided, but the behavior is the same as plain `.sort()`.

If you'd prefer, I can drop these three sort changes entirely — the two remaining fixes (`reduce` initial value and the `!== undefined` check) still stand on their own.

fehmer: If both do the same let's keep the simple `.sort`

ennajari: Reverted all three sort changes back to plain `.sort()`. The PR now only contains the `reduce` initial value fix and the `!== undefined` cache check.

ennajari (commented):
Everything alright

packages/util/src/numbers.ts
```diff
@@ -46,7 +46,8 @@ export function stdDev(array: number[]): number {
     const n = array.length;
     const meanValue = mean(array);
     return Math.sqrt(
-      array.map((x) => Math.pow(x - meanValue, 2)).reduce((a, b) => a + b) / n,
+      array.map((x) => Math.pow(x - meanValue, 2)).reduce((a, b) => a + b, 0) /
```
nadalaba: I don't think this is needed. `reduce((a, b) => a + b)` is a known summing pattern

ennajari: the surrounding `try/catch` already returns `0` if the array is empty, so the initial value has no effect. I'll revert this.

Miodec:
Thanks

PR #7806, opened 2026-04-10
fix(tests): handle unknown .jsx extension for TanStack solid dependencies (@ennajari)

### LINKED ISSUE #7801: vitest crash: Unknown file extension ".jsx" for TanStack dependencies
### Current Behavior

`vitest` is currently crashing when running any unit test that pulls in a dependency from the `@tanstack/solid-*` ecosystem (specifically `@tanstack/solid-store`).

existing tests work fine, but if a test imports anything that uses these newer `TanStack` libraries, it throws this error: `TypeError: Unknown file extension` ".jsx" for `.../node_modules/@tanstack/solid-store/dist/source/index.jsx`

ennajari:
Fixes #7801

`@tanstack/solid-*` packages ship `.jsx` source files that Node.js cannot load without a JSX transform.

**Fix:** Add `ssr.noExternal: ["@solidjs/meta", /@tanstack\/solid-.*/]` to all three test projects (unit, jsdom, jsx). This forces Vite to bundle and transform them through `vite-plugin-solid`.

**Test plan**
- pnpm vitest run completes without "Unknown file extension" errors
- Tests importing `@tanstack/solid-*` packages run correctly


byseif21:
hi @ennajari , did you test that really? cuz that doesn't seem like the full fix

frontend/vitest.config.ts
```diff
@@ -61,7 +69,15 @@ export default defineConfig({
...
-          include: ["@monkeytype/funbox"],
+          include: [
+            "@monkeytype/funbox",
+            "@tanstack/solid-store",
+            "@tanstack/solid-form",
+            "@tanstack/solid-query",
+            "@tanstack/solid-table",
+            "@tanstack/solid-hotkeys",
+            "@tanstack/solid-db",
+          ],
```
byseif21: why the optimizer for? `noExternal` only seems to fix it

frontend/vitest.config.ts
```diff
@@ -36,7 +36,15 @@ export const projects: UserWorkspaceConfig[] = [
...
-      noExternal: ["@solidjs/meta"],
+      noExternal: [
+        "@solidjs/meta",
+        "@tanstack/solid-store",
+        "@tanstack/solid-form",
+        "@tanstack/solid-query",
+        "@tanstack/solid-table",
+        "@tanstack/solid-hotkeys",
+        "@tanstack/solid-db",
+      ],
```
byseif21: could be `noExternal: ["@solidjs/meta", /@tanstack\/solid-.*/]` , also i think needed for the two others test i guess

byseif21:
btw @fehmer has assigned it to himself, so he may want to look at it himself,  i think you should have asked before the PR

fehmer:
i don't mind somebody else fixing it, I haven't started working on it yet.

@byseif21 could you provide the test-case to reproduce the problem?

frontend/vitest.config.ts
ennajari: Removed optimizer entries dropped, `noExternal` is the correct fix.

byseif21:
> @byseif21 could you provide the test-case to reproduce the problem?

sure, here
```ts
import { test, expect } from "vitest";
import { createHotkey } from "@tanstack/solid-hotkeys";

test("triggers jsx crash", () => {
  expect(createHotkey).toBeDefined();
});
```
try as a new test file (e.g `repro.spec.ts` in `frontend/__tests__/`), running it should fail in master

fehmer:
thanks @byseif21 ,

shouldn't the test be named `repo.spec.tsx`? Do we need to add the fix to all three kinds of test or only the `jsx`test?

Just adding the regex to the `jsx`tests works for me:

```
  ssr: {
      noExternal: ["@solidjs/meta", /@tanstack\/solid-.*/],
    },
```

byseif21:
> shouldn't the test be named `repo.spec.tsx`? Do we need to add the fix to all three kinds of test or only the `jsx`test?

hmm the thing is, wouldn't naming tests as .tsx be a bit inconsistent if they don't actually have UI components? I thought it was better to keep logic tests as .ts (like the ones in `frontend/__tests__/`) and only use .tsx for components?

tests like for the hotkeys one when we add it, it will be only logic test and if it'll be as ts as i think, importing e.g `@tanstack/solid-hotkeys` or anything that use it, makes it fail

ennajari:
Agreed with @byseif21. The repro imports from `@tanstack/solid-hotkeys` in a `.spec.ts` file, which matches the `unit` project (`__tests__/**/*.spec.ts`) — not the `jsx` one (`.spec.tsx`). Applying `noExternal` only to the `jsx` project would leave this failure path open.

Current PR already applies the regex to all three projects via a shared constant, so the repro should pass as-is. Happy to narrow it down if you prefer a different split.

fehmer (approved):

Miodec:
Thanks

PR #7808, opened 2026-04-10
fix(test-timer): slow timer notification reappearing after new test (@J-Karthikeyan)

J-Karthikeyan:
### Description

When a test fails due to the slow timer, the error notification ("Stopping the test due to bad performance...") has no auto dismiss timeout so it stays in the notification store indefinitely so it reappears once the next test is completed.

When we start a new test, the notification gets visually hidden by focus mode (non important notifications get a `hidden` class when getFocus() is true). When the test ends, focus mode turns off and the notification reappears making it look like it fired again when it didn't.

### Steps to Reproduce
  1. Start a test and let the slow timer trigger the notification by just typing one or two letter and leaving it.
  2. Do NOT dismiss the notification
  3. Start and complete a new test without a page refresh
  4. The notification reappears at the end of the new test

### Fix
Fix: track the error notification's ID when the slow timer fires, and remove it from the store when a new test starts via TestTimer.start().

### DIFF
--- a/frontend/src/ts/states/notifications.ts
+++ b/frontend/src/ts/states/notifications.ts
@@ -115,7 +115,7 @@ export function addNotificationWithLevel(
   message: string,
   level: NotificationLevel,
   options: AddNotificationOptions = {},
-): void {
+): number {
   let details = options.details;
 
   if (options.response !== undefined) {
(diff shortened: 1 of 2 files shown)


Miodec (approved):
Thanks

PR #7818, opened 2026-04-12
feat(leaderboard): added page indicator (@Repr-dev)

Repr-dev:
### Description

At the moment, it is impossible to tell which page of [monkeytype.com/leaderboards](https://www.monkeytype.com/leaderboards) you are on just from the content of the webpage. One must look at the URL parameters (*/leaderboards?...&page=X*) in order to get this information. So, I added a numerical indicator next to the page control buttons.

All this code does is add a **span** element in the same container as the page control buttons. The span is placed to the left of the buttons and is given a lot of the same Tailwind classes. The page number in the span comes from the data that is already being passed down to the buttons.

**Old:**
image

**New** (replication of my approach, made with Chrome Inspector)**:**
image

### Closes
Discussion #7815. My first open source contribution

### DIFF
--- a/frontend/src/ts/components/pages/leaderboard/Navigation.tsx
+++ b/frontend/src/ts/components/pages/leaderboard/Navigation.tsx
@@ -5,6 +5,8 @@ import { showSimpleModal } from "../../../states/simple-modal";
 import { cn } from "../../../utils/cn";
 import { Button } from "../../common/Button";
 import { LoadingCircle } from "../../common/LoadingCircle";
+import { PageIndicator } from "./PageIndicator";
+
 export function Navigation(props: {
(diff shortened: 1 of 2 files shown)


Miodec:
Maybe something like this instead? 
image

Miodec:
Or we could go with something more classic 
image

Repr-dev:
@Miodec [5c9f425](https://github.com/monkeytypegame/monkeytype/pull/7818/changes/5c9f4253b275cc9b77890790ae64f7bd49538483), should look like this:
image

I thought I'd go with your first proposition to preserve the clear user page and go-to-page buttons.

Miodec:
Thanks

PR #7825, opened 2026-04-14
refactor: solid presets (@miodec)

Miodec:


frontend/src/ts/collections/presets.ts
```diff
@@ -0,0 +1,203 @@
...
+  return {
+    ...preset,
+    display: preset.name.replaceAll("_", " "),
+  };
+}
+
+// oxlint-disable-next-line typescript/explicit-function-return-type
+export function usePresetsLiveQuery() {
+  return useLiveQuery((q) => {
+    return q.from({ preset: presetsCollection }).select((p) => ({ ...p }));
```
fehmer: lets order by name here and in the `getPresets` function and remove from the  `fillPresetsCollection`

frontend/src/ts/collections/presets.ts
```diff
@@ -0,0 +1,203 @@
...
+      }
+      presetsCollection.utils.writeDelete(presetId);
+    },
+  }),
+};
+
+// --- Public API ---
+
+// todo: this might not be reactive
+export function getPresets(): PresetItem[] {
```
fehmer: ig `getPresets` and `getPreset` is for legacy/non-reactive code. Can we mark it somehow? maybe suffix with `Legacy` or `nonReactive` or we handle it like `__testing` and export a `__nonReactive` and add a lint rule to not use `__nonReactive` in `tsx` files?

frontend/src/ts/collections/presets.ts
```diff
@@ -0,0 +1,203 @@
...
+
+function toPresetItem(preset: Preset): PresetItem {
+  return {
+    ...preset,
+    display: preset.name.replaceAll("_", " "),
+  };
+}
+
+// oxlint-disable-next-line typescript/explicit-function-return-type
+export function usePresetsLiveQuery() {
```
fehmer, line 21: should we rename to ConfigPresets for more clarity? We have FilterPresets as well

frontend/src/ts/collections/presets.ts
```diff
@@ -0,0 +1,208 @@
...
+export type PresetItem = Preset & { display: string };
+
+const queryKeys = {
+  root: () => [...baseKey("presets", { isUserSpecific: true })],
+};
+
+function toPresetItem(preset: Preset): PresetItem {
+  return {
+    ...preset,
+    display: preset.name.replaceAll("_", " "),
```
fehmer: should we do the same as with tags? update the name to be without underscores, only convert when talking to the api?

frontend/src/ts/collections/presets.ts
```diff
@@ -0,0 +1,208 @@
...
+  deletePreset: {
+    presetId: string;
+  };
+};
+
+const actions = {
+  addPreset: createOptimisticAction<ActionType["addPreset"]>({
+    onMutate: ({ name, config, settingGroups }) => {
+      presetsCollection.insert({
+        _id: "temp-" + Date.now(),
```
fehmer: use tempId helper

frontend/src/ts/collections/presets.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+function getPreset(id: string): PresetItem | undefined {
+  return presetsCollection.get(id);
+}
+
+export function fillPresetsCollection(presets: Preset[]): void {
+  const presetItems = presets.map(toPresetItem);
+
+  presetsCollection.utils.writeBatch(() => {
```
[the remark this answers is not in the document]
fehmer: logout clears the presetsCollection

PR #7830, opened 2026-04-15
fix(user): migrate lastReultHashes typo to lastResultHashes safely (@mukundangopalachary)

mukundangopalachary:
### Description
This PR fixes a typo in the user hashes field used for duplicate-result detection.

#### Problem:
- The codebase used a misspelled key: `lastReultHashes`.
- This creates long-term schema/DB inconsistency and makes future updates risky.

#### Repro (before this change):
1. Inspect user schema and backend result logic.
2. Notice reads/writes use `lastReultHashes` (typo) instead of `lastResultHashes`.
3. Any new code using the correct key would not align with existing data shape.

#### Approach:
- Read from both keys for backward compatibility:
  - `lastResultHashes ?? lastReultHashes ?? []`
- Write to the corrected key `lastResultHashes`.
- Unset legacy key `lastReultHashes` during updates.
- Keep both keys in schema for compatibility during migration.
- Exclude both keys from relevant user response shaping.
 
### Checks
Closes #
Suggested PR title format (per their rule):
- fix(user): migrate lastReultHashes to lastResultHashes safely (@mukundangopalachary)

### DIFF
--- a/backend/src/api/controllers/user.ts
+++ b/backend/src/api/controllers/user.ts
@@ -509,7 +509,8 @@ type RelevantUserInfo = Omit<
   | "nameHistory"
   | "lastNameChange"
   | "_id"
-  | "lastReultHashes" //TODO fix typo
+  | "lastResultHashes"
+  | "lastReultHashes"
   | "note"
   | "ips"
   | "testActivity"
@@ -524,7 +525,8 @@ function getRelevantUserInfo(user: UserDAL.DBUser): RelevantUserInfo {
     "nameHistory",
     "lastNameChange",
     "_id",
-    "lastReultHashes", //TODO fix typo
+    "lastResultHashes",
+    "lastReultHashes",
     "note",
     "ips",
     "testActivity",
--- a/packages/schemas/src/users.ts
+++ b/packages/schemas/src/users.ts
@@ -243,7 +243,8 @@ export const UserSchema = z.object({
   uid: z.string(), //defined by firebase, no validation should be applied
   addedAt: z.number().int().nonnegative(),
   personalBests: PersonalBestsSchema,
-  lastReultHashes: z.array(z.string()).optional(), //TODO: fix typo (it's in the db too)
+  lastResultHashes: z.array(z.string()).optional(),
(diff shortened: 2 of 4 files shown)


packages/schemas/src/users.ts
```diff
@@ -243,7 +243,8 @@ export const UserSchema = z.object({
   uid: z.string(), //defined by firebase, no validation should be applied
   addedAt: z.number().int().nonnegative(),
   personalBests: PersonalBestsSchema,
-  lastReultHashes: z.array(z.string()).optional(), //TODO: fix typo (it's in the db too)
+  lastResultHashes: z.array(z.string()).optional(),
```
fehmer, line 246: lastResultHashes is not used on the frontend. We should just remove it here and move it to the `DBUser`type.

backend/src/api/controllers/user.ts
```diff
@@ -509,7 +509,8 @@ type RelevantUserInfo = Omit<
   | "nameHistory"
   | "lastNameChange"
   | "_id"
-  | "lastReultHashes" //TODO fix typo
```
fehmer, line 512: this is obsolete when we move the property to the `DBUser`

fehmer (changes requested):
hi @mukundangopalachary, thank you for your contribution.

This topic keeps popping up in PRs so I guess we should fix it.

To tackle this without doing an database migration lets do the following:

- move the `lastResultHashes`from the `UserSchema` to `DBUser`. Frontend is not using this property
- dal writes to the new property, removing the old one (as you already did)
- add the soft-migration to `UserDal.migrateUser`
```ts
function migrateUser<
  T extends Pick<DBUser, "personalBests" | "lastResultHashes">,
>(user: T): T {
  user.personalBests ??= {
    time: {},
    words: {},
    quote: {},
    zen: {},
    custom: {},
  };

  if ("lastReultHashes" in user) {
    user.lastResultHashes = user.lastReultHashes as string[];
  }

  return user;
}
  ```

fehmer:
inactive, #8055

PR #7849, opened 2026-04-18
fix(typed effect): allow fade effect with reduced motion (@d1rshan)

### LINKED ISSUE #7847: Don't apply reduced motion to typed effects
make sure typed effects work even with reduced motion applied. https://github.com/monkeytypegame/monkeytype/discussions/7844

d1rshan:
### Current Behavior for Reduced Motion

1. keep  - Letters stay (No change)                   
2. hide  - Letters vanish instantly (No change)
3. fade - Letters stay **(Effect is lost)**                   
4. dots -  Letters become dots instantly  (has a Fallback)

### Changes
This PR makes sure that `fade` effect works for reduced motion by adding it to the global exclude list for reduced motion.

Closes #7847

### DIFF
--- a/frontend/src/styles/media-queries.scss
+++ b/frontend/src/styles/media-queries.scss
@@ -43,7 +43,12 @@ body {
 
 @media (prefers-reduced-motion) {
   body:not(.ignore-reduced-motion)
-    *:not(.fa-spin, .animate-\[loader\], .preloader) {
+    *:not(
+      .fa-spin,
+      .animate-\[loader\],
+      .preloader,
+      .typed-effect-fade .word.typed
+    ) {
     animation: none !important;
     transition: none !important;
 
--- a/frontend/src/styles/test.scss
+++ b/frontend/src/styles/test.scss
@@ -530,6 +530,10 @@
   &.typed-effect-fade {
     .word.typed {
       animation: fadeOut 250ms ease-in 1 forwards;
+
+      @media (prefers-reduced-motion) {
+        opacity: 0;
+      }
     }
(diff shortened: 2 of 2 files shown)


fehmer:
I think fade should fade even with reduced motion applied. 
Selecting `fade` is a user choice and it is okay to override reduced motion.

d1rshan:
makes sense, i misunderstood it. will update this pr accordingly

fehmer:
hi @d1rshan,

please revert the changes for the `dots` effect. Just replace with dots without the animation is fine for reduced-motion.

Can this be fixed by adding the`.typed-effect-fade` to the exclude list for reduced motion https://github.com/monkeytypegame/monkeytype/blob/5f43b94a10b1c8f348040b7f06b30a5baf9ee082/frontend/src/styles/media-queries.scss#L46 ?

d1rshan:
hey @fehmer, i reverted the changes for `dots` effect and yes adding fade effect class to exclude list also works, so i did that too. 

also can i get clarification regarding what behavior should this new typed effect #7793 follow (for reduced motion) ? - currently i updated it so that it behaves the same on reduced motion by adding to exclude list.

fehmer (approved):

fehmer:
> also can i get clarification regarding what behavior should this new typed effect #7793 follow (for reduced motion) ? - currently i updated it so that it behaves the same on reduced motion by adding to exclude list.

Should be fine, the tumble effect does not make sense without any motion.

Miodec:
Thanks

PR #7851, opened 2026-04-18
fix(profile-search): prevent form.reset() from clearing first character on input (@AzureNightlock)

### LINKED ISSUE #7850: Profile Search textbox doesn't register the first character on input

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

AzureNightlock

### Account config

_No response_

### Current Behavior

When you look up a profile on https://monkeytype.com/profile, The first character that you enter doesn't register and is blank

### Expected Behavior

The textbox should immediately show the first character when its typed into it

### Steps To Reproduce

1. Go to https://monkeytype.com/profile
2. Click on the Profile Lookup textbox
3. Enter any character
4. Notice how the field is blank but the submit button is activated

### Environment

- OS: Windows 11 
- Browser: Google Chrome
- Browser Version: Version 147.0.7727.101 (Official Build) (64-bit)

### Anything else?

Image

The submit button is activated when nothing is present

AzureNightlock:
### Description
This PR: Fixes the first character being skipped on input by only clearing the form when the page isn't open.

Changes Made:
* form.reset() runs only when isOpen is false, Hence clearing the form when the user leaves the page. When the user comes back  to the page, the form is cleared instead of clearing when the user first enters the page

Additional Changes Made:
* Removed the unused catch parameter ("e") from onChangeAsync.
* showIndicator doesn't show up when the field is blank

### Checks

Closes #7850


frontend/src/ts/components/pages/profile/ProfileSearchPage.tsx
```diff
@@ -86,7 +87,7 @@ export function ProfileSearchPage(): JSXElement {
                   <InputField
                     field={field}
                     placeholder="username"
-                    showIndicator
+                    showIndicator={field().state.value.length > 0}
```
fehmer: I would keep the indicator on empty field after the user changed it. The error message states you have to input at least one character and the button to navigate to the profile is deactivated

fehmer (changes requested):
hi @AzureNightlock , thank you for your contribution. I tested your fix it works fine. I left one comment on the `showIndicator`

frontend/src/ts/components/pages/profile/ProfileSearchPage.tsx
AzureNightlock: I mainly added that because if a user types something and then removes it completely, the indicator still stays visible, which feels a little awkward. I can remove it if you'd like.

image

fehmer: Afair that is how all indicators work on the site, example change email modal:

image

If we want to change it we should do it site wide in another pr.

@Miodec what do you think?

AzureNightlock: I reverted the change and kept the indicator as is.

fehmer (approved):

Miodec:
Thanks

PR #7861, opened 2026-04-23
feat(lang): add 6502 assembly language (@szachy-a)

szachy-a:
Add 6502 assembly language
I have added a new language, which is 6502 assembly code. The 6502 CPU was used in many 8-bit machines from the '80s, including Commodore 64, Atari 400/800, Apple II and so on.
Taken from https://en.wikipedia.org/wiki/MOS_Technology_6502
Only using legal original 6502 instruction set (no illegal opcodes or exxtensions like 65c02)


Miodec:
Thanks

PR #7862, opened 2026-04-24
feat(command-line): add profile search (@AzureNightlock)

AzureNightlock:
### Description
This PR: Adds the ability to search for a profile from the command line

Changes Made:
* Added a search profile command in navigation.ts

Additional Changes Made:
* Added an alias to toggleFullscreen command

### Checks

### DIFF
--- a/frontend/src/ts/commandline/lists/navigation.ts
+++ b/frontend/src/ts/commandline/lists/navigation.ts
@@ -51,8 +51,18 @@ const commands: Command[] = [
     },
   },
   {
+    id: "searchProfile",
+    display: "Search Profile",
+    alias: "profile user search find lookup",
+    icon: "fa-search",
+    exec: (): void => {
+      void navigate("/profile");
+    },
+  },
+  {
     id: "toggleFullscreen",
     display: "Toggle Fullscreen",
+    alias: "fullscreen full screen toggle maximize exit escape",
     icon: "fa-expand",
     exec: (): void => {
       toggleFullscreen();
--- a/frontend/src/ts/commandline/lists/navigation.ts
+++ b/frontend/src/ts/commandline/lists/navigation.ts
@@ -55,8 +55,11 @@ const commands: Command[] = [
     display: "Search Profile",
     alias: "profile user search find lookup",
     icon: "fa-search",
-    exec: (): void => {
-      void navigate("/profile");
+    input: true,
+    exec: ({ input }): void => {
+      const username = input?.trim();
+      if (username === undefined || username === "") return;
+      void navigate(`/profile/${username}`);
     },
   },
   {


fehmer:
It would be cool if you could search directly from the command line instead of navigate to the profile search page first.

AzureNightlock:
I have implemented the direct profile lookup change. But I have a doubt, how are users going to access the profile search 
page from within the site?

frontend/src/ts/commandline/lists/navigation.ts
```diff
@@ -50,9 +50,22 @@ const commands: Command[] = [
...
+  {
+    id: "searchProfile",
+    display: "Search Profile",
+    alias: "profile user search find lookup",
+    icon: "fa-search",
+    input: true,
+    exec: ({ input }): void => {
+      const username = input?.trim();
+      if (username === undefined || username === "") return;
+      void navigate(`/profile/${username}`);
```
Leonabcd123, line 62: `username` is injected directly into the url, which means that it will be interpreted as a path. If it contains `..`, it can be used to access pages outside `/profile/xxx`. 

For example, entering `../settings` and pressing enter will navigate to `/settings`.

Maybe we can do something like this:

```diff
diff --git a/frontend/src/ts/commandline/lists/navigation.ts b/frontend/src/ts/commandline/lists/navigation.ts
index 80dbe9f20..37efd95f1 100644
--- a/frontend/src/ts/commandline/lists/navigation.ts
+++ b/frontend/src/ts/commandline/lists/navigation.ts
@@ -1,7 +1,8 @@
 import { navigate } from "../../controllers/route-controller";
 import { isAuthenticated } from "../../states/core";
 import { toggleFullscreen } from "../../utils/misc";
-import { Command } from "../types";
+import { Command, withValidation } from "../types";
+import { UserNameSchema } from "@monkeytype/schemas/users";

 const commands: Command[] = [
   {
@@ -50,18 +51,19 @@ const commands: Command[] = [
       isAuthenticated() ? void navigate("/account") : void navigate("/login");
     },
   },
-  {
+  withValidation({
     id: "searchProfile",
     display: "Search Profile",
     alias: "profile user search find lookup",
     icon: "fa-search",
     input: true,
+    validation: { schema: UserNameSchema },
     exec: ({ input }): void => {
       const username = input?.trim();
       if (username === undefined || username === "") return;
       void navigate(`/profile/${username}`);
     },
-  },
+  }),
   {
     id: "toggleFullscreen",
     display: "Toggle Fullscreen",
```

Leonabcd123 (changes requested):

frontend/src/ts/commandline/lists/navigation.ts
fehmer: good point. We could add the remote validation as well: https://github.com/monkeytypegame/monkeytype/blob/master/frontend/src/ts/components/pages/profile/ProfileSearchPage.tsx#L71-L84

fehmer: this was the validation before we converted to solid: 
https://github.com/monkeytypegame/monkeytype/blob/9d95ad5bcdc03f43e2d44b638e090770233fbd2b/frontend/src/ts/pages/profile-search.ts#L33-L44

AzureNightlock: @Leonabcd123 That is a great idea! But I noticed one UX issue: if the user types something and then clears the input, the validation warning still shows because UserNameSchema requires at least one character.

Do you think we should only validate when the user types something and not when the command line is empty?

fehmer: https://github.com/monkeytypegame/monkeytype/pull/7851#discussion_r3105782224

PR #7863, opened 2026-04-24
fix(challenge): fix requirements (@fehmer)

fehmer:


Leonabcd123 (changes requested):
The following challenges also have a minimum wpm requirement when they shouldn't have:

- What are words at this point?
- Specials
- Aeiou.
- i KINda LikE HoW inEFFICIeNt QwErtY Is.
- One Nauseous Monkey

Can you remove the minimum wpm requirement for those as well?

PR #7868, opened 2026-04-25
impr(turbo): enable tui ui (@d1rshan)

d1rshan:
Add `"ui": "tui"` to `turbo.json` so Turbo uses the TUI by default,
Makes local turbo output cleaner and easier during dev :)

turbo-tui

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "globalDependencies": [".env"],
+  "ui": "tui",
   "tasks": {
     "parallel": {
       "dependsOn": ["^parallel"]


fehmer:
will it fall back to non-tui in workflows or do we have to adjust them?

d1rshan:
> will it fall back to non-tui in workflows or do we have to adjust them?

i looked into it, turbo automatically detects the CI environment variable `CI=true` (which github actions sets by default) and falls back to stream mode. 

so no changes are needed

Miodec:
While its nice, i think i would rather see the logs all at once, especially when working on the backend and frontend at the same time, i wouldnt want to manually switch to see the logs. Maybe this could be configured as a user preference somehow.

PR #7905, opened 2026-05-04
chore: replace hyphens with em dashes (@HeroHunterIguess)

HeroHunterIguess:
### Description

Replacing hyphens used as em dashes with em dashes (—) in UI pages, GitHub documents, and email presets for proper punctuation.


Miodec:
I dont think this is necessary, thanks though.

PR #7917, opened 2026-05-06
refactor: decouple collections from db/snapshot (@fehmer)

fehmer:


frontend/src/ts/db.ts
```diff
@@ -105,37 +96,22 @@ export async function initSnapshot(): Promise<Snapshot | false> {
...
     }
 
-    const userData = userResponse.body.data;
-    const presetsData = presetsResponse.body.data;
     const connectionsData = connectionsResponse.body.data;
 
-    if (userData === null) {
+    if (userData === null || userData === undefined) {
       throw new SnapshotInitError(
         `Request was successful but user data is null`,
```
Leonabcd123:
```suggestion
        `Request was successful but user data is null/undefined`,
```

Leonabcd123 (changes requested):

PR #7918, opened 2026-05-06
impr(input-indicator): add input indicator to all fields with validators (@Leonabcd123)

Leonabcd123:
Remove the `showIndicator` property from `InputField`, instead check whether `validators` exists to decide whether we should show the input indicator.


fehmer:
I havent tested this yet, but maybe we could remove the `showIndicator`property and check for `props.field().options.validators`to be present? 

It is kind of odd to define validations on a field but don't want to show the indicator for it

fehmer (approved):

Miodec:
Thanks

PR #7962, opened 2026-05-17
fix(settings): tag deletion doesn't update results (@fehmer)

fehmer:


frontend/src/ts/collections/results.ts
```diff
@@ -382,6 +400,17 @@ export async function insertLocalResult(
...
 
+export async function deleteLocalTag(
+  params: ActionType["deleteLocalTag"],
+): Promise<void> {
+  if (!resultsCollection.isReady()) {
+    //not loaded yet, don't need to update
+    return;
+  }
+  const transtaction = actions.deleteLocalTag(params);
+  await transtaction.isPersisted.promise;
```
Leonabcd123:
```suggestion
  const transaction = actions.deleteLocalTag(params);
  await transaction.isPersisted.promise;
```

Leonabcd123 (changes requested):

PR #7963, opened 2026-05-17
fix: fetch last 10 average when enabling showAverage (@d1rshan)

### LINKED ISSUE #7961: Modes notice for last10 average not updating
- set show average to none
- reload the page
- set show average to both using the command line

d1rshan:
### Description

Modes notice for last 10 average was not updating when changing showAverage from "off" to a different value after page reload because `Last10Average.update()` never ran (it's only called during test restart) and data stayed at default values (wpm = 0, acc = 0) and hence button was hidden.

### Fix
Fixed by adding a configEvent listener in modes-notice.ts that fetches the average data before updating the UI when `showAverage` is enabled.

Closes #7961


frontend/src/ts/elements/modes-notice.ts
```diff
@@ -183,6 +183,9 @@ export async function update(): Promise<void> {
   }
 
   if (Config.showAverage !== "off") {
+    if (!Last10Average.hasData()) {
```
fehmer: Could we listen to config events for `showAverage` in elements/last-10-average and `update` in case it is not set to `off`? if the data is already fetched it won't fetch again and calculating the average should be cheap enough.

I think the current implementation with additional `hasFetch`state will break because it is never set back to false.

Leonabcd123 (changes requested):
Not sure how you tested this, but it's still broken for me when following the reproduction steps in the bug report. There seems to be a race condition going on between the `update` function in `last-10-average` and `update` in `modes-notice`. This:

https://github.com/monkeytypegame/monkeytype/blob/b570bd1b62e1ccd123c00897ba31408a1ea1eb37/frontend/src/ts/elements/modes-notice.ts#L186-L187

Runs before this:

https://github.com/monkeytypegame/monkeytype/blob/2f00391a84b3375d13aa6d5a03221d6b99515c9d/frontend/src/ts/elements/last-10-average.ts#L17-L18

d1rshan:
hey @Leonabcd123 yeah ik i thought it worked because i left a stale line in modes-notice.ts to update everytime 😭 (that is what i force pushed ) i realized about the race condition, i'm figuring out what is the best solution here.

we could do this, but will trigger ModesNotice.update() twice, so not sure if this is a good idea.

configEvent.subscribe(({ key, newValue }) => {
  if (key === "showAverage" && newValue !== "off") {
    void update().then(() => ModesNotice.update());
  }
});

frontend/src/ts/elements/modes-notice.ts
d1rshan: <del>
yeah the hasFetch approach fails because we didn't reset it, sry didnt think about this scenario.

i tried the approach u asked for and there is a race condition https://github.com/monkeytypegame/monkeytype/pull/7963#pullrequestreview-4305508973. 
there is a work around of doing  this -> `void update().then(() => ModesNotice.update())` but this will trigger modes notice update twice, i'm not sure if this is okay tbh.

or we can directly fix the issue by modiying the configEvent.subscribe in modes-notice.ts to trigger update for Last10Average and this wont call the ModeNotice.update() twice.
https://github.com/monkeytypegame/monkeytype/blob/b570bd1b62e1ccd123c00897ba31408a1ea1eb37/frontend/src/ts/elements/modes-notice.ts#L43-L45

pls let me know which approach i should go with, i will update the fix accordingly. thanks!
</del>

d1rshan: update: 
i didn't realize we could remove `showAverage` key in the modes notice config event subscription as we are taking care of it via this seperate listener so there wont be modes notice triggering twice at all.
also i have put this logic in modes-notice.ts file itself to prevent circular dependency.

fehmer:
@d1rshan  thanks for the fix anyways

PR #7970, opened 2026-05-19
docs: add network hosting and reverse proxy troubleshooting (@isma-021)

isma-021:
### Description 

I noticed that when trying to host Monkeytype over a local network or the internet instead of `localhost`, the frontend crashes with an `Uncaught TypeError: crypto.randomUUID is not a function` error. This happens because modern browsers restrict this API to secure contexts (HTTPS). 

When putting a reverse proxy in front to handle HTTPS, it's very easy to run into Mixed Content blocks or 404 errors due to incorrect `.env` configurations (like adding a trailing slash to `MONKEYTYPE_FRONTENDURL` or forgetting to completely recreate the containers).

This PR updates `SELF_HOSTING.md` to:
- Explain the HTTPS requirement for network deployments.
- Detail how to properly configure the backend URL in the `.env` file to avoid API 404 errors (the trailing slash issue).
- Highlight the importance of using `docker compose up -d --force-recreate` and clearing the browser cache, since environment variables are baked into the SPA static files at startup.

### Checks

- [X] Make sure to include your GitHub username prefixed with @ inside parentheses at the end of the PR title.

### DIFF
--- a/docs/SELF_HOSTING.md
+++ b/docs/SELF_HOSTING.md
@@ -10,6 +10,7 @@
   - [Table of contents](#table-of-contents)
   - [Prerequisites](#prerequisites)
   - [Quickstart](#quickstart)
+    - [Hosting over the network (HTTPS)](#hosting-over-the-network-https)
   - [Account System](#account-system)
     - [Setup Firebase](#setup-firebase)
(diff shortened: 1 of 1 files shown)


Miodec:
Thanks!

PR #7975, opened 2026-05-19
fix(blind-mode): on button has incorrect height on firefox (@Leonabcd123)

Leonabcd123:
image

For some reason, Chromium and Firefox differ on the height of a button containing only `&lrm;`, use `&ensp;` like we did before 1cefe7cf40c1ed42cb68dc0e837a07bcfc5d7325.

### DIFF
--- a/frontend/src/ts/config/metadata.tsx
+++ b/frontend/src/ts/config/metadata.tsx
@@ -304,7 +304,7 @@ export const configMetadata: ConfigMetadataObject = {
     fa: { icon: "fa-eye-slash" },
     optionsMetadata: {
       true: {
-        displayString: "‎",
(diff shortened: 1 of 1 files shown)


frontend/src/ts/config/metadata.tsx
```diff
@@ -304,7 +304,7 @@ export const configMetadata: ConfigMetadataObject = {
     fa: { icon: "fa-eye-slash" },
     optionsMetadata: {
       true: {
-        displayString: "‎",
+        displayString: " ",
```
fehmer, line 309: could you add a comment why we have this

Miodec:
Thanks

PR #7984, opened 2026-05-20
impr(quotes): add english quotes (@TemperTimothy)

TemperTimothy:
Has to be added to end of existing code because for me it says that the raw code is too big to show so i couldn't edit it.

Added kryptos quotes and poppy playtime quotes (english).

Sorry if adding this to the end of the existing code is an issue.
I put a post in bug reports for help.


fehmer (changes requested):
diff should not be 40k

TemperTimothy:
> diff should not be 40k

I added al of the existing quotes back if that's what you mean.

fehmer:
> > diff should not be 40k
> 
> I added al of the existing quotes back if that's what you mean.

you need to get this numbers down to <100
image

TemperTimothy:
> > > diff should not be 40k
> > 
> > 
> > I added al of the existing quotes back if that's what you mean.
> 
> you need to get this numbers down to <100 image

How would I do that? I think it's showing that because I added all the existing quotes manually.

fehmer:
don't add all the existing quotes manually

TemperTimothy:
> don't add all the existing quotes manually

I'll probably delete this pull request and delete this repo because I can only fork one repo at any given moment.

PR #7996, opened 2026-05-21
fix: prevent collection cleanup (@fehmer)

fehmer:
But only occurs on prod  builds. After a while the tags collection gets cleaned up because there is no active query. 
The commandline fetches the tags each time on open, causing them to "disappear" from the commandline after a while.

As a workaround we keep one liveQuery open for each collection exposing __nonreactive

### DIFF
--- a/frontend/src/ts/collections/custom-themes.ts
+++ b/frontend/src/ts/collections/custom-themes.ts
@@ -44,6 +44,10 @@ const customThemesCollection = createCollection(
         );
       }
 
+      if (_keepAlive === null) {
+        _keepAlive = useCustomThemesLiveQuery();
+      }
+
       return response.body.data.map(applyIdWorkaround);
     },
   }),
@@ -174,3 +178,10 @@ export const __nonReactive = {
   getCustomThemes,
   getCustomTheme,
 };
+
+/**
+ * On prod the collection gets cleaned up after a while.
+ * Keeping a query active fixes that. Remove when removing __nonReactive
+ */
+// oxlint-disable-next-line typescript/no-explicit-any
+let _keepAlive: any = null;
--- a/frontend/src/ts/collections/presets.ts
+++ b/frontend/src/ts/collections/presets.ts
@@ -203,3 +203,9 @@ export const __nonReactive = {
   getPresets,
   getPreset,
 };
+
+/**
+ * On prod the collection gets cleaned up after a while.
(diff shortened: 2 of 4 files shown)


Leonabcd123 (commented):
About the approach: wouldn't it be simpler to just disable garbage collection for these collections (by setting `gcTime` to 0)?

fehmer:
I tried that, doesn't work. The collection gets cleaned up after five minutes, which is the default value for gcTime. Might be a bug in tanstack/db. I'll try to create a minimal sample and report a bug.

~~Or it is something else. After five minutes the state of the collection is still ready but it is empty. Setting the  `gcTime` to 10 seconds causes the collection to be in `cleaned-up` state~~

When we introduced tags collection in https://github.com/monkeytypegame/monkeytype/commit/2f00391a84b3375d13aa6d5a03221d6b99515c9d#diff-fe1913383773926c10c4c348ad18d3aab7fec6a22f9567b922de32d76b5b6d29 we also added a live query in settings.ts keeping this collection alive

Leonabcd123:
I think I figured it out:

Tanstack/db is not the only one responsible for garbage collecting. It also uses Tanstack/query, which has its own `gcTime` (also defaults to 5 minutes). When Tantstack/query GC'es a query, Tanstack/db checks if tracks it, and if it does, it cleans it up as well. This is why we get the following warning:

```
[cleanupQueryIfIdle] Invariant violation: refcount=1 but no listeners. Cleaning up to prevent leak. 
Object { hashedQueryKey: '["user","tags"]' }
```

Tanstack/query has already cleaned it up, so Tanstack/db cleans up as well. I guess we can fix it by changing the `queryClient`'s options `gcTime` to `Infinity`, but I'm not sure if that is even better than the current approach.

Relevant code in Tanstack/db:

https://github.com/TanStack/db/blob/e21ef3246e80decbf57134defbc0e41230363360/packages/query-db-collection/src/query.ts#L1624-L1637

Note that this was all tested on dev, I'm not sure why you can only reproduce on prod.

Edit: Discussed on discord, #7997 was opened (which is based on the suggested approach here), but at the end it was decided that the approach in this pr is better. Also a bug was opened in https://github.com/TanStack/db/issues/1546, which would simplify the workaround used in #7997.

Edit 2: https://github.com/TanStack/db/issues/1546 was fixed, #8088 was opened, which simplifies the approach taken in #7997.

Miodec (approved):
Thanks

PR #8000, opened 2026-05-22
refactor: improve simple modals (@fehmer)

fehmer:


frontend/src/ts/components/ui/form/InputField.tsx
```diff
@@ -16,10 +16,13 @@ export function InputField(props: {
   dir?: "ltr" | "rtl" | "auto";
   maxLength?: number;
   onFocus?: () => void;
+  nativeType?: boolean;
```
fehmer: unused, remove

Miodec (approved):

Issue #8007, opened 2026-05-23
Custom Text: Limit conflict error on mode change


akartsky:
### Did you clear cache before opening an issue?

- [x] I have cleared my cache

### Is there an existing issue for this?

- [x] I have [searched](https://github.com/monkeytypegame/monkeytype/issues?q=is%3Aissue) the existing open and closed issues

### Does the issue happen when logged in?

Yes

### Does the issue happen when logged out?

Yes

### Does the issue happen in incognito mode when logged in?

Yes

### Does the issue happen in incognito mode when logged out?

Yes

### Account name

-

### Account config

-

### Current Behavior

In custom text mode, switching to a different word generation mode incorrectly triggers the validation error: "You can only specify one limit". And it does not let you proceed making the feature unusable.

### Expected Behavior

There shouldn't be any error and should let you proceed

### Steps To Reproduce

1. Go to custom text mode 
2. Press change
3. [Optional] Add some new text 
4. Change the mode to any other mode than "simple". Eg: Random
5. Press ok 
6. Results in error notice "You can only specify one limit" 

### Environment

- OS: Windows 11
- Browser: Google Chrome
- Browser Version: Version 148.0.7778.179 (Official Build) (64-bit)

### Anything else?

Workaround : You can change the word limit number which clears the other limit fields and the issue goes away.

akartsky:
Just following up since it's been a few days without any updates. 

I've also opened a PR with a one-line fix, if someone could take a quick look. 
Thanks!

_(Screenshot: Blocks users from proceeding while using different modes in custom text)_
<img width="1935" height="996" alt="Image" src="https://github.com/user-attachments/assets/6ab65dff-c988-4cfb-be29-a0eae545f3bf" />

Leonabcd123:
Fixed by 562a17ab9d6958b147de15e6b40f9feda3d60c1a, can you close this please?

PR #8008, opened 2026-05-23
fix(CustomTextModal): Resolve limit validation error on mode switch (@akartsky)

akartsky:
Fixes bug https://github.com/monkeytypegame/monkeytype/issues/8007

-----
### Root cause :
Looks like there is a ["limitSection"](https://github.com/monkeytypegame/monkeytype/blob/35a211854b528d356766228c3be28be875605f5a/frontend/src/ts/components/modals/CustomTextModal.tsx#L386) (hidden in UI) which was also being set during mode switch. Resulting in [validation error](https://github.com/monkeytypegame/monkeytype/blob/35a211854b528d356766228c3be28be875605f5a/frontend/src/ts/components/modals/CustomTextModal.tsx#L83-L93).

### Testing :
I temporarily made the hidden limit 'sections' input visible just to debug and validate the fix

### DIFF
--- a/frontend/src/ts/components/modals/CustomTextModal.tsx
+++ b/frontend/src/ts/components/modals/CustomTextModal.tsx
@@ -383,7 +383,7 @@ export function CustomTextModal(): JSXElement {
         const text = cleanUpText();
         form.setFieldValue("limitWord", `${text.length}`);
         form.setFieldValue("limitTime", "");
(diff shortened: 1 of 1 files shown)


Miodec:
Slight tweak - we need to either apply section or word limit based on the pipeDelimiter value. Thanks for the pr.

PR #8019, opened 2026-05-27
fix: unresponsive cookies modal (@d1rshan)

d1rshan:
<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>


Miodec:
Thanks, changed it to max so we can still control the width, otherwise we are at the mercy of whatever width the browser decides and it could change depending on the content inside.

PR #8020, opened 2026-05-28
fix(account): fix "load more" button state for filtered results (@d1rshan)

d1rshan:
"load more" button used `getResultsSize()` (unfiltered total) to decide if more results exist. With active filters this always showed the button as enabled. 

### Fix
  - fetch `limit + 1` filtered results
  - render only the first `limit`
  - use the extra row to determine whether more results exist

### DIFF
--- a/frontend/src/ts/components/pages/account/AccountPage.tsx
+++ b/frontend/src/ts/components/pages/account/AccountPage.tsx
@@ -4,7 +4,6 @@ import { createMemo, createSignal, JSXElement, Show } from "solid-js";
 import {
   createResultsQueryState,
   getResultsQueryOnce,
-  getResultsSize,
   useResultsLiveQuery,
 } from "../../../collections/results";
 import { SnapshotResult } from "../../../constants/default-snapshot";
@@ -124,7 +123,8 @@ export function AccountPage(): JSXElement {
                   <Button
                     text="load more"
                     disabled={
-                      resultsQuery.isLoading || getResultsSize() < limit() + 10
+                      resultsQuery.isLoading ||
+                      resultsQueryData().length < limit()
                     }
                     onClick={() => setLimit((limit) => limit + 10)}
                     class="w-full text-center"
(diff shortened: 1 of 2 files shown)


frontend/src/ts/components/pages/account/AccountPage.tsx
```diff
@@ -4,7 +4,6 @@ import { createMemo, createSignal, JSXElement, Show } from "solid-js";
 import {
   createResultsQueryState,
   getResultsQueryOnce,
-  getResultsSize,
```
fehmer, line 7: Can you please remove this function from collections/results, this was the only usage

d1rshan: done

frontend/src/ts/components/pages/account/AccountPage.tsx
```diff
@@ -46,7 +45,13 @@ export function AccountPage(): JSXElement {
   );
   const [isExporting, setIsExporting] = createSignal(false);
 
-  const resultsQuery = useResultsLiveQuery({ queryState, sorting, limit });
+  const queryLimit = createMemo(() => limit() + 1);
```
fehmer: do we need the memo for queryLimit and hasMoreResults or could we just fetch one more with

```ts
  const resultsQuery = useResultsLiveQuery({
    queryState,
    sorting,
    limit: () => limit() + 1,
  });
```

and disable based on the query data
```ts
 <Button
  disabled={
    resultsQuery.isLoading ||
    resultsQueryData().length <= limit()
  }
```

fehmer:
hi @d1rshan , nice solution. Few comments on the pr

frontend/src/ts/components/pages/account/AccountPage.tsx
d1rshan: oh thanks, we don't need that. 
i simplified it now.

fehmer (approved):

PR #8026, opened 2026-05-29
docs(rate-limit): update comments in rate limit file (@Leonabcd123)

Leonabcd123:
- Capitalization
- Add another comment for `5000` ms
- Change every `x min` to `x minutes` (could be changed the other way, just important to be consistent)
- Update `resultsMismatchReport` comment to be `5 minutes` instead of `15 min`.


Miodec (approved):
Thanks

PR #8030, opened 2026-05-30
feat(quotes): add Witch Hat Atelier quote (@donmaruko)

donmaruko:
Added a quote from Witch Hat Atelier (manga). The quote is in English and sourced from the official English translation.


Miodec:
Thanks

PR #8031, opened 2026-05-30
refactor: solid practise words modal (@d1rshan)

d1rshan:
- migrates practice words modal to solidjs

image


frontend/src/ts/commandline/lists/result-screen.ts
```diff
@@ -50,11 +50,8 @@ const practiceSubgroup: CommandsSubgroup = {
...
       display: "custom...",
       opensModal: true,
-      exec: (options): void => {
-        PractiseWordsModal.show({
-          animationMode: "modalOnly",
-          modalChain: options.commandlineModal,
-        });
+      exec: (): void => {
+        showModal("PractiseWords");
       },
```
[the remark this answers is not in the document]
d1rshan: commandline is correctly closed now.

frontend/src/ts/components/modals/PractiseWordsModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  return (
+    <AnimatedModal
+      id="PractiseWords"
+      title="Practice words"
+      modalClass="max-w-[400px]"
+      animationMode="modalOnly"
+    >
+      <form
```
fehmer, line 35: I wonder if we could use ts/forms and create a ButtonGroup component. Part of it is already done in AutoSettings component.

This way we don't need to maintain the signals

frontend/src/ts/components/modals/PractiseWordsModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+        onSubmit={(e) => {
+          e.preventDefault();
+          apply();
+        }}
+      >
+        <div class="grid gap-6">
+          <div class="grid gap-2">
```
fehmer: This style is identical to the settings page. Could we extract a component for it and use it here and on the settings page

frontend/src/ts/components/modals/PractiseWordsModal.tsx
d1rshan: da99797fa73d86c58dd15180ee83c224458cf438,@fehmer like this?

PR #8034, opened 2026-05-30
fix(utils): preserve newlines in sanitizeString (@bijanmurmu)

### LINKED ISSUE #8033: Profile bio loses line breaks (newlines) if they are adjacent to spaces

- [x] I have cleared my cache

Yes

No

Yes

No

### Account name

bijanmurmu

### Account config

{"theme":"serika_dark","themeLight":"serika","themeDark":"serika_dark","autoSwitchTheme":false,"customTheme":false,"customThemeColors":["#323437","#e2b714","#e2b714","#646669","#2c2e31","#d1d0c5","#ca4754","#7e2a33","#ca4754","#7e2a33"],"favThemes":[],"showKeyTips":true,"smoothCaret":"medium","codeUnindentOnBackspace":false,"quickRestart":"off","punctuation":false,"numbers":false,"words":25,"time":15,"mode":"time","quoteLength":[1],"language":"english","fontSize":2,"freedomMode":false,"difficulty":"normal","blindMode":false,"quickEnd":false,"caretStyle":"default","paceCaretStyle":"default","flipTestColors":false,"layout":"default","funbox":[],"confidenceMode":"off","indicateTypos":"off","compositionDisplay":"replace","timerStyle":"mini","liveSpeedStyle":"off","liveAccStyle":"off","liveBurstStyle":"off","colorfulMode":false,"randomTheme":"off","timerColor":"main","timerOpacity":"1","stopOnError":"off","showAllLines":false,"keymapMode":"off","keymapStyle":"staggered","keymapLegendStyle":"lowercase","keymapLayout":"overrideSync","keymapShowTopRow":"layout","keymapSize":1,"fontFamily":"Roboto_Mono","smoothLineScroll":false,"alwaysShowDecimalPlaces":false,"alwaysShowWordsHistory":false,"singleListCommandLine":"on","capsLockWarning":true,"playSoundOnError":"off","playSoundOnClick":"off","soundVolume":0.5,"startGraphsAtZero":true,"showOutOfFocusWarning":true,"paceCaret":"off","paceCaretCustomSpeed":100,"repeatedPace":true,"accountChart":["on","on","on","on"],"minWpm":"off","minWpmCustomSpeed":100,"highlightMode":"letter","typedEffect":"keep","typingSpeedUnit":"wpm","ads":"result","hideExtraLetters":false,"strictSpace":false,"minAcc":"off","minAccCustom":90,"monkey":false,"repeatQuotes":"off","resultSaving":true,"oppositeShiftMode":"off","customBackground":"","customBackgroundSize":"cover","customBackgroundFilter":[0,1,1,1],"customLayoutfluid":["qwerty","dvorak","colemak"],"customPolyglot":["english","spanish","french","german"],"monkeyPowerLevel":"off","minBurst":"off","minBurstCustomSpeed":100,"burstHeatmap":false,"britishEnglish":false,"lazyMode":false,"showAverage":"off","showPb":false,"tapeMode":"off","tapeMargin":50,"maxLineWidth":0,"playTimeWarning":"off"}

### Current Behavior

If a user adds a newline in their bio but leaves spaces around it (for example, `"Hello \n \n World"`), the `sanitizeString` function collapses the content into a single line.

This happens because the regex `/\s{3,}/g` matches newline characters in addition to spaces, causing intended line breaks to be removed.

### Expected Behavior

The `sanitizeString` function should only collapse consecutive horizontal whitespace (spaces and tabs) while preserving intentional newlines.

Multiple consecutive newlines (3+) should continue to be handled by the existing newline-specific sanitization logic.

### Steps To Reproduce

1. Navigate to the profile settings page.
2. Edit your bio.
3. Enter text, add a space, press Enter, add another space, then enter more text.
4. Save the profile.
5. Observe that the newline is removed and the text is merged into a single line.

### Environment

- OS: Windows 11
- Browser: Google Chrome
- Browser Version: 148.0.7778.217 (Official Build) (64-bit)

### Anything else?

How I faced it: I noticed some user bios had their paragraphs completely merged into one line unintentionally. I checked the `sanitizeString` logic and saw the `\s` regex was the culprit. I would love to submit a PR to fix this if approved!

bijanmurmu:
Closes #8033.
Changes the regex in  sanitizeString  to only collapse horizontal spaces so it no longer accidentally deletes newlines.

This PR fixes an issue where the `sanitizeString` utility deletes intentional newlines if they are surrounded by spaces. The regex `/\s{3,}/g` was replaced with `/[^\S\r\n]{3,}/g` so that it only collapses horizontal whitespace and preserves user line breaks (such as those used in profile bios).

### DIFF
--- a/backend/src/utils/misc.ts
+++ b/backend/src/utils/misc.ts
@@ -124,7 +124,7 @@ export function sanitizeString(str: string | undefined): string | undefined {
     .replace(/[\u0300-\u036F]/g, "")
     .trim()
     .replace(/\n{3,}/g, "\n\n")
-    .replace(/\s{3,}/g, "  ");
+    .replace(/[^\S\r\n]{3,}/g, "  ");
 }
 
 const suffixes = ["th", "st", "nd", "rd"];
--- a/backend/src/utils/misc.ts
+++ b/backend/src/utils/misc.ts
@@ -1,5 +1,6 @@
 import { MILLISECONDS_IN_DAY } from "@monkeytype/util/date-and-time";
 import { roundTo2 } from "@monkeytype/util/numbers";
+export { sanitizeString } from "@monkeytype/util/strings";
 import uaparser from "ua-parser-js";
 import { MonkeyRequest } from "../api/types";
 import { ObjectId } from "mongodb";
@@ -113,18 +114,6 @@ export function flattenObjectDeep(
   });
 
   return result;
-}
-
-export function sanitizeString(str: string | undefined): string | undefined {
-  if (str === undefined || str === "") {
-    return str;
-  }
-
-  return str
-    .replace(/[\u0300-\u036F]/g, "")
-    .trim()
-    .replace(/\n{3,}/g, "\n\n")
-    .replace(/[^\S\r\n]{3,}/g, "  ");
 }
 
 const suffixes = ["th", "st", "nd", "rd"];
--- a/packages/schemas/src/validation/validation.ts
+++ b/packages/schemas/src/validation/validation.ts
@@ -1,4 +1,5 @@
 import { replaceHomoglyphs } from "./homoglyphs";
+import { sanitizeString } from "@monkeytype/util/strings";
 import { ZodEffects, ZodString } from "zod";
 
 // Sorry for the bad words
@@ -390,18 +391,6 @@ const disallowedWords = [
   "wichser",
   "zabourah",
 ];
-
-function sanitizeString(str: string | undefined): string | undefined {
-  if (str === undefined || str === "") {
-    return str;
-  }
-
-  return str
-    .replace(/[\u0300-\u036F]/g, "")
-    .trim()
-    .replace(/\n{3,}/g, "\n\n")
-    .replace(/[^\S\r\n]{3,}/g, "  ");
-}
 
 function containsDisallowedWords(
   text: string,
--- a/packages/schemas/src/validation/validation.ts
+++ b/packages/schemas/src/validation/validation.ts
@@ -400,7 +400,7 @@ function sanitizeString(str: string | undefined): string | undefined {
     .replace(/[\u0300-\u036F]/g, "")
     .trim()
     .replace(/\n{3,}/g, "\n\n")
-    .replace(/\s{3,}/g, "  ");
+    .replace(/[^\S\r\n]{3,}/g, "  ");
 }
 
 function containsDisallowedWords(
--- a/backend/__tests__/utils/misc.spec.ts
+++ b/backend/__tests__/utils/misc.spec.ts
@@ -197,43 +197,6 @@ describe("Misc Utils", () => {
     );
   });
 
-  it("sanitizeString", () => {
-    const testCases = [
-      {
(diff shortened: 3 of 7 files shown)


packages/schemas/src/validation/validation.ts
fehmer, line 394: Why do we have this twice. Can we call misc.sanitizeString instead?

bijanmurmu: Actually, I decided to revert that extraction. Since `schemas` cannot safely import from `backend`, the only way to avoid duplication is extracting it to `@monkeytype/util`. However, I thought it would be safer to just keep the PR scope small and leave the duplication for now. How would you like me to proceed?

packages/schemas/src/validation/validation.ts
```diff
@@ -400,7 +400,7 @@ function sanitizeString(str: string | undefined): string | undefined {
     .replace(/[\u0300-\u036F]/g, "")
     .trim()
     .replace(/\n{3,}/g, "\n\n")
-    .replace(/\s{3,}/g, "  ");
+    .replace(/[^\S\r\n]{3,}/g, "  ");
```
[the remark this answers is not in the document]
bijanmurmu: Applied the exact same fix here as well. Since the function is currently duplicated in the  schemas  package, I wanted to make sure the bug was completely squashed everywhere it's used.

backend/src/utils/misc.ts
```diff
@@ -124,7 +124,7 @@ export function sanitizeString(str: string | undefined): string | undefined {
     .replace(/[\u0300-\u036F]/g, "")
     .trim()
     .replace(/\n{3,}/g, "\n\n")
-    .replace(/\s{3,}/g, "  ");
+    .replace(/[^\S\r\n]{3,}/g, "  ");
```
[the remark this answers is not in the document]
bijanmurmu: Applied the regex fix here to ensure newlines aren't accidentally deleted by the backend when users save the profile.

packages/schemas/src/validation/validation.ts
fehmer: Hi, 

I think this fits perfectly in the PR scope as you are fixing a problem with the sanitizeString function.

Moving it to https://github.com/monkeytypegame/monkeytype/blob/master/packages/util/src/strings.ts and use this in schema/validation and backend code seems the cleanest.

Please also move the test for sanitizeString from backend to the utils package and update it with a test case for the bug you are fixing.

You can earn extra points for converting the existing test to use  parameterized tests `it.each` instead. Example for this is https://github.com/monkeytypegame/monkeytype/blob/master/backend/__tests__/utils/misc.spec.ts#L35

bijanmurmu: Hi @fehmer,

Thanks for the suggestions! I have implemented them:

• Moved  `sanitizeString`  to  `packages/util/src/strings.ts`  and updated the  `schema/validation`  and  `backend`  code to use the import.
• Moved the tests to the utils package and added the test case covering the newline bug.
• Converted the tests to use  `it.each`  as suggested for the extra points!

Let me know if everything looks good now.

Miodec:
Thanks

bijanmurmu:
Thanks for the merge and the review, appreciate it!

PR #8044, opened 2026-06-02
refactor: use solid simple modals (@fehmer)

fehmer:


frontend/src/ts/components/modals/account-settings/RemoveAuthMethodModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  if (!hasRemainingAuth) {
+    showNoticeNotification("No remaining authentication enabled");
+    return;
+  }
+
+  const provider =
+    options.authMethod === "password"
+      ? "password"
+      : options.authMethod === "github.com"
+        ? "Github"
```
Leonabcd123: `Github` should be formatted as `GitHub` as it is on master.

```suggestion
        ? "GitHub"
```

frontend/src/ts/components/modals/account-settings/RemoveAuthMethodModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+          message,
+        };
+      }
+
+      options.callback();
+
+      reloadAfter(3);
+      return {
+        status: "success",
+        message: `${provider} authentication removed`,
```
Leonabcd123: Note that on master `provider` is capitalized for all cases (`Google`, `GitHub`, and `Password`) in this message, but here if `provider = password` it won't be capitalized.

Leonabcd123 (changes requested):

frontend/src/ts/components/modals/account-settings/RemoveAuthMethodModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    isUsingGithubAuthentication() && options.authMethod !== "github.com",
+    isUsingGoogleAuthentication() && options.authMethod !== "google.com",
+  ].find((it) => it);
+
+  if (!hasRemainingAuth) {
+    showNoticeNotification("No remaining authentication enabled");
+    return;
+  }
+
+  const methodDisplay =
```
Leonabcd123: For this, I think that using an object to convert `options.authMethod` to `provider` is better than the current approach with ternaries, as it's more readable imo and it allows for easier expansion when adding more auth methods.

Something like this:

```
const authMethodToMethodDisplay: Record<AuthMethod, string> = {
  password: "Password",
  "google.com": "Google",
  "github.com": "GitHub",
};
```

Leonabcd123 (commented):
For your consideration, more of a style comment.

frontend/src/ts/utils/firebase-auth.ts
```diff
@@ -0,0 +1,143 @@
...
+  reauthenticateWithPopup,
+  User,
+} from "firebase/auth";
+import { getAuthenticatedUser, isAuthAvailable } from "../firebase";
+import { FirebaseError } from "firebase/app";
+import { isDevEnvironment } from "./env";
+import { PasswordSchema } from "@monkeytype/schemas/users";
+import { z, ZodString } from "zod";
+
+export type AuthMethod = "password" | "github.com" | "google.com";
```
Leonabcd123: I think we should first define an array with all of these values, then infer a union type out of it. This will remove

`const authMethods = ["password", "github.com", "google.com"] as AuthMethod[];`

in `getPreferredAuthenticationMethod`, which would make it easier to expand `authMethods` (only need to change the array instead of both the array and the type).

```suggestion
const authMethods = ["password", "github.com", "google.com"] as const;
export type AuthMethod = typeof authMethods[number];
```

And remove `authMethods` definition in `getPreferredAuthenticationMethod`.

frontend/src/ts/components/modals/account-settings/RemoveAuthMethodModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  callback: () => void;
+}): void {
+  if (!isAuthenticated()) return;
+
+  //check there is at least one authentication remaining
+  const hasRemainingAuth = [
+    isUsingPasswordAuthentication() && options.authMethod !== "password",
+    isUsingGithubAuthentication() && options.authMethod !== "github.com",
+    isUsingGoogleAuthentication() && options.authMethod !== "google.com",
+  ].find((it) => it);
```
Leonabcd123: This suggestion should be addressed after the previous review comment.

```suggestion
const hasRemainingAuth = authMethods.some((authMethod) => isUsingAuthentication(authMethod) && options.authMethod !== authMethod);
```

Also need to export `isUsingAuthentication` and `authMethods`, and remove `isUsingGithubAuthentication` and `isUsingGoogleAuthentication`.

Leonabcd123 (changes requested):

Miodec (approved):
Go go gadget

PR #8054, opened 2026-06-04
feat: add header streak indicator (@hunterpaulson)

hunterpaulson:
### Description

Problem being solved: 

In order to see your current streak, and whether it was claimed today, you have to click your profile then hover over your streak under your username.

Solution:

Add a streak indicator in the header next to the notification bell that shows your current streak along with a fire icon to indicate whether or not you claimed your streak for today.

image

### Design

- When the streak is not claimed yet today the icon is hollow and its border uses the theme's `sub` color. matching the rest of the heading
- When the streak *is claimed* for the day the icon uses the theme's `main` color with solid fill. matching the monkeytype keyboard logo
- The streak number uses the theme's normal `text` color and inherits the existing header navigation font size. 
    - used tabular numerals to prevent the layout from shifting as the number changes

### examples

Logged out. shows no streak indicator (notice nothing next to the bell)

Screenshot 2026-06-03 190517

Before streak is claimed. Logged in users see a hollow indicator and the length of the current streak. hinting to them to begin/claim their streak for the day.

Screenshot 2026-06-03 182530

After streak is claimed. Indicator becomes solid and uses main color

Screenshot 2026-06-03 182603

Timezone streak hour offset tool tip still shows when hovering streak on profile page

Screenshot 2026-06-03 184920

When streak is lost it goes back to zero. 

Screenshot 2026-06-03 182825

### Implementation notes

The account page now displays a one-day streak instead of waiting until day two, keeping it consistent with the header indicator. [previously it took two days to show streak indicator](https://github.com/monkeytypegame/monkeytype/discussions/4155).

Moved the streak formatting and tooltip logic from the profile component into `utils/streak.ts` so the header and account page share the same behavior and wording. account-only tooltips like the changing timezone hour offset remain opt-in.

In order to determine whether today's streak was claimed immediately after login, before result history is first loaded I used the existing `streak.lastResultTimestamp` value from the user response as `streakLastResultTimestamp`. It is also updated after saving a result so the indicator changes immediately.

### Checks

related discussion [#5799](https://github.com/monkeytypegame/monkeytype/discussions/5799)

### DIFF
--- a/frontend/__tests__/utils/streak.spec.ts
+++ b/frontend/__tests__/utils/streak.spec.ts
@@ -1,0 +1,245 @@
+import type { Mode } from "@monkeytype/schemas/shared";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import type { SnapshotResult } from "../../src/ts/constants/default-snapshot";
+import {
+  getStreakExtraText,
+  getStreakIndicatorState,
+  getStreakHoverText,
+  hasClaimedStreakToday,
+} from "../../src/ts/utils/streak";
+
+function resultAt(timestamp: number): SnapshotResult<Mode> {
+  return { timestamp } as SnapshotResult<Mode>;
+}
+
+describe("streak utils", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("hides the header indicator when signed out", () => {
+    expect(getStreakIndicatorState(undefined, undefined)).toEqual({
+      show: false,
+      claimedToday: false,
+      hoverText: "",
+    });
+  });
+
+  it("shows a hollow flame with a zero label before the first streak claim", () => {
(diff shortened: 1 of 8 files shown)


fehmer:
hi @hunterpaulson , thank you for your contribution, this is a nice feature.

I think we should create a signal for the StreakSnapshot/StreakIndicatorState` in `states/streak.ts`and update it based on the snapshot + every 60 seconds.

`UserDetails`and  `StreakIndicator` can just render based on the signal.

hunterpaulson:
agreed. I moved the streak state into states/streak.ts and updated the header indicator and account profile display to render from the shared state.

hunterpaulson:
@fehmer I fixed the merge conflicts and refactored to maintain state

fehmer:
fixes #8254

PR #8060, opened 2026-06-04
fix(quick-restart): handle quick restart in long tests (@Leonabcd123)

### LINKED ISSUE #8059: Quick restart isn't respected after switching to long test

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

RobiRabbit

### Account config

_No response_

### Current Behavior

After switching to a long test, pressing the quick restart key doesn't restart the test, and when trying to press it mid test it won't show the expected notification which says that quick restart is disabled in long tests.

Everything works fine after reloading.

### Expected Behavior

Pressing the quick restart key before the test has started should restart it, pressing it mid test should show a notification.

### Steps To Reproduce

1. Set quick restart key to `tab`
3. Set `Test duration` to `0`
4. Press `tab`
5. Notice how the test hasn't restarted

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 150.0.3

### Anything else?

_No response_

Leonabcd123:
Fixes #8059

The idea of the fix is to create both hotkeys with a single function (`createHotkeys`), which guarantees that the second hotkey in the array passed to it will always be updated last, and thus won't be overridden by the first one. This means that when the primary and secondary hotkeys are both set to `tab` (or whatever the quick restart key is set to), the primary hotkey will override the secondary one, rather then the opposite happening.


fehmer:
We recently updated tanstack/hotkeys. Could you verify the issue already existed with the old version? Thanks

Leonabcd123:
> We recently updated tanstack/hotkeys. Could you verify the issue already existed with the old version? Thanks

Tested on 4a1e6f7ee1ac4bc652763bc38d1bd5dc0ae0dea7 (one commit before the `@tanstack/*` bumps), bug still occurs.

PR #8074, opened 2026-06-07
feat: quotation marks localization (@anakojm)

anakojm:
### Description

this PR introduces localization for quotation marks in most languages. the data is from [wikipedia](https://en.wikipedia.org/wiki/Quotation_mark#Summary_table)

# primary punctuation symbols

## languages that use two or more

- Bosnian: ”…” and „…”
- Danish: »…« and „…“
- Latvian: “…” and „…”
- toki pona: any/none

# secondary punctuation symbols

## languages that use two or more

- Danish: ›…‹ and ‚…‘
- French: «…» and “…”
- Greek: “…” and ‟…”
- Icelandic: ‚…‘ and ‘…’
- Ukrainian: “…” and „…“
- toki pona: any/none

i do not know how to implement these so i did not. one of the two options coud be arbitrarily chosen, and the ability to set custom quotation marks added, but this out of the scope of this PR

# languages not in monkeytype

format:
- language: primary, secondary
<br>

- Basque: «…», “…”
- French Swiss: «…», ‹…›
- Ido: «…», ‘…’
- Interlingua: “…”, ‘…’
- Italian Swiss: «…», ‹…›
- Karakalpak: «…», none
- New Tai Lue: 《…》, ⟨…⟩
- Romansh: «…», ‹…›
- Scottish Gaelic: ‘…’, “…”
- Sorbian: „…”, ‘…’
- Tai Le: 《…》, ⟨…⟩
- Tigrinya: «…», ‹…›
- Uyghur: «…», ‹…›

## language exists, but alternate script does not

- Uzbek (Cyrillic): «…», „…“
- Mongolian (Mongolian script): ⟪…⟫, ⟨…⟩

## language exists, but alternate typographical convention does not

- Korean (North Korea): 《…》, ⟨…⟩
- Portuguese (Portugal): «…», “…”

there exists one language where there are toggle able alternate typographical conventions: English (“british english”). i believe the same thing could be done for Portuguese, Korean, and perhaps with French & French Swiss and Italian & Italian Swiss and others, but this is out of the scope of this PR

# languages in monkeytype, and not in the wikipedia data

currently we fall back to the default ("…", '…') for these, as for “gimmick” languages (leagues of legend, pokemon, typing of the dead…)

note that because of the changes in `frontend/src/ts/utils/strings.ts`, all of these changes are purely visual, and a user can input any quote mark they want, given it corresponds to the correct level of quotation.

- Bangla
- Bashkir
- Bemba
- Euskera
- Frisian
- Friulian
- Greeklish
- Gujarati
- Hausa
- Hawaiian
- Hinglish
- Jyutping
- Kabyle
- Kannada
- Kinyarwanda
- Klingon
- Kokanu
- Kurdish
- Kyrgyz
- Latin
- Likanu
- Lorem
- Malagasy
- Malay
- Malayalam
- Maori
- Marathi
- Myanmar
- Nepali
- Oromo
- Pinyin
- Quenya
- Sanskrit
- Santali
- Shona
- Sinhala
- Swahili
- Tamil
- Tanglish
- Tatar
- Telugu
- Udmurt
- Urdish
- Viossa
- Xhosa
- Yiddish
- Yoruba
- Zulu

note that wikipedia uses the term “Slovene”, instead of the term used by monkeytype “Slovenian”, but this probably doesn’t matter
also note that Hungarian has a third level of quotes: ’…’. Portuguese (Portugal) also has: ‘…’

### Checks


fehmer:
If we want to support this, we should add the quotation marks inside the language files and not add a huge if/else

anakojm:
you’re right, that would be cleaner. i will try to do that. i just kinda took how it was implemented already to the extreme

anakojm:
@fehmer  sorry, i am not familiar with this codebase. how should i access the key—say `"quot_marks"`—of a language file (ex `frontend/static/languages/english.json` i assume). i don’t know how to get this data cleanly from the front end

anakojm:
another possibility that introduces less changes and is more lightweight, is to only keep the changes to [‎frontend/src/ts/utils/strings.ts](https://github.com/monkeytypegame/monkeytype/pull/8074/changes#diff-144dfe8b71d729b1547f34f24195bbe1caf1f716bc142c2384357ed03b8d9d0a). all the other changes are purely cosmetic, and might actually clash in some cases with pre‐existing non‐use of typographical characters (`'` instead of `’` for instance). as such they might need their own PR for a “typography mode”. a compromise between these two position is to conserve `"`/`'` for languages that use `“` & `”`/`‘` & `’`, and only introduce the visual change for languages that use `«` & `»` and other visually very different quotation marks

fehmer:
```diff
--- a/frontend/src/ts/test/words-generator.ts
+++ b/frontend/src/ts/test/words-generator.ts
@@ -42,6 +42,7 @@ export async function punctuateWord(
   currentWord: string,
   index: number,
   maxindex: number,
+  language:LanguageObject,
 ): Promise<string> {
   let word = currentWord;
 
@@ -1132,6 +1133,7 @@ export async function getNextWord(
       randomWord,
       wordIndex,
       wordsBound,
+      currentLanguage
     );
   }
```

you can just pass it in as a new parameter like this.

docs/LANGUAGES.md
```diff
@@ -29,6 +33,13 @@ The contents of the file should be as follows:
 ```
 
 It is recommended that you familiarize yourselves with JSON before adding a language. For the `name` field, put the name of your language.
+`quotationMarks` indicates the quotation marks used for that language, at depth 1 and 2 respectively, and for the left quotation mark and the right one respectevely. For instance, American English uses `“”` at depth 1 and `‘’` at depth 2, in that case it would look like:
```
Leonabcd123:
```suggestion
`quotationMarks` indicates the quotation marks used for that language, at depth 1 and 2 respectively, and for the left quotation mark and the right one respectively. For instance, American English uses `“”` at depth 1 and `‘’` at depth 2, in that case it would look like:
```

Leonabcd123 (changes requested):

docs/LANGUAGES.md
fehmer: Let's use a more descriptive format like
```
quotationMarks: {
 primary:[”...”,"...”],
 secondary...
}
```

anakojm: good idea. implementing it

packages/schemas/src/languages.ts
```diff
@@ -460,10 +460,10 @@ export const LanguageObjectSchema = z
   .object({
     name: LanguageSchema,
     quotationMarks: z
-      .tuple([
-        z.tuple([z.string(), z.string()]),
-        z.tuple([z.string(), z.string()]),
-      ])
+      .object({
```
fehmer, line 465: lets make sure no unknown properties are set by using ` .strict()` on the quotationMarks  schema

fehmer:
hi @anakojm , one last comment, the rest looks fine, thanks

anakojm:
i have no idea why ci-fe is failing. it is not failing on my machine, and only started to fail after i merged changes from master, which themselves don’t fail any checks, and the error message doesn’t make sense to me
edit: well, it fixed itself

anakojm:
i have implemented all the requested changes, hope it is satisfactory

PR #8082, opened 2026-06-10
fix(quote-search): quote length description is incorrect for some languages (@Leonabcd123)

### LINKED ISSUE #8081: Wrong length description for quotes in quote search modal for some languages

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

Selecting the language to be `code systemverilog`, opening the quote search modal and filtering by `length = short`, all of the resulting quotes show `long` in the `length` field.

### Expected Behavior

The quotes should show `short` in the `length` field.

### Steps To Reproduce

1. Change language to `code systemverilog`
2. Switch to `quote` mode
3. Click on the search icon
4. Click on `filter by length`
5. Click on `short`
6. Notice how all quotes have `long` in the `length` field

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 150.0.3

### Anything else?

_No response_

Leonabcd123:
Fixes #8081.


fehmer:
Hi @Leonabcd123,

we have the translation from quote.group to a in multiple places. Should we just add it to the `Quote` type on the frontend as e.g. `groupDescription`and map it once in the controller after reading the json file and check where we can use this instead of the number

packages/schemas/src/quotes.ts
```diff
@@ -9,6 +9,15 @@ export const QuoteIdSchema = z
   .or(z.string().regex(/^\d+$/).transform(Number));
 export type QuoteId = z.infer<typeof QuoteIdSchema>;
 
+export const QuoteLengthSchema = z.union([
```
fehmer: why moving from shared to here?

Leonabcd123: It seems more appropriate to have `QuoteLengthSchema` be in the `quotes` file, and I don't see a reason to have it in the `shared` file (as far as I understand, `shared` is to not have circular dependencies, maybe I misunderstood the use case).

fehmer:
https://github.com/monkeytypegame/monkeytype/blob/6f16874dcad10ccbbc1627081dc2408179d4eac4/frontend/src/ts/test/result.ts#L787 should use the new `groupDescription`as well

fehmer:
https://github.com/monkeytypegame/monkeytype/blob/6f16874dcad10ccbbc1627081dc2408179d4eac4/frontend/src/ts/collections/results.ts#L500 does the reverse mapping, would be great if we use one source for it

fehmer:
https://github.com/monkeytypegame/monkeytype/blob/6f16874dcad10ccbbc1627081dc2408179d4eac4/frontend/src/ts/components/modals/MobileTestConfigModal.tsx#L23

another one

fehmer:
and another :) 
https://github.com/monkeytypegame/monkeytype/blob/6f16874dcad10ccbbc1627081dc2408179d4eac4/frontend/src/ts/components/pages/test/TestConfig.tsx#L306

frontend/src/ts/controllers/quotes-controller.ts
```diff
@@ -28,6 +28,17 @@ const defaultQuoteCollection: QuoteCollection = {
   groups: [],
 };
 
+export const groupToDesc: Record<number, QuoteLength> = {
```
fehmer: what about we move this to schema/quotes to get rid of this duplicate? have one mapping, define the quotelentghschema out of it?

fehmer (dismissed):

suzunn:
I noticed one runtime regression in the current reverse-map shape. `groupToDesc` is built with `Object.entries(descToGroup).map(...)` and then cast to `Record<number, QuoteLength>`, but the cast does not change the runtime value. That leaves `groupToDesc[0]` as a tuple like `[0, short]` instead of the string `short`, so `getLengthDesc()` can assign an array to `groupDescription` and the UI can render `0,short` instead of `short`.

I would change that reverse map to `Object.fromEntries(Object.entries(descToGroup).map(([desc, group]) => [group, desc]))` and add a small assertion around the controller result for a quote in group 0/1. That should keep the schema source of truth while catching this specific runtime/type-cast mismatch.

Leonabcd123:
> I noticed one runtime regression in the current reverse-map shape. `groupToDesc` is built with `Object.entries(descToGroup).map(...)` and then cast to `Record<number, QuoteLength>`, but the cast does not change the runtime value. That leaves `groupToDesc[0]` as a tuple like `[0, short]` instead of the string `short`, so `getLengthDesc()` can assign an array to `groupDescription` and the UI can render `0,short` instead of `short`.
> 
> I would change that reverse map to `Object.fromEntries(Object.entries(descToGroup).map(([desc, group]) => [group, desc]))` and add a small assertion around the controller result for a quote in group 0/1. That should keep the schema source of truth while catching this specific runtime/type-cast mismatch.

Fixed in ed50376bd9159f2874d77cdf9b0c4dd9db26ff5b.

PR #8091, opened 2026-06-12
fix(quote-length-filter): custom length filter modal not opening (@Leonabcd123)

Leonabcd123:
To reproduce:

1. Switch to `quote` mode
2. Click on the search icon
3. Click on `filter by length`
4. Select `custom`
5. Fill out the min and max values and save
6. Remove it by clicking the `x` next to the word `custom`
7. Repeat steps 3-4
8. Notice how it doesn't open the modal anymore

Also removes the notification after saving a custom length filter.


Miodec:
Thanks

PR #8121, opened 2026-06-16
fix(test): keymap not rendering correctly if row5 has two keys (@fehmer)

### LINKED ISSUE #8110: dynamic keymap legend not working with two keys in row5
- select keymap layout `BEAKL Zi` (or any other keymap with two keys in row5
- select keymap legend style `dynamic`
- press shift

Keymap should show shifted layer but is stuck on the unsifted layer

fehmer:
fixes #8110

### DIFF
--- a/frontend/src/ts/elements/keymap.ts
+++ b/frontend/src/ts/elements/keymap.ts
@@ -204,7 +204,7 @@ function buildRow(options: {
 
   if (isMatrix) {
     if (rowId !== "row5" && layoutData.matrixShowRightColumn) {
-      keysHtml += `<div class="keymapKey" data-key=" "></div>`;
+      keysHtml += `<div class="keymapKey"></div>`;
     } else {
       keysHtml += `<div></div>`;
     }
--- a/frontend/src/ts/elements/keymap.ts
+++ b/frontend/src/ts/elements/keymap.ts
@@ -204,7 +204,7 @@ function buildRow(options: {
 
   if (isMatrix) {
     if (rowId !== "row5" && layoutData.matrixShowRightColumn) {
-      keysHtml += `<div class="keymapKey data-key=" "></div>`;
+      keysHtml += `<div class="keymapKey" data-key=" "></div>`;
     } else {
       keysHtml += `<div></div>`;
     }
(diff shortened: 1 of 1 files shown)


frontend/src/ts/elements/keymap.ts
```diff
@@ -204,7 +204,7 @@ function buildRow(options: {
 
   if (isMatrix) {
     if (rowId !== "row5" && layoutData.matrixShowRightColumn) {
-      keysHtml += `<div class="keymapKey"></div>`;
+      keysHtml += `<div class="keymapKey data-key=" "></div>`;
```
Leonabcd123: Looks like there are missing closing quotes for the `class` attribute here.

fehmer: thanks

frontend/src/ts/elements/keymap.ts
```diff
@@ -204,7 +204,7 @@ function buildRow(options: {
 
   if (isMatrix) {
     if (rowId !== "row5" && layoutData.matrixShowRightColumn) {
-      keysHtml += `<div class="keymapKey"></div>`;
+      keysHtml += `<div class="keymapKey" data-key=" "></div>`;
```
Leonabcd123: (applicable to both this line and the second line changed) why did you add a `data-key` attribute with just a space?

fehmer: matrixShowRightColumn seems not to work as expected, i removed the data-key.

for the other line it is the space key

Leonabcd123: Do we need to add `data-key` to the space key? It's skipped anyways in `updateLegends` (which I think is the only place where we read `data-key`), and currently it doesn't have the delimiter all of the other keys have.

Leonabcd123: > matrixShowRightColumn seems not to work as expected, i removed the data-key.

Let's fix this instead of just removing `data-key`. The bug this pr fixes isn't specific to layouts with row5 with 2 keys, layouts with `matrixShowRightColumn` enabled don't work with dynamic keymap style for the same reason (there's no `data-key` attribute). Example layout for this is `swedish_qwerty`.

For that, we can just add `data-key="${keyDataDelimiter}"` to:

https://github.com/monkeytypegame/monkeytype/blob/f870bc9d9c4b2cc09b4b5456f6356ccff72241fe/frontend/src/ts/elements/keymap.ts#L207

Instead of the space.

fehmer: matrixShowRightColumn always has empty row on the far left since it was implemented. This is out of scope for this fix. I started migrating the keymap to solidjs, i'll fix it there.

fehmer:
fixed by #8134

PR #8129, opened 2026-06-17
fix: reset animation fps limit to native when custom field is cleared on blur

### LINKED ISSUE #8119: Animation fps limit setting doesn't reset on blur after being emptied

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

### Expected Behavior

### Steps To Reproduce

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 151.0.4

### Anything else?

_No response_

klfk:
## Description

Fixes #8119

### Problem
When a user set a custom animation fps limit (e.g. `60`), then cleared the input field and clicked away, nothing happened — the `native` button stayed inactive, the field stayed empty, and the previous custom value was still in effect (no visual feedback).

**Root cause:** The `onBlur` validator called `form.handleSubmit()` with an empty string, which failed the `onChange` validator (`"Must be a number"`), so `onSubmit` was never called. The `resetToDefaultIfEmptyOnBlur` prop had no effect because the form's `defaultValues.fpsLimit` was `""` — already empty.

### Fix
In the `onBlur` validator, explicitly detect an empty value and call `setfpsLimit(1000)` to reset to native, making the **native** button active and clearing the indicator — which matches the expected behavior described in the issue.

```tsx
// Before
onBlur: () => {
  void form.handleSubmit();
},

// After
onBlur: ({ value }) => {
  if (String(value) === "") {
    setfpsLimit(1000);
    form.setFieldValue("fpsLimit", "");
    savedIndicator.hide();
    return;
  }
  void form.handleSubmit();
},
```

Also removed the now-redundant `resetToDefaultIfEmptyOnBlur` prop from `<InputField>` since blur is handled manually.

### Testing
1. Go to Settings → search **animation fps limit**
2. Enter `60` in the custom limit field → click away → ✅ saved indicator flashes, native button inactive
3. Clear the field → click away → ✅ **native** button becomes active, field stays empty

### DIFF
--- a/frontend/src/ts/components/pages/settings/custom-setting/AnimationFpsLimit.tsx
+++ b/frontend/src/ts/components/pages/settings/custom-setting/AnimationFpsLimit.tsx
@@ -62,7 +62,13 @@ export function AnimationFpsLimit(): JSXElement {
                     value: val,
                   });
                 },
-                onBlur: () => {
+                onBlur: ({ value }) => {
+                  if (String(value) === "") {
+                    setfpsLimit(1000);
+                    form.setFieldValue("fpsLimit", "");
+                    savedIndicator.hide();
+                    return;
+                  }
                   void form.handleSubmit();
                 },
               }}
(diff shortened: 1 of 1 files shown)


fehmer:
If the stored value was  `60` shouldn't it reset to  `60` on empty input.
In addition the field should not be marked with an error if the user clicks on native:
image

fehmer:
duplicate #8130

PR #8130, opened 2026-06-17
fix(settings): prevent fps input from resetting (@Rajparikh18)

### LINKED ISSUE #8119: Animation fps limit setting doesn't reset on blur after being emptied

Rajparikh18:
### Description
 Fixes: #8119 
 
Problem

The FPS input was resetting, even after giving input to it and saving it, once the page is refreshed it shows empty field there again.

Cause

The FPS input always shows empty field , just because the by default value of the fpsLimit is an empty string ( "")

Fix

Updated the FPS input handling by adding String(getfpsLimit()) this instead of "" so that valid user-entered values are preserved instead of being reset. The input now correctly reflects the saved FPS value after interacting with the field.

`
defaultValues: {
      fpsLimit: String(getfpsLimit()),
    },
`

Testing
Open Settings and search for "animation fps limit".
Enter a custom FPS value.
Click away from the input.
Once it is saved,
Refresh the page and verify that the saved value is still displayed correctly.

EDIT: addressed the issue , added a conditional statement, so that the input will not show 1000 and added a if statement to remove the "must be a number" error when selected native


fehmer:
Hi @Rajparikh18 , thank you for your contribution.

 Reloading the settings page with fps limit set to native shows

image

input field should be empty.

when setting custom value and pressing "native" an error is show but shouldnt:

image

fehmer (changes requested):
see comment

fehmer (approved):

Miodec (approved):
Thanks

PR #8153, opened 2026-06-23
feat: add regex word filter (@benhuang3)

benhuang3:
### Description
This PR adds a regex filter to left column of words filter page, allowing users to filter words using regular expressions along with the other rules in the words filter tab. 
- Like "exclude", box is cleared and expressions ignored when "exact match" is true. 
- Parses string into regex using https://stackoverflow.com/a/874742 
- Shows notification on invalid regex expressions and other errors
- Layout may not be suitable, please give suggestions!

### Checks

Screenshot 2026-06-23 at 3 33 44 PM


Miodec:
Thanks

PR #8157, opened 2026-06-24
refactor(generator): include commit spaces in generated words (@miodec, @Leonabcd123)

Miodec:
Reworks the generation to include commit/separation character in the target word.


frontend/src/ts/elements/caret.ts
```diff
@@ -287,7 +286,13 @@ export class Caret {
...
       );
-      const wordText = TestWords.words.getText(options.wordIndex) ?? "";
+
+      const wordText =
+        word
+          ?.qsa("letter")
+          .map((l) => l.native.textContent ?? "")
+          .join("") ?? "";
+
       const wordLength = Array.from(wordText).length;
```
Miodec, line 291: Not sure which approach is better yet

PR #8177, opened 2026-06-29
style(quotes): add english quotes (@kingbahrani)

kingbahrani:
### i just added 1 english quote from metallica - to live is to die 

"When a man lies, he murders some part of the world.
These are the pale deaths which men miscall their lives.
All this I cannot bear to witness any longer.
Cannot the kingdom of salvation take me home?"

— [Cliff Burton](https://en.wikipedia.org/wiki/Cliff_Burton), "To Live Is To Die"

### DIFF
--- a/frontend/static/quotes/english.json
+++ b/frontend/static/quotes/english.json
@@ -39345,6 +39345,12 @@
       "source": "Doctor Who, Fallen Heroes",
       "id": 7770,
       "length": 485
+    },
+    {
+      "text": "When a man lies, he murders some part of the world. These are the pale deaths which men miscall their lives. All this I cannot bear to witness any longer. Cannot the kingdom of salvation take me home?",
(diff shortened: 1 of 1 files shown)


frontend/static/quotes/english.json
```diff
@@ -39345,6 +39345,12 @@
       "source": "Doctor Who, Fallen Heroes",
       "id": 7770,
       "length": 485
+    },
+    {
+      "text": "When a man lies, he murders some part of the world. These are the pale deaths which men miscall their lives. All this I cannot bear to witness any longer. Cannot the kingdom of salvation take me home?",
+      "source": " Cliff Burton, To Live Is To Die",
```
Leonabcd123: Why is the source prefixed with a space?

kingbahrani: Fixed, thanks for the heads-up

PR #8186, opened 2026-07-03
fix: use lenient validation for existing passwords (@fehmer)

fehmer:


frontend/src/ts/auth.tsx
```diff
@@ -578,8 +578,20 @@ export function isUsingAuthenticationReactive(authMethod: AuthMethod): boolean {
...
+/**
+ * Returns the Zod schema for password validation.
+ *
+ * Set `isNew: true` for registration/creation flows (strict rules).
+ * Omit it for re-authentication flows (lenient: just non-empty).
+ *
+ * @param options - Set `isNew: true` for password creation/registration.
+ * @returns A Zod string schema.
+ */
+export function getPasswordSchema(options?: { isNew: boolean }): ZodString {
```
Miodec, line 590: Maybe move the isdevenv guard to the top?

`if (isDevEnv) return z.string().min6`

Miodec (changes requested):

PR #8191, opened 2026-07-03
fix: fix caret error on incorrect final word (@Positiveoo1)

### LINKED ISSUE #8180: no letters found in word error when committing an incorrect last word with a space

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

When typing all the letters in the last word of a test such that the word is incorrect and pressing space, the following error will appear in the console:

```
Uncaught Error: Caret getTargetPositionAndWidth: no letters found in word
    getTargetPositionAndWidth caret.ts:401
    goTo caret.ts:323
    n debounced-animation-frame.ts:10
    ji debounced-animation-frame.ts:8
    goTo caret.ts:286
    hd caret.ts:34
    Vb test-ui.ts:1782
    Hb test-ui.ts:1803
    JA insert-text.ts:327
    async* input.ts:115
```

### Expected Behavior

No error should be thrown.

### Steps To Reproduce

1. Change mode to `custom`
2. Enter `a a` as the custom text
3. Type `a` + ` ` + `f` + ` ` (second and last chars are spaces)
4. Open the console
5. Notice the error

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 151.0.4

### Anything else?

I haven't noticed any problems as a result of this error, but it's still worth looking into why it happens.

Positiveoo1:
Fixes #8180

## Summary
- Render incorrect committed words before word change UI updates
- Prevent caret positioning from throwing when target word has no rendered letters yet

## Testing
- Custom text: `a a`
- Typed `a`, space, `f`, space
- Confirmed console no longer throws `Caret getTargetPositionAndWidth: no letters found in word`

### DIFF
--- a/frontend/src/ts/elements/caret.ts
+++ b/frontend/src/ts/elements/caret.ts
@@ -395,13 +395,19 @@ export class Caret {
     isDirectionReversed: boolean;
   }): { left: number; top: number; width: number } {
     const letters = options.word?.qsa("letter");
-    let letter = letters?.[options.letterIndex] ?? letters?.at(-1);
-    if (letter === undefined) {
-      return {
-        left: options.word.getOffsetLeft(),
-        top: options.word.getOffsetTop(),
-        width: getTotalInlineMargin(options.word.native),
-      };
+
+    if (letters.length === 0) {
+      throw new Error(
+        "Caret getTargetPositionAndWidth: no letters found in word",
+      );
+    }
+
+    let letter = letters[options.letterIndex] ?? letters[letters.length - 1];
+
+    if (!letter) {
+      throw new Error(
+        `Caret getTargetPositionAndWidth: letter not found for index ${options.letterIndex}`,
+      );
     }
 
     if (caretDebug) {
--- a/frontend/src/ts/elements/caret.ts
+++ b/frontend/src/ts/elements/caret.ts
@@ -395,12 +395,13 @@ export class Caret {
     isDirectionReversed: boolean;
   }): { left: number; top: number; width: number } {
     const letters = options.word?.qsa("letter");
-    let letter;
-    if (!letters?.length || !(letter = letters[options.letterIndex])) {
-      // maybe we should return null here instead of throwing
-      throw new Error(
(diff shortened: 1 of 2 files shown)


frontend/src/ts/elements/caret.ts
```diff
@@ -395,12 +395,13 @@ export class Caret {
...
-      throw new Error(
-        "Caret getTargetPositionAndWidth: no letters found in word",
-      );
+    let letter = letters?.[options.letterIndex] ?? letters?.at(-1);
+    if (letter === undefined) {
+      return {
+        left: options.word.getOffsetLeft(),
+        top: options.word.getOffsetTop(),
+        width: getTotalInlineMargin(options.word.native),
+      };
```
Miodec: I think i would change this to something like
```ts
const letters = options.word?.qsa("letter");

if (letters.length === 0) {
  throw new Error(
    "Caret getTargetPositionAndWidth: no letters found in word",
  );
}

const letter = letters[options.letterIndex] ?? letters[letters.length - 1];

if (!letter) {
  throw new Error(
    `Caret getTargetPositionAndWidth: letter not found for index ${options.letterIndex}`,
  );
}

if (caretDebug) {
```

frontend/src/ts/test/test-ui.ts
```diff
@@ -1841,6 +1841,7 @@ export function beforeTestWordChange(
   if (
     (Config.stopOnError === "letter" && (correct || correct === null)) ||
     nospaceEnabled ||
+    correct === false ||
```
Miodec: And revert this. This code will be changed soon anyway.

Miodec (changes requested):

frontend/src/ts/elements/caret.ts
Positiveoo1: Updated, thanks. Reverted the test-ui change and changed caret lookup to fall back only when the requested index is missing.

PR #8198, opened 2026-07-05
feat: add finger training practice mode (@OmTiwariPandey)

OmTiwariPandey:
### Description

I was practicing on my account (uname: itsom) and figured I was not hitting my left
pinky as much (words containing q and z were simply not appearing), so I
constantly felt like I wasn't training that finger. I'm sure others suffer
from the same, so I built a finger training mode: a modal (reachable from the
commandline and from a card on the settings page) where you pick the fingers
you want to train and optionally how often a training word appears (every
word, every 2nd, every 3rd).

- Words are picked in per-letter buckets from the current language, so rare
  letters (like q and z for left pinky) get an equal share instead of being
  crowded out by common ones (like a). Short letter drills fill in when the language
  has few or no words for a letter.
- Letters per finger come from the active layout via the existing
  layoutKeyToKeycode converter.
- Training survives test restarts; stopping it, changing the mode, or
  reloading the page all restore your previous settings.
- A notice above the test shows which fingers are being trained.
- Added 20 unit tests (finger mapping incl. iso/matrix, scoring, pool
  building, drills, frequency mixing).
- Known limitation: angle-mod iso layouts get standard physical-column
  fingering, since layout JSONs can't express angle-mod intent.
### Checks

Closes #


OmTiwariPandey:
I'll solve the conflicts but @fehmer @Miodec what do you guys think of this?

Miodec:
Seems to me like it essentially does what words filter already does?

OmTiwariPandey:
eh, well, kinda. I suggest you still take a look at the PR locally. It does what sorta can be achieved via word filter, however, I see it less as a replacement for word filter and more as a guided training layer on top of the current language/layout. The user chooses “train left pinky” instead of manually building and maintaining a q/z/a/etc word filter. Thought it would be cool to have.

OmTiwariPandey:
Some more things I noticed: the custom word filter only generates one set at a time and restarting the test doesn't change the word set. not with the finger training thing. plus the users will have better ease of use. What do you think of adding like a beta program, to test features like this? cuz I think users would be a better judge than people who code. well obviously we use the product too but yk, the scale would be larger. Sorry for the pings but what do you think? @Miodec @fehmer?

PR #8202, opened 2026-07-06
fix(font): Add Sindhi-specific CSS for font clarity (@Shamsuddin45)

Shamsuddin45:
Add language-specific CSS for Sindhi language font clarity.

### Checks

- [ ] Adding quotes?
  - Make sure to follow the [quotes documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/QUOTES.md)
  - [ ] Make sure to include translations for the quotes in the description (or another comment) so we can verify their content.
- [ ] Adding a language?
  - Make sure to follow the [languages documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/LANGUAGES.md)
  - [ ] Add language to `packages/schemas/src/languages.ts`
  - [ ] Add language to exactly one group in `frontend/src/ts/constants/languages.ts`
  - [ ] Add language json file to `frontend/static/languages`
- [ ] Adding a theme?
  - Make sure to follow the [themes documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/THEMES.md)
  - [ ] Add theme to `packages/schemas/src/themes.ts`
  - [ ] Add theme to `frontend/src/ts/constants/themes.ts`
  - [ ] (optional) Add theme css file to `frontend/static/themes`
  - [ ] Add some screenshots of the theme, especially with different test settings (colorful, flip colors) to your pull request
- [ ] Adding a layout?
  - [ ] Make sure to follow the [layouts documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/LAYOUTS.md)
  - [ ] Add layout to `packages/schemas/src/layouts.ts`
  - [ ] Add layout json file to `frontend/static/layouts`
- [ ] Adding a font?
  - Make sure to follow the [fonts documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/FONTS.md)
  - [ ] Add font file to `frontend/static/webfonts`
  - [ ] Add font to `packages/schemas/src/fonts.ts`
  - [ ] Add font to `frontend/src/ts/constants/fonts.ts`
- [ ] Check if any open issues are related to this PR; if so, be sure to tag them below.
- [ ] Make sure the PR title follows the Conventional Commits standard. (https://www.conventionalcommits.org for more info)
- [ ] Make sure to include your GitHub username prefixed with @ inside parentheses at the end of the PR title.

Closes #

PR #8208, opened 2026-07-08
feat: allow languages to set specific fonts (@Shamsuddin45, @fehmer)

fehmer:


fehmer:
#8202

docs/LANGUAGES.md
```diff
@@ -24,6 +24,7 @@ The contents of the file should be as follows:
   "joiningScript": boolean,
   "orderedByFrequency": boolean,
   "bcp47": string,
+  "fallbackFont":string,
```
nadalaba: I think `"preferredFont"` is a better name, since it's used at the top of the list

fehmer: it is second in the list, so the first fallback font. Maybe we just name it `font`?!

PR #8213, opened 2026-07-09
fix(settings): animation fps limit input not resetting on blur (@codewithyuvraj24)

### LINKED ISSUE #8119: Animation fps limit setting doesn't reset on blur after being emptied

codewithyuvraj24:
Closes #8119

## Problem

The animation fps limit custom input field doesn't reset to the previously saved value when the user clears it and clicks outside (blur). Other settings like `max line width` handle this correctly.

## Root Cause

The form's `defaultValues.fpsLimit` was hardcoded to `""` (empty string). When `resetToDefaultIfEmptyOnBlur` fires in `InputField`, it resets to this empty default — effectively doing nothing.

In contrast, `MaxLineWidth` uses the actual saved config value as its default, so the reset works correctly.

## Fix

Changed the default value from a hardcoded empty string to the actual saved fps limit value:

```diff
 defaultValues: {
-  fpsLimit: "",
+  fpsLimit: getfpsLimit() === 1000 ? "" : String(getfpsLimit()),
 },
```

- When set to native (1000), keeps the empty string default (native button is used instead)
- When set to a custom value, uses the saved value so `resetToDefaultIfEmptyOnBlur` correctly restores it


fehmer:
hi @codewithyuvraj24 , thanks for your contribution. There is already a PR for this #8130

PR #8273, opened 2026-07-24
fix(friends page): Fixed a loader position which is position next to heading "Friends" (@self-yash)

self-yash:
### Description

Problem:
- The circular loader on the Friends page was positioned too close and not exactly centered to the "Friends" heading, causing it to touch the last letter.

Solution:
- Adjusted the loader's positioning to provide proper spacing and improve the visual alignment next to the heading.

Scope:
- UI-only change. No functional behavior was modified.

Attached an image of the issue:
image


Miodec (approved):
Thanks

PR #8274, opened 2026-07-24
fix(quotes): prevent wrong quote length when moderator edits quote (@HmonWutt)

HmonWutt:
### Description

When a moderator edits a quote, the length of quote was calculated from the original (unedited) text instead of the new (edited) quote text.


Miodec (approved):
Thanks

PR #8284, opened 2026-07-26
impr(quotes): add an english quote and fix a workflow document issue (@RedPandaGuy1234)

RedPandaGuy1234:
I added a new quote to the quote list. This quote was from Harry Potter and The Deathly Hallows. Also, while having the checks run, I realized that there was an issue with one of the workflow documents, so I fixed that too.

PR #8285, opened 2026-07-26
fix(keymap): restore handling of shift/capslock (@fehmer)

### LINKED ISSUE #8282: Keymap still shows uppercased letters when both Caps Lock and shift are enabled

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

When `keymap legend style` is set to `dynamic`, Caps Lock is enabled and shift is held, the keymap still shows uppercased letters, even though the actual output is lowercased.

### Expected Behavior

Letters should be lowercased.

### Steps To Reproduce

1. Set `keymap legend style` to `dynamic`
2. Enable Caps Lock
3. Hold Shift
4. Notice how the letters in the keymap are still uppercased

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 151.0.4

### Anything else?

_No response_

### LINKED ISSUE #8283: Keymap shows wrong layer for symbols when Caps Lock is enabled

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

Setting `keymap keys` to `full`, `keymap legend style` to `dynamic` and enabling Caps Lock shows the shifted layer for the top row (e.g., `1` turns to `!`), even though pressing that key still outputs the symbol in the default layer.

### Expected Behavior

Symbols should not change, unless shift is held.

### Steps To Reproduce

1. Set `keymap keys` to `full`
2. Set `keymap legend style` to `dynamic`
3. Enable Caps Lock
4. Notice how the top row changed to the shifted layer

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 151.0.4

### Anything else?

_No response_

fehmer:
fixes #8283
fixes #8282

### DIFF
--- a/frontend/src/ts/components/pages/test/Keymap.tsx
+++ b/frontend/src/ts/components/pages/test/Keymap.tsx
@@ -35,25 +35,26 @@ export function Keymap() {
 function Keyboard(props: { displayName: string; layoutData: LayoutObject }) {
   const layer = createMemo(() => {
     const { alt, shift } = getModifierState();
+    const isShifted = isMacLike()
+      ? shift || isCapsLockOn()
+      : shift !== isCapsLockOn();
     switch (getConfig.keymapLegendStyle) {
       case "blank":
-        return -1;
+        return { index: -1 };
       case "lowercase":
-        return 0;
+        return { index: 0 };
       case "uppercase":
-        return 1;
+        return { index: 1 };
       case "dynamic": {
         if (shift && alt) {
-          return 3;
+          return { index: 3 };
         } else if (alt) {
-          return 2;
-        } else if (shift || isCapsLockOn()) {
-          return 1;
+          return { index: 2 };
         }
-        return 0;
+        return { index: isShifted ? 1 : 0, firstRow: shift ? 1 : 0 };
       }
       default:
-        return 0;
+        return { index: 0 };
     }
   });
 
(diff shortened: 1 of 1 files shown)


fehmer:
@Leonabcd123 can you give it a test please?

RedPandaGuy1234:
> @Leonabcd123 can you give it a test please?

Hi, I'm not Leonabcd123, but the reason why there is a failing check is because in the github workflow document, there is an error, which I fixed in #8284. Once it gets approved, the check will be successful.

frontend/src/ts/components/pages/test/Keymap.tsx
```diff
@@ -35,25 +35,26 @@ export function Keymap() {
 function Keyboard(props: { displayName: string; layoutData: LayoutObject }) {
   const layer = createMemo(() => {
     const { alt, shift } = getModifierState();
+    const isShifted = isMacLike()
```
Leonabcd123, line 46: What do you think about adding back the comment we had?

```
// MacOS has different CapsLock and Shift logic than other operating systems
// Windows and Linux only capitalize letters if either Shift OR CapsLock are
// pressed, but not both at once.
// MacOS instead capitalizes when either or both are pressed,
// so we have to check for that.
```

Leonabcd123 (commented):
This fixes #8282, but still doesn't fully fix #8283, because we still show the shifted layer for symbols outside the first row (say the comma key) when Caps Lock is enabled.

frontend/src/ts/components/pages/test/Keymap.tsx
fehmer: sure

fehmer:
@Leonabcd123  what about now?

Leonabcd123 (commented):
Both bugs seem to be gone, nice job.

PR #8300, opened 2026-07-28
ci: attach SBOM attestations to the published images (@kobihikri)

kobihikri:
add SBOM to published docker images.

Hi, and thanks for Monkeytype.

`.github/workflows/publish-docker-images.yml` builds and pushes two images — the frontend and the backend — but neither pushed manifest carries a provenance or SBOM attestation. Someone pulling either cannot check that it was built by this workflow, from this repository, at that tag.

Monkeytype is self-hosted by a fair number of people who want their own typing history rather than an account somewhere, and the images are what they run. There is nothing between the registry and the deployment that could vouch for them today.

The change is two lines on each of the two build steps:

```yaml
          push: true
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the images. **No permissions change is needed** — the workflow's `contents: read` stays exactly as it is, and nothing has to gain `id-token`.

Consumers can then check either image with:

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats worth stating: `mode=max` records the full build including build arguments, so `provenance: true` is the smaller option if any have ever been sensitive; and attestations add an extra manifest to the index, which the registry handles but is worth knowing if a mirror sits in front.

No SLSA level claimed — the attestation is what BuildKit produces.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and both build steps myself.


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

king452323:
M y typing speed is average is 40 as per minute how is can increase it

PR #8301, opened 2026-07-29
fix(quote): correcting source of quote (@Technet8394)

Technet8394:
### Description

Fixed an attribution in a Quote not being specific, "The End" could have multiple different sources
### Checks

Not applicable

Closes #

None


Technet8394:
updated branch in fork

PR #8322, opened 2026-08-04
impr: support more filetypes for text import (@fehmer)

fehmer:
fixes #8316

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -64,6 +64,7 @@
     "lz-ts": "1.1.2",
     "modern-screenshot": "4.6.8",
     "object-hash": "3.0.0",
+    "officeparser": "7.5.1",
     "slim-select": "2.9.2",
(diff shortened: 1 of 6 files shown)


frontend/vite.config.ts
```diff
@@ -103,9 +102,6 @@ function getPlugins({
     tailwindcss(),
 
     solidPlugin(),
-    devtools({
```
fehmer, line 106: devtools breaks officeparser on load, figure out why

frontend/package.json
```diff
@@ -64,6 +64,7 @@
     "lz-ts": "1.1.2",
     "modern-screenshot": "4.6.8",
     "object-hash": "3.0.0",
+    "officeparser": "7.5.1",
```
fehmer, line 67: adds a cool 840kb zipped :(

PR #8324, opened 2026-08-06
fix(activity): keep tooltips within viewport (@ahamSel)

### LINKED ISSUE #8240: Test activity issue

- [x] I have cleared my cache

No

No

No

No

### Account name

Aurelix

### Account config

_No response_

### Current Behavior

You need to shift the test asset slightly to the left; the reason is that the date isn't fully visible when hovering over the active day.

### Expected Behavior

It would be better if the date were positioned slightly to the left when clicking on the active day.

### Steps To Reproduce

1. Tap on an active profile.
2. Hover the cursor over its latest weekly or daily asset.
3. Then you can see the bug.

### Environment

- OS:
- Browser:
- Browser Version:

### Anything else?

_No response_

ahamSel:
### Description

Activity-calendar tooltips could be clipped when their trigger day was close to either viewport edge, particularly on small screens.

This update measures each tooltip when it is shown and shifts only the tooltip box far enough to keep it inside an 8 px viewport margin. The arrow stays centered over the hovered day. Tooltip text also scales modestly on narrow viewports and wraps when needed.

Only opacity and transform animate, preventing responsive tooltip sizing from starting animations for every calendar day during resize.

Closes #8240


fehmer:
hi @ahamSel , thank you for your contribution.

The problem still exists for tooltips close to the edge on small screens on both sides:

image
image

ahamSel:
Thanks for pointing this out. I replaced the final-week alignment with viewport-edge detection on both sides. The tooltip box now shifts inward only when needed while its arrow stays centered over the hovered day; the text also scales modestly and wraps on narrow screens. I verified the behavior manually at small viewport widths.

Miodec:
https://github.com/user-attachments/assets/ec99af22-4685-42b2-8250-e4b635f025ba

Something expensive is running on resize because its very slow.

ahamSel:
Thanks for catching this. The tooltip library's `transition: all` was animating the responsive font size, padding, and max-width on every day, even when its tooltip was hidden. Resizing the profile triggered 2,184 of those transitions across 364 days.

I limited the tooltip transitions to opacity and transform. The sizing transitions are now gone, while the tooltip fade, edge positioning, and centered arrow remain. I checked the left, middle, and right tooltips at narrow and wide viewport widths and ran the existing calendar tests (29 passed), formatting, and lint checks. Also rebased onto current master.

I also tested by dragging the browser window narrower and wider, and resizing now feels smooth.

PR #8328, opened 2026-08-07
fix(settings): clear animation fps limit field on blur when invalid (@karnan7)

### LINKED ISSUE #8326: Animation fps limit setting doesn't reset on blur after entering an invalid value

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

_No response_

### Account config

_No response_

### Current Behavior

When the animation fps limit setting it set to `native`, an invalid value is entered in the field below, and the field is blurred, the invalid value remains and isn't emptied.

### Expected Behavior

Field should empty on blur if an invalid value has been entered and the current value is `native`.

### Steps To Reproduce

2. Click on `native`
3. In the field below it, enter `asdf`
4. Click outside of the field
5. Notice how `asdf` remains in the field

### Environment

- OS: Arch Linux
- Browser: Firefox
- Browser Version: 151.0.4

### Anything else?

A very brief summary of my debugging attempt:

When the attribute list of an element contains spreads:

https://github.com/monkeytypegame/monkeytype/blob/d7eb4b76f3b3000199022ea52a52365b9346b8d0/frontend/src/ts/components/ui/form/InputField.tsx#L107-L108

[solidjs compiles the code differently](https://playground.solidjs.com/anonymous/6f5e94b8-7d30-4720-8b01-a2ff0025de32), causing it to use the `spread` function, rather than using [direct assignment via `effect`](https://playground.solidjs.com/anonymous/7f931690-f65a-4639-929f-840ba38f3034). When using `spread`, it will only write to the dom (`element.value = ...`) if the return value of `convertValueToString` changes, rather than writing whenever `props.field().state.value` changes. In the case of our animation fps limit setting, typing `asdf` will set `props.field().state.value` to undefined (which gets converted to an empty string by `convertValueToString`:

https://github.com/monkeytypegame/monkeytype/blob/d7eb4b76f3b3000199022ea52a52365b9346b8d0/frontend/src/ts/components/ui/form/InputField.tsx#L161

And when we reset on blur, we reset to the default value which is also an empty string. The value hasn't changed, so solid doesn't write to the dom and as a result the invalid value isn't removed.

karnan7:
### Description

When you type letters into the animation fps limit input in Firefox, they stay in the box after you click away, as if they had been accepted. They have not been. Nothing stored or validated them, the text is just left behind on screen.

Firefox does not pass the typed letters to our code. For a number input with bad text it reports the value as empty, so the code thinks the field is empty. On blur the handler resets the field to its default, which for this setting is also empty when `native` is selected. The value did not change, so Solid skips writing it to the input and the letters remain visible.

On blur we now write the field's current value into the input directly, rather than depending on Solid noticing a change.

Tested by hand in Firefox.
Note on tests: jsdom sanitizes the value the same way Firefox does, so this case cannot be reproduced in a unit test. Verified manually instead.

Closes #8326

### DIFF
--- a/frontend/src/ts/components/ui/form/InputField.tsx
+++ b/frontend/src/ts/components/ui/form/InputField.tsx
@@ -70,7 +70,7 @@ export function InputField(props: {
         autocomplete={props.autocomplete}
         name={props.field().name as string}
         value={convertValueToString(props.field().state.value)}
-        onBlur={() => {
+        onBlur={(e) => {
           if (
             props.resetToDefaultIfEmptyOnBlur &&
             (props.field().state.value === undefined ||
@@ -83,6 +83,9 @@ export function InputField(props: {
           }
           shakeItIfYouWantIt();
           props.field().handleBlur();
+          e.currentTarget.value = convertValueToString(
+            props.field().state.value,
+          );
         }}
         onInput={(e) => {
           const value: unknown = convertStringToValue(


Miodec:
I think a cleaner fix here would be to add a new `resetToDefaultIfInvalidOnBlur` param to the `InputField` that works simmilarly to `resetToDefaultIfEmptyOnBlur`. Then in the blur handler, if that prop is true and state is invalid, reset the value to default. Then ofcourse use the prop for the fps animation limit element.

karnan7:
I gave this a go, but I don't think it actually fixes the bug. Sharing what I found in case it helps.

The reset already happens today. When you type letters into a number input, Firefox won't hand us the text at all, it just tells us the field is empty. So the value becomes undefined, which the existing resetToDefaultIfEmptyOnBlur check already catches, and the default already gets set. The state is correct both before and after.

The problem is that the box on screen never gets updated. Because of the spread props, Solid compares the old string with the new one, sees "" both times, and skips writing to the input. I logged the writes to be sure: with the new prop, blur still does zero writes.

That's why I went with forcing the write on blur. It's the one moment where the user is done typing and it's safe to touch the box.

PR #6527, opened 2025-05-05
refactor:  split key labels in layout files (@fehmer)

fehmer:
wip

PR #8345, opened 2026-08-15
fix(backwards): backwards funbox breaks emojis (@Leonabcd123)

Leonabcd123:
To reproduce:

1. Change mode to `custom`
2. Set the custom text to `🏳️‍🌈`
3. Enable the `backwards` funbox
4. Notice how the character breaks

### DIFF
--- a/frontend/src/ts/test/funbox/funbox-functions.ts
+++ b/frontend/src/ts/test/funbox/funbox-functions.ts
@@ -348,6 +348,8 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
   },
   backwards: {
     alterText(word: string): string {
+      if (!("Segmenter" in Intl)) return [...word].reverse().join("");
+
       const segmenter = new Intl.Segmenter();
       const graphemes = [...segmenter.segment(word)].map((s) => s.segment);
       const emojiRegex = /\p{Emoji}/u;
--- a/frontend/src/ts/test/funbox/funbox-functions.ts
+++ b/frontend/src/ts/test/funbox/funbox-functions.ts
@@ -351,12 +351,13 @@ const list: Partial<Record<FunboxName, FunboxFunctions>> = {
       const segmenter = new Intl.Segmenter();
       const graphemes = [...segmenter.segment(word)].map((s) => s.segment);
 
+      const emojiRegex = /\p{Emoji}/u;
       const emojiComponentRegex = /\p{Emoji_Component}/u;
 
       const units = [];
       for (const g of graphemes) {
         // Handles emojis that have ZWJs and regional indicators
-        if (emojiComponentRegex.test(g)) {
+        if (emojiRegex.test(g) && emojiComponentRegex.test(g)) {
           units.push(g);
           continue;
         }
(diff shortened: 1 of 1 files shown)


nadalaba:
I believe `Intl.Segmenter` uses grapheme splitting, which will not work correctly for text `"\u0628\u064B"` (should be split into 2 characters).

I remember `lett arr = []; for (const c of str) arr.push(c)` doing a better job at splitting.

Leonabcd123:
> I believe `Intl.Segmenter` uses grapheme splitting, which will not work correctly for text `"\u0628\u064B"` (should be split into 2 characters).
> 
> I remember `lett arr = []; for (const c of str) arr.push(c)` doing a better job at splitting.

You're correct that we're splitting by grapheme clusters, however I don't understand why `\u0628\u064B` should be split into 2 characters? Isn't it just a base character and a diacritic? With `Intl.Segmenter` it looks exactly the same with `backwards` or without, but with the loop approach it looks different with `backwards` (the diacritic is split from the base character).

nadalaba:
> Isn't it just a base character and a diacritic?

Yes, but there's no 1 key that can produce this combination in 1 go, so without backwards we should type `"\u0628"` first, and with backwards we should type `"\u064B"` first. However, with `Intl.Segmenter` we will be forced to type it `\u0628\u064B` regardless of funbox.

> with the loop approach it looks different with backwards (the diacritic is split from the base character).

Try setting the language to Arabic first

p.s., for a list of key combinations that will cause issues with grapheme splitting see [this](https://github.com/monkeytypegame/monkeytype/pull/6527#issuecomment-2859232467)

p.s.2, `lett arr = []; for (const c of str) arr.push(c)` can be shortened into `[...str]`

Leonabcd123:
> > Isn't it just a base character and a diacritic?
> 
> Yes, but there's no 1 key that can produce this combination in 1 go, so without backwards we should type `"\u0628"` first, and with backwards we should type `"\u064B"` first. However, with `Intl.Segmenter` we will be forced to type it `\u0628\u064B` regardless of funbox.

I see the problem now.

> p.s.2, `lett arr = []; for (const c of str) arr.push(c)` can be shortened into `[...str]`

The problem with this approach is that it breaks surrogate pairs (although I still made it the fallback for when `Intl.Segmenter` doesn't exist, not sure of a better alternative).

nadalaba:
> The problem with this approach is that it breaks surrogate pairs

`[...str]` does not break surrogate pairs, it splits any string into its Unicode characters (works ok with all planes of Unicode):
```js
"𐑩".split("") // splits into 2 surrogate pairs.
[..."𐑩"] // splits into 1 Unicode character.
```
Your reproduction example uses emoji ZWJ sequences with variation selectors which are in essence 4 Unicode characters. I'd prioritize Unicode chars integrity over combining emoji's any day tbh

Leonabcd123:
> > The problem with this approach is that it breaks surrogate pairs
> 
> `[...str]` does not break surrogate pairs, it splits any string into its Unicode characters (works ok with all planes of Unicode):
> 
> ```js
> "𐑩".split("") // splits into 2 surrogate pairs.
> [..."𐑩"] // splits into 1 Unicode character.
> ```
> 

Ah sorry I confused .split("") (which breaks surrogate pairs) with the spread.

> Your reproduction example uses emoji ZWJ sequences with variation selectors which are in essence 4 Unicode characters. I'd prioritize Unicode chars integrity over combining emoji's any day tbh

Yea, that's why I chose that emoji as the reproduction, if I hadn't known about it I would've gone with the spread approach. I'd like to try and find another solution before going with the spread.

PR #8348, opened 2026-08-17
fix: handle live stats correctly during test restart

pankajbindra91:
##Description

Fixes live stats handling during test restarts by ensuring the timer logic respects the restart state before updating the displayed value.

##Changes
- Added restart-state handling to the live timer logic.
- Prevented stale state from affecting live stats after a restart.
- Preserved existing time-limited test behavior.
- Verified the fix with custom text tests.

## Testing
- Ran the type-aware Ox lint check.
- Result: **0 warnings, 0 errors. **
- Tested the change locally with test restarts.

### DIFF
--- a/frontend/src/ts/states/live-stats.ts
+++ b/frontend/src/ts/states/live-stats.ts
@@ -13,6 +13,7 @@ import {
   getFocus,
   isResultCalculating,
   isTestActive,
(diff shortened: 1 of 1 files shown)


fehmer:
Duplicate #8341

PR #8358, opened 2026-08-24
fix: reset mode 2 when unchecking mode (@AzureNightlock)

### LINKED ISSUE #8357: Mode2 value isn't reset when Mode is unchecked

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

azurenightlock

### Account config

_No response_

### Current Behavior

When you click on share test settings and you select the Mode option and the Mode 2 option. Then unselect the Mode option, Mode 2 option remains selected.

### Expected Behavior

When Mode is unselected, Mode 2 should also reset since if the test is not being shared with time selected, then the custom time duration for the said test should also be unselected

### Steps To Reproduce

1. Go to https://monkeytype.com/
2. Hover over the time settings
3. Click the arrow popup
4. Select Mode and then Mode 2
5. Unselect Mode

### Environment

- OS: Windows 11
- Browser: Google Chrome
- Browser Version: Version 151.0.7922.172 (Official Build) (64-bit)

### Anything else?

Image

AzureNightlock:
### Description
This PR: Fixes ShareTestSettings.tsx when both mode and mode 2 is checked and only mode is unchecked, it resets mode 2 aswell

### Checks

Closes #8357

### DIFF
--- a/frontend/src/ts/components/modals/ShareTestSettings.tsx
+++ b/frontend/src/ts/components/modals/ShareTestSettings.tsx
@@ -103,6 +103,7 @@ export function ShareTestSettings(): JSXElement {
           onChange: ({ value }) => {
             if (!value) {
               form.setFieldValue("mode2", false);
+              form.setFieldValue("customText", false);
             }
           },
         }}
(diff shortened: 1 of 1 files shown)


Leonabcd123 (commented):
I think we should also reset the "Custom Text" field when mode is unchecked. It isn't really mode2, but it's the equivalent for custom mode and it's also a subcategory of mode in the modal.
image

PR #8367, opened 2026-08-26
impr: replace twitter bird icon with X logo (@rileygramlich)

rileygramlich:
### Description

The footer, about page, and profile social links still used the old Twitter bird icon (which is 3 years outdated now by the way), even though all three already link to x.com.

Font Awesome only added the `x-twitter` icon in 6.4, and this repo is pinned to `@fortawesome/fontawesome-free@5.15.4`, so there was no icon class to swap to. Rather than bump the dependency — which would mean icon renames and a much larger diff across the whole app — I inlined the logo as a small `XIcon` component, following the existing pattern in `DiscordAvatar.tsx`, which already embeds raw Font Awesome SVG paths with the license comment.

`Button` previously only accepted a Font Awesome icon via its `fa` prop, so I added an optional `icon` prop that renders in the same leading slot. That keeps the icon before the label, matching every other button.

I updated all three usages rather than just the footer, so the branding stays consistent.

Changed files:
- `frontend/src/ts/components/common/XIcon.tsx` (new)
- `frontend/src/ts/components/common/Button.tsx` — optional `icon` prop
- `frontend/src/ts/components/layout/footer/Footer.tsx`
- `frontend/src/ts/components/pages/AboutPage.tsx`
- `frontend/src/ts/components/pages/profile/UserDetails.tsx`

The button labels also changed from "twitter" to "x". Happy to revert that part if you'd rather keep the old wording.

### Checks

- [ ] Adding quotes?
- [ ] Adding a language?
- [ ] Adding a theme?
- [ ] Adding a layout?
- [ ] Adding a font?
- [x] Make sure the PR title follows the Conventional Commits standard.


fehmer:
https://github.com/monkeytypegame/monkeytype/pull/6066#issuecomment-2490796876

rileygramlich:
> [#6066 (comment)](https://github.com/monkeytypegame/monkeytype/pull/6066#issuecomment-2490796876)

I see that @[Miodec](https://github.com/Miodec) said "Long live the bird." Though at this point in 2026, it simply looks outdated, unprofessional, and like the website lacks upkeep.

Hope we can get this changed as it is actually quite annoying to still see.

Cheers,
Riley

PR #8375, opened 2026-08-29
fix(ttl-cache): concurrent calls during the first fetch get undefined (@boergeson)

### LINKED ISSUE #8363: Bug (backend): cacheWithTTL marks cache fresh before fetch resolves — failed fetch serves stale data for full TTL, no in-flight dedup

- [x] I have cleared my cache

N/A

N/A (backend source bug)

N/A

N/A

# Issue details

### Current Behavior

`cacheWithTTL` in `backend/src/utils/ttl-cache.ts` marks the cache as fresh *before* the fetch resolves:

```ts
// backend/src/utils/ttl-cache.ts:20-26
return async () => {
  if (lastFetchTime < Date.now() - ttlMs) {
    lastFetchTime = Date.now(); // updated before await completes
    cache = await fn();
  }
  return cache;
};
```

Consequences:

1. **Failed fetch poisons the cache for the full TTL.** If `fn()` rejects, the rejection propagates to that caller, but `lastFetchTime` has already been advanced — every subsequent call within the TTL returns stale cached data instead of retrying. This utility backs the PSA endpoint (`controllers/psa.ts`), so one failed upstream fetch serves stale content for the whole TTL window.
2. **No in-flight promise dedup.** When the TTL expires under concurrent requests, all callers run `fn()` simultaneously (thundering herd) since nothing records the pending promise.

### Expected Behavior

Only advance `lastFetchTime` after a successful fetch, and dedupe concurrent calls by caching the promise itself:

```ts
let lastFetchTime = 0;
let cache: T | undefined;
let inflight: Promise<T> | undefined;

return async () => {
  if (lastFetchTime < Date.now() - ttlMs) {
    inflight ??= fn()
      .then((result) => {
        cache = result;
        lastFetchTime = Date.now();
        return result;
      })
      .finally(() => {
        inflight = undefined;
      });
    return inflight;
  }
  return cache;
};
```

### Steps To Reproduce

1. Call a `cacheWithTTL`-wrapped getter whose `fn` throws.
2. Call again within the TTL — stale data is returned with no retry until TTL expiry.

### Environment

- Backend, `master` @ 91bd24bb8

boergeson:
### Description

Concurrent calls to a `cacheWithTTL` function while the first fetch is still running each got `undefined` back, since the cache was marked fresh before the data arrived. Now the first call starts the fetch and later calls wait on the same promise, so there is only one fetch and everyone gets the value.

A failed fetch still marks the cache fresh, so the stale value keeps being served until the ttl passes. That part is unchanged on purpose, the cache also throttles the database.

Closes #8363

### DIFF
--- a/backend/__tests__/utils/ttl-cache.spec.ts
+++ b/backend/__tests__/utils/ttl-cache.spec.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { cacheWithTTL } from "../../src/utils/ttl-cache";
+
+describe("cacheWithTTL", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(10_000);
+  });
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("refetches after the ttl", async () => {
+    const fn = vi.fn(async () => Date.now());
+    const cache = cacheWithTTL(1000, fn);
+
+    expect(await cache()).toBe(10_000);
+    vi.advanceTimersByTime(500);
+    expect(await cache()).toBe(10_000);
+    vi.advanceTimersByTime(600);
+    expect(await cache()).toBe(11_100);
+    expect(fn).toHaveBeenCalledTimes(2);
+  });
+
+  it("keeps the stale value until the ttl when the fetch fails", async () => {
+    const fn = vi
+      .fn<() => Promise<string>>()
+      .mockResolvedValueOnce("first")
+      .mockRejectedValueOnce(new Error("down"))
+      .mockResolvedValueOnce("third");
+    const cache = cacheWithTTL(1000, fn);
+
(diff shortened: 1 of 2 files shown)


fehmer:
hi @boergeson ,

thank you for your contribution.

Fetching only once if expired and waiting on the fetch on concurrent calls is fine.

I don't think the behavior in case of a fetch error is correct. The cache also acts like a throttel and if there is a problem with the database now each request will re-try and add additional load on the database. 

Currently the cache is only used for PSA and the plan was to use it for leaderboard as well, both are non-critical and I think it is better to serve stale data in case of an error instead of an error or no data. Therefor I would keep the current behavior.

@Miodec  what is your take on this?

boergeson:
Makes sense, didnt think of it as a throttle. Reworked it: a failed fetch still marks the cache fresh and the stale value is served until the ttl passes, only the in-flight dedup is left. Updated the test to check the stale value is kept after a failure.

PR #8376, opened 2026-08-29
fix: clear timeout ids from array in clearTimeouts (@priyanshu1976)

### LINKED ISSUE #8366: Bug: outOfFocusTimeouts array grows unboundedly — cleared timeout ids never removed from array

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

# Issue details

### Current Behavior

`outOfFocusTimeouts` in `frontend/src/ts/states/test.ts` accumulates timeout ids forever:

```ts
// states/test.ts:38-54
const outOfFocusTimeouts: (number | NodeJS.Timeout)[] = [];
...
} else {
  outOfFocusTimeouts.push(
    setTimeout(() => { set(val); }, 1000),
  );
}
```

Every unfocus event pushes a new id onto the array. `clearTimeouts()` (`utils/misc.ts:148-153`) only calls `clearTimeout` on each entry but **never removes them from the array**, so the array grows monotonically for the entire session. Long sessions with frequent alt-tabbing accumulate thousands of stale numeric ids (small but unnecessary retained memory, and the array scan in `clearTimeouts` grows linearly).

### Expected Behavior

Remove consumed entries after clearing, e.g.:

```ts
export function clearTimeouts(timeouts: (number | NodeJS.Timeout)[]): void {
  timeouts.forEach((to) => clearTimeout(to));
  timeouts.length = 0;
}
```

or store only the latest timeout id, since scheduling a newer unfocus supersedes older pending ones anyway.

### Steps To Reproduce

1. Open devtools memory snapshot / add a breakpoint log on `outOfFocusTimeouts.push`.
2. Blur/refocus the window repeatedly during tests.
3. Array length increases without bound and is never reset.

### Environment

- OS: Any
- Browser: Any
- Found via source review of `master` @ 91bd24bb8

priyanshu1976:
## Description

`clearTimeouts` in `frontend/src/ts/utils/misc.ts` cleared each timeout but never removed the ids from the array, so `outOfFocusTimeouts` in `states/test.ts` grew unboundedly for the whole session (every blur pushes a new id, focus never empties the array).

This empties the array after clearing (`timeouts.length = 0`) and collapses the identical if/else branches. Added a unit test verifying timeouts are cancelled and the array is emptied.

Closes #8366

🤖 Generated with [Claude Code](https://claude.com/claude-code)


priyanshu1976:
Closing in favor of a fresh PR.

PR #8387, opened 2026-09-03
fix(input-emulation): prevent char duping race condition (@HenryHYeary)

### LINKED ISSUE #8383: Deduplicating key presses when emulating

- [x] I have cleared my cache

Yes

Yes

Yes

Yes

### Account name

informthemusic

### Account config

_No response_

### Current Behavior

Image
Some letters are duplicated - only when emulating layout.
The issue never happens outside of layout emulation.

### Expected Behavior

Keys should not deduplicate.

### Steps To Reproduce

Let [LAYOUT] a layout you can type with
1. Emulate [LAYOUT] (regardless of if it already is your current layout)
2. Run a few tests, possibly notice some keys deduplicating
3. Un-emulate
4. Notice that the keys no longer deduplicate

### Environment

- OS: NixOS (`Linux 6.12.93 #1-NixOS SMP PREEMPT_DYNAMIC Tue Jun  9 10:26:06 UTC 2026 x86_64 GNU/Linux`)
- Browser: Firefox
- Browser Version: `Gecko/20100101 Firefox/152.0`

### Anything else?

_No response_

HenryHYeary:
#### Changes

I moved `event.preventDefault()` in the keydown handler to before the asynchronous call to `getCharFromEvent` where the previous placement of the preventDefault call (after the asynchronous call) was causing a race condition which meant it could lose the race against the browser's native default action. If the browser inserted the native character before preventDefault() ran, both the native character and the emulated character would get inserted, producing a duplicate.

I was able to reproduce the bug after adding a 50ms setTimeout delay
right after `layout = await __nonReactive.getInputLayout();`. I used `await new Promise(r => setTimeout(r, 50));`. 

I also refactored the  `keyEventCodes` reassignments in layout-emulator.ts to `ANSI_KEYCODES`, `ISO_KEYCODES`, and `MATRIX_KEYCODES`. I then used these constants in a set to specify all of the emulated key codes, so that a new `mayEmulateChar` check function in layout-emulator could do quick lookups on which keys it would be necessary to emulate. Without this check, preventDefault() would need to run unconditionally for every key while a custom layout is active, which breaks Backspace, Tab, Enter, and other keys the emulator doesn't handle.

Notably a key in mayEmulateChar's set could still resolve to null if a layout's keyVariants data is incomplete, but I believe it's the responsibility of the layout creator to ensure the keyVariants data is complete rather than relying on fallbacks.

### Checks

- [ ] Adding quotes?
  - Make sure to follow the [quotes documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/QUOTES.md)
- [ ] Adding a language?
- [ ] Adding a theme?
  - [ ] (optional) Add theme css file to `frontend/static/themes`
  - [ ] Add some screenshots of the theme, especially with different test settings (colorful, flip colors) to your pull request
- [ ] Adding a layout?
- [ ] Adding a font?
  - Make sure to follow the [fonts documentation](https://github.com/monkeytypegame/monkeytype/blob/master/docs/FONTS.md)
  - [ ] Add font file to `frontend/static/webfonts`

Closes #8383


HenryHYeary:
It looks like the input of the commit message was truncated and that's why the run failed since I did include my username between parentheses at the end of it

HenryHYeary:
Please let me know if I need to shorten the commit message to fix this

PR #7119, opened 2025-11-18
refactor: input rewrite v3 (@miodec)

Miodec:
input goes brr

PR #8389, opened 2026-09-04
fix: split space-separated event names in input listeners (@zacharybergin)

### LINKED ISSUE #8361: Bug: copy/paste prevention listeners never fire — space-separated event names passed to addEventListener

- [x] I have cleared my cache

N/A

Yes (source-code bug, independent of account state)

N/A

N/A

# Issue details

### Current Behavior

In `frontend/src/ts/input/listeners/misc.ts`, native `addEventListener` is called with space-separated event names (jQuery-style syntax), which registers literal event types that are never dispatched:

```ts
// frontend/src/ts/input/listeners/misc.ts:12-19
inputEl.addEventListener("copy paste", (event) => {
  event.preventDefault();
});

//this might not do anything
inputEl.addEventListener("select selectstart", (event) => {
  event.preventDefault();
});
```

The browser treats `"copy paste"` as a single event type named `copy paste`, so **neither listener ever fires**. As a result, copy/paste on the hidden words input is not actually prevented during a test.

### Expected Behavior

Each event name should be registered separately:

```ts
for (const eventName of ["copy", "paste"]) {
  inputEl.addEventListener(eventName, (event) => event.preventDefault());
}
```

(The `select`/`selectstart` pair should be evaluated separately — the inline comment already suspects it may be dead code and could possibly be removed.)

### Steps To Reproduce

1. Start a test.
2. Focus the words input and attempt to select/copy text or paste into it.
3. The default browser behavior is not prevented because the handlers were never attached to real event types.

### Environment

- OS: Any
- Browser: Any
- Found via source review of `master` @ 91bd24bb8

zacharybergin:
### Description

addEventListener only takes one event name at a time. So "copy paste" and "select selectstart" were not really registering both events.

This meant copy and paste into the words input was never actually being blocked. It looks like the space separated syntax was left over from the jQuery .on() code before the input rewrite in #7119.

I split these into separate listeners so each event should actually be registered correctly.

One thing I noticed is that fixing selectstart also makes the selection blocking work again. It's been broken since #7119. So this is technically a behavior change and not just a bug fix. Let me know if you'd rather I leave that part out.

Also select isn't cancelable so preventDefault() doesn't really do anything for that event.

Closes #8361


fehmer:
duplicate #8371

